PR #20, opened 2024-08-15
set up DB infra and prompting

broskicodes:
This pr adds set up for a postgres DB to store registered components. 

I also change the prompting strategy to only provide relevant context and allow AI to send text msgs.

I've made changes to hydra client and backend structure to facilitate these updates. Please comment any questions that you have.

### DIFF
--- a/package/src/hydra-ai/hydra-ai-backend.ts
+++ b/package/src/hydra-ai/hydra-ai-backend.ts
@@ -1,7 +1,7 @@
 import "server-only";
 import AIService from "./ai-service";
 import { ComponentChoice } from "./model/component-choice";
-import { ComponentMetadata } from "./model/component-metadata";
+import { AvailableComponents, ComponentMetadata } from "./model/component-metadata";
 import { InputContext } from "./model/input-context";
 import { drizzle, NodePgDatabase } from "drizzle-orm/node-postgres";
 import { Pool } from "pg";
@@ -46,7 +46,7 @@ export default class HydraBackend {
 
   public async generateComponent(
     message: string,
-    availableComponents: ComponentMetadata[]
+    availableComponents: AvailableComponents
   ): Promise<ComponentChoice> {
     const context: InputContext = {
       prompt: message,
--- a/package/src/hydra-ai/hydra-ai-backend.ts
+++ b/package/src/hydra-ai/hydra-ai-backend.ts
@@ -27,7 +27,7 @@ export default class HydraBackend {
     propsDefinition?: ComponentPropsMetadata
   ): Promise<boolean> {
     try {
-      let _ = await this.dbConnection.insert(registeredComponents).values({
+      await this.dbConnection.insert(registeredComponents).values({
         name,
         description,
         prop_definitions: propsDefinition,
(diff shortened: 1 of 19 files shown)


package/src/hydra-ai/hydra-server-action.ts
```diff
@@ -2,27 +2,36 @@
...
-  message: string,
-  availableComponents: ComponentMetadata[],
-  systemInstructions?: string
-): Promise<ComponentChoice> {
-  const hydra = getHydraBackend(systemInstructions);
-  const response = await hydra.generateComponent(message, availableComponents);
-  return response;
-}
-
 const getHydraBackend = (systemInstructions?: string): HydraBackend => {
```
MichaelMilstead: I guess we can remove `systemInstructions` param if it's not used

package/src/hydra-ai/hydra-ai-client.ts
```diff
@@ -13,19 +13,26 @@ interface ComponentRegistry {
...
-    getComponentContext?: () => any | Promise<any>
-  ): void {
+    getComponentContext?: () => any | Promise<any>,
+    callback: (
+      name: string,
+      description: string,
+      propsDefinition: ComponentPropsMetadata
+    ) => Promise<boolean> = saveComponent
+  ): Promise<void> {
+    let success = await callback(name, description, propsDefinition);
```
MichaelMilstead: nit: const instead of 'let' if the variable isn't reassigned

package/src/hydra-ai/hydra-ai-backend.ts
```diff
@@ -1,23 +1,52 @@
...
+import { AvailableComponents, ComponentMetadata } from "./model/component-metadata";
 import { InputContext } from "./model/input-context";
+import { drizzle, NodePgDatabase } from "drizzle-orm/node-postgres";
+import { Pool } from "pg";
+import { registeredComponents } from "../db/schema";
+import { ComponentPropsMetadata } from "./model/component-props-metadata";
 
 export default class HydraBackend {
   private aiService: AIService;
+  private dbConnection: NodePgDatabase;
```
MichaelMilstead, line 13: Could be cool if we eventually use a generic "ComponentRepositoryInterface" so people can use whatever storage they want, but we can mess with that later.

package/src/hydra-ai/hydra-ai-backend.ts
```diff
@@ -1,23 +1,52 @@
...
+    this.aiService = new AIService(openAIKey, openAIModel);
+  }
+
+  public async registerComponent(
+    name: string,
+    description: string,
+    propsDefinition?: ComponentPropsMetadata
+  ): Promise<boolean> {
+    try {
+      let _ = await this.dbConnection.insert(registeredComponents).values({
```
MichaelMilstead: nit: Instead of `let _` we can just do `await this.db...` without assigning a variable

MichaelMilstead (approved):

package/src/hydra-ai/hydra-ai-backend.ts
broskicodes: atm any postgres db should work.

what other storage methods were u thinking?

MichaelMilstead: For us I think postgres is good but for somebody who wants to self-host hydra it would be cool if they could use whatever they want, as long as they implement some interface

broskicodes: i think postgres is most common for a lot of devs and we can add other implementations as the need arises.

if you've seen ppl request other data stores, i'm happy to add them?

i just think we should get the demo app built out first and can make the pkg more extensible later.

MichaelMilstead: yep agreed! No need to worry about it right now

PR #21, opened 2024-08-19
Add component tool definition and calling

MichaelMilstead:
Allows package users to define data fetching functions, or "tools", for each component with parameters.

Updates registerComponent to store the function definitions.
Updates generateComponent to let AI decide when to use the tools. 

To enable this, generateComponent is split into two parts:
1: choose component (or decide no component should be used)
2: hydrate component

hydrate component passes the chosen component and any of its tools to AI, and lets AI either create props, or request that one of the tools be called to collect more data.

Currently, the tool needs to be called clientside where it is defined. So if AI decides to run a tool, that decision is passed back to the client, the client runs the tool to get the extra context data, and then calls hydrate component, passing the tool run results, so the AI can finally generate props for the chosen component, and return it back to the client.

This also means that we should be using fewer tokens and generation should be faster, since each component's context function doesn't need to be run each time we generate a component.

Some things to note:
- Currently assumes that a tool's definition defines its parameters in the order they are used in the actual function
- Because the tool needs to run on the client, the tool results are not passed to openai in the usual way (in the following message of the same conversation openai requested a tool call)
- There is no way to define an action that is not associated with a component, so we don't yet have the possibility of having the AI take "actions" without showing a component (creating, updating, deleting)


package/src/hydra-ai/hydra-ai-client.ts
```diff
@@ -55,9 +66,16 @@ export default class HydraClient {
     callback: (
```
broskicodes: nit: should prolly give this a more descriptive name since there r multiple fns

package/src/hydra-ai/model/component-choice.ts
```diff
@@ -3,3 +3,15 @@ export interface ComponentChoice {
   props: any | null;
   message: string;
 }
+
+export interface ComponentDecision extends ComponentChoice {
+  toolCallRequest?: ToolCallRequest;
+}
```
broskicodes, line 6: why not add `toolCallRequest` param to `ComponentChoice` instead of creating new type?

package/src/hydra-ai/hydra-ai-backend.ts
```diff
@@ -57,4 +67,12 @@ export default class HydraBackend {
 
     return componentChoice;
```
broskicodes: nit: can return from aiService call for consistency with `hydrateComponentWithData` fn

package/src/hydra-ai/hydra-ai-backend.ts
```diff
@@ -1,36 +1,46 @@
...
-    propsDefinition?: ComponentPropsMetadata
+    propsDefinition?: ComponentPropsMetadata,
+    contextToolDefinitions?: ComponentContextToolMetadata[]
   ): Promise<boolean> {
     try {
       await this.dbConnection.insert(registeredComponents).values({
         name,
         description,
         prop_definitions: propsDefinition,
+        // context_tools: contextToolDefinitions,
```
broskicodes: update once schema changes are complete

package/src/db/schema.ts
```diff
@@ -1,11 +1,11 @@
-import { uuid, text, jsonb, pgTable, timestamp } from "drizzle-orm/pg-core";
+import { jsonb, pgTable, text, timestamp, uuid } from "drizzle-orm/pg-core";
 
 export const registeredComponents = pgTable("registered_components", {
```
broskicodes, line 3: add column to store `context_tools`

package/src/hydra-ai/ai-service.ts
```diff
@@ -79,9 +79,13 @@ Finally, if you decide that a component should be generated, you will output the
       },
     ]);
 
-    const shouldGenerate = decisionResponse.match(/<decision>(.*?)<\/decision>/)?.[1];
+    const shouldGenerate = decisionResponse.message.match(
+      /<decision>(.*?)<\/decision>/
+    )?.[1];
     if (shouldGenerate === "false") {
```
broskicodes, line 82: nit: add newline above for readability

package/src/hydra-ai/ai-service.ts
```diff
@@ -119,51 +125,103 @@ This response should be short and concise.`;
...
+    );
+
+    const componentDecision: ComponentDecision = {
+      message: "Fetching additional data",
+      componentName: component.name,
+      props: null,
+      toolCallRequest: generateComponentResponse.toolCallRequest,
+    };
+
+    if (generateComponentResponse.message !== "") {
```
broskicodes: is `response.message` guaranteed to be empty if it call a tool?

might be a better way to check/separate the cases.

PR #30, opened 2024-09-01
WIP: Add JSON schemas for all OpenAI requests

Antel0pe:
Added JSON Schemas for
- If a component should be generated
- If a component shouldn't be generated, the message to display
- If a component should be generated, the details required

Modified API call to require these schemas be included

### DIFF
--- a/package/src/hydra-ai/ai-service.ts
+++ b/package/src/hydra-ai/ai-service.ts
@@ -197,16 +197,11 @@ ${
         toolResponse
       )}`
     : `You can also use any of the provided tools to fetch data needed to pass into the component.`
-}
-
-${this.generateZodTypePrompt(generateSpecifiedComponentSchema)}`;
+}`;
 
     const tools = toolResponse
       ? undefined
       : this.openAIToolFromMetadata(component.contextTools);
-
-    const componentContextTools = component.contextTools.find(tool => tool.definition.name === component.name);
-    
 
     const generateComponentResponse = await this.callOpenAI(
       [
@@ -256,8 +251,8 @@ ${this.generateZodTypePrompt(generateSpecifiedComponentSchema)}`;
 
   async callOpenAI(
     messages: ChatCompletionMessageParam[],
-    schema: z.ZodSchema<any>,
-    schemaName: string,
+    responseSchema: z.ZodSchema<any>,
+    responseSchemaName: string,
     tools?: ChatCompletionTool[],
   ): Promise<OpenAIResponse> {
     let componentTools = tools;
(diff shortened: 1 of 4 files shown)


Antel0pe:
Running into bit of an error here and I **believe** it is because of this option: https://github.com/Antel0pe/hydraai/blob/e5ffbef100f2de372c56132c91f90427a36b4ddc/package/src/hydra-ai/ai-service.ts#L31

When it's not included, no error. When it IS included I get the following error: "open ai Internal error: Error: 400 Invalid schema for response_format 'generateSpecifiedComponentSchema': In context=('properties', 'props'), 'additionalProperties' is required to be supplied and to be false."

The .passthrough option says that you can include any additional properties you need with this schema and that is being used by chatgpt to supply the component props values. 

Since before my commits, this schema+option was given in a string along with the request, it didn't cause a problem. Now that I'm using Structured Output and demanding the output be given with this schema option, it's giving an error that I can't ask it to specify additional properties in the response.

The problem I'm currently in is dynamically defining the zod schema according to the component prop definitions so that we don't need the passthrough option and can avoid the error... which looks pretty close to this: https://github.com/Antel0pe/hydraai/blob/e5ffbef100f2de372c56132c91f90427a36b4ddc/package/src/hydra-ai/ai-service.ts#L319

package/src/hydra-ai/ai-service.ts
```diff
@@ -156,12 +199,15 @@ ${
...
 }
 
-${this.generateZodTypePrompt(schema)}`;
+${this.generateZodTypePrompt(generateSpecifiedComponentSchema)}`;
 
     const tools = toolResponse
       ? undefined
       : this.openAIToolFromMetadata(component.contextTools);
 
+    const componentContextTools = component.contextTools.find(tool => tool.definition.name === component.name);
```
MichaelMilstead: looks like this var isn't used anywhere? Maybe we can remove it

package/src/hydra-ai/ai-service.ts
```diff
@@ -156,12 +199,15 @@ ${
     : `You can also use any of the provided tools to fetch data needed to pass into the component.`
 }
 
-${this.generateZodTypePrompt(schema)}`;
+${this.generateZodTypePrompt(generateSpecifiedComponentSchema)}`;
```
MichaelMilstead: If we specify the response schema in the openAI call now, I guess we can remove this part of the prompt. Also I think this is the only place the `generateZodTypePrompt` function is used, so we could probably remove that too

package/src/hydra-ai/ai-service.ts
```diff
@@ -209,8 +256,9 @@ ${this.generateZodTypePrompt(schema)}`;
 
   async callOpenAI(
     messages: ChatCompletionMessageParam[],
+    schema: z.ZodSchema<any>,
+    schemaName: string,
```
MichaelMilstead: can we rename these to `responseSchema` and `responseSchemaName` ?

michaelmagan:
hi antelope can you help me understand what this is for? also mind rebasing with latest commits?

alecf:
closing this out, we went another way (autogenerating APIs with stainless.com)

PR #129, opened 2025-03-16
fix: update tsconfig

MichaelMilstead:
updates tsconfig so relative imports don't need file extension


alecf:
we had to take a different approach

PR #169, opened 2025-03-21
feat: add debounce, optimistic updates, and helper functions for prop updates

michaelmagan:
## Summary by CodeRabbit

- **New Features**
  - Introduced debouncing support to optimize update frequency, enhancing performance.
  - Added an automatic synchronization mechanism that updates component state when properties change.
  - Streamlined hook exports for smoother integration.

- **Enhancements**
  - Improved state management to reduce redundant updates, providing a more responsive user experience.

### DIFF
--- a/react-sdk/src/hooks/use-component-state.tsx
+++ b/react-sdk/src/hooks/use-component-state.tsx
@@ -1,11 +1,16 @@
-import { useCallback, useEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useState } from "react";
+import { useDebouncedCallback } from "use-debounce";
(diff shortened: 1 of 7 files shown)


react-sdk/src/hooks/use-component-state.tsx
```diff
@@ -70,52 +238,56 @@ export function useTamboComponentState<S>(
...
-        },
-        false,
-      );
-      await client.beta.threads.messages.updateComponentState(
-        threadId,
-        messageId,
-        { state: { [keyName]: newValue } },
-      );
+          false,
+        );
```
MichaelMilstead: I think we can remove this updateThreadMessage call since it will happen in the debouncedsave

PR #343, opened 2025-04-24
refactor(UI): update UI components and styles for improved usability and integration

akhileshrangani4:


cli/src/registry/canvas-space/canvas-space.tsx
```diff
@@ -0,0 +1,355 @@
...
+
+/**
+ * Props for the CanvasSpace component
+ * @interface
+ */
+interface CanvasSpaceProps {
+  /** Optional CSS class name for custom styling */
+  className?: string;
+  /** Position of the canvas space, either 'left' or 'right' (default: 'right') */
+  position?: "left" | "right";
```
michaelmagan: could size/position all be done with classNames instead of variants?

what if want the entire page to be canvas? or have four canvases with different chats?

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
```diff
@@ -22,6 +24,12 @@ export interface MessageThreadCollapsibleProps
   contextKey?: string;
   /** Whether the collapsible should be open by default (default: false) */
   defaultOpen?: boolean;
+  /** Whether to enable the canvas space */
+  enableCanvasSpace?: boolean;
```
michaelmagan: i wonder if we should handle this with a provider?

<CanvasThreadProvider>

it wraps the component with a shared context?

we'd need the component to check if its within a provider. @alecf  or @MichaelMilstead  thoughts?

why not use tambo provider? 
con: what if you want multiple chats + canvas combos
pro: one less provider already there...

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
```diff
@@ -22,6 +24,12 @@ export interface MessageThreadCollapsibleProps
...
+  /** Optional styling variant for the message container */
+  variant?: VariantProps<typeof messageVariants>["variant"];
+  /** position of the panel */
+  position?: "left" | "right";
```
michaelmagan: same feedback here could we just use classes? make it way more flexible.

in our layouts we can put nice defaults and in our docs. this is other ui libraries do it.

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
```diff
@@ -22,6 +24,12 @@ export interface MessageThreadCollapsibleProps
   contextKey?: string;
   /** Whether the collapsible should be open by default (default: false) */
   defaultOpen?: boolean;
+  /** Whether to enable the canvas space */
+  enableCanvasSpace?: boolean;
+  /** Optional styling variant for the message container */
+  variant?: VariantProps<typeof messageVariants>["variant"];
```
michaelmagan, line 46: not sure what this does without reading the code. could we document it better.

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
```diff
@@ -33,124 +41,144 @@ export interface MessageThreadCollapsibleProps
...
+    const { thread } = useTambo();
 
-  React.useEffect(() => {
-    const handleKeyDown = (event: KeyboardEvent) => {
-      if ((event.metaKey || event.ctrlKey) && event.key === "k") {
-        event.preventDefault();
-        setIsOpen((prev) => !prev);
-      }
-    };
+    React.useEffect(() => {
```
michaelmagan: i feel like this should in the message-input component?

probably should also be a primitive component sou can choose to have it or not.

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
```diff
@@ -33,124 +41,144 @@ export interface MessageThreadCollapsibleProps
...
+            aria-expanded={isOpen}
+            aria-controls="message-thread-content"
           >
-            <ThreadContent />
-          </div>
-          <MessageSuggestions />
-          <div className="p-4 border-t border-gray-200">
-            <MessageInput contextKey={contextKey} />
+            <div className="flex items-center gap-2">
+              <span>{isOpen ? "Conversations" : "Use AI"}</span>
```
michaelmagan: we should use a config at the top to define this value and set it as "start chatting with tambo"

or something like that.

cli/src/registry/message-thread-full/message-thread-full.tsx
```diff
@@ -1,43 +1,100 @@
...
+interface ScrollableMessageContainerProps
+  extends React.HTMLAttributes<HTMLDivElement> {
+  /** Children elements to render inside the container */
+  children: React.ReactNode;
+}
+
+/**
+ * A scrollable container for message content with auto-scroll functionality
+ */
+const ScrollableMessageContainer = React.forwardRef<
```
michaelmagan: feels like it should be it's own component and imported

cli/src/registry/message-thread-panel/message-thread-panel.tsx
```diff
@@ -32,89 +40,159 @@ export interface MessageThreadPanelProps
...
-    if (scrollContainerRef.current && thread?.messages?.length) {
-      const timeoutId = setTimeout(() => {
-        if (scrollContainerRef.current) {
-          scrollContainerRef.current.scrollTo({
-            top: scrollContainerRef.current.scrollHeight,
-            behavior: "smooth",
-          });
-        }
-      }, 100);
+    useEffect(() => {
```
michaelmagan: again anything that needs to be re-written in 3 different blocks components should for sure be it's own component.

cli/src/registry/message/markdownComponents.tsx
```diff
@@ -2,14 +2,34 @@
...
+import "highlight.js/styles/github.css";
 
+/**
+ * Markdown Components for React-Markdown
+ *
+ * This module provides customized components for rendering markdown content with syntax highlighting.
+ * It uses highlight.js for code syntax highlighting and supports streaming content updates.
+ *
+ * @example
+ * ```tsx
```
michaelmagan, line 18: love this so clean :)

cli/src/types/components.d.ts
```diff
@@ -16,25 +21,87 @@ type ComponentVariant = "default" | "solid" | "bordered" | null | undefined;
...
+    children?: React.ReactNode;
   }
+
+  export type ThreadContentMessagesProps = React.HTMLAttributes<HTMLDivElement>;
+
   export const ThreadContent: React.ForwardRefExoticComponent<
     ThreadContentProps & React.RefAttributes<HTMLDivElement>
   >;
+
+  export const ThreadContentRoot: React.ForwardRefExoticComponent<
```
michaelmagan, line 87: do wee need to include root?

i'm on the fence but leaning slightly to omitting it? CC @alecf  @MichaelMilstead

cli/src/registry/canvas-space/canvas-space.tsx
```diff
@@ -0,0 +1,355 @@
...
+  /** Whether the canvas space is the only panel visible (default: false), useful for when using the control bar */
+  isAlone?: boolean;
```
michaelmagan: feels like we could drop this if we do the comment above? @akhileshrangani4 right?

cli/src/registry/message-input/message-input.tsx
```diff
@@ -30,58 +30,100 @@ const messageInputVariants = cva("w-full", {
...
+  textareaRef: React.RefObject<HTMLTextAreaElement>;
+  submitError: string | null;
+  setSubmitError: React.Dispatch<React.SetStateAction<string | null>>;
+}
+
+/**
+ * React Context for sharing message input data and functions among sub-components.
+ * @internal
+ */
+const MessageInputContext =
```
michaelmagan, line 65: this is really nice. i like it.

cli/src/registry/message-thread-full/message-thread-full.tsx
```diff
@@ -55,30 +112,87 @@ export const MessageThreadFull = React.forwardRef<
...
+        </div>
+
+        {/* Message thread content */}
+        <ScrollableMessageContainer>
+          <ThreadContentRoot
+            enableCanvasSpace={enableCanvasSpace}
+            className="py-4"
+            variant={variant}
+          >
+            <ThreadContentMessages />
```
michaelmagan, line 138: could we turn this into wrapper component of messages

<ThreadContentMessages>
<AIMessage/>
<UserMessage/>
<ThreadContentMessages>

The drop the variants from ThreadContentMessages and do something like below?


<ThreadContentMessages>
<AIMessage className="rounded-lg border-bg">
<MessageCopy>
<MessageFeedback>
<ThumbsUpDown/>
<Label>Give us Feedback</Label>
</MessageFeedback>
</AIMessage>
<UserMessage>
<EditMessage>
<Icon.Pencil> Edit
</Edit Message>
</UserMessage>
<ThreadContentMessages>

you get the idea?

CC @alecf @MichaelMilstead any thoughts on this?

michaelmagan (changes requested):

michaelmagan (changes requested):
heyasfasdf

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
MichaelMilstead: hmm I'm kind of wondering whether we need the CanvasAwareContainer (and this variable) at all

cli/src/registry/message-thread-full/message-thread-full.tsx
MichaelMilstead: mm not sure I understand. What's the "give us feedback, edit"?

cli/src/types/components.d.ts
MichaelMilstead: Is the decision about whether we expect devs to edit ThreadContent directly or import and configure ThreadContentRoot when they want to customize?

cli/src/registry/canvas-space/canvas-space.tsx
akhileshrangani4: actually, position prop is needed for the canvas space to be able to auto-adjust with different types of message thread components, the logic determines the width of canvas based on its position on either side.

If the canvas space is used alone, it will spread throughout the entire page by default. 

To have 4 different canvases with 4 different chats is not possible with how the current canvas is designed, as it is made to adjust itself with every type of message thread, so it has logic that would conflict with more canvases. Also, it currently pulls data from `const { thread } = useTamboThread();` so there's no way to specify which thread a particular canvas should connect to - it always uses whatever thread is provided by the context

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
akhileshrangani4: Actually, CanvasAwareContainer handles adjusts layout when canvas space component is used. It dynamically resizes and repositions the message thread based on:

- Whether a canvas is present (enableCanvasSpace)
- The canvas position (left or right side)
- The width of both the canvas and sidebar(ThreadHistory)

Without it, the message thread wouldn't properly adapt when users are working with a canvas alongside their conversation. It's basically helping our UI be responsive to different workspace configurations.

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
akhileshrangani4: Yup, here the position prop can be removed easily as it just determines the position and that could be put into classes. Updated it.

```
<MessageThreadCollapsible
  contextKey="tambo-template"
  enableCanvasSpace={true}
  className="left-4"
/>
```

akhileshrangani4: Also, for the `MessagethreadPanel` component, I have removed the position prop and now using className for it.

```
<MessageThreadPanel
  contextKey="tambo-template"
  enableCanvasSpace={true}
  className="left"
/>
```

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
MichaelMilstead: Makes sense, but wondering if instead of manually setting positions based on the size of the canvas, can we just make the size of each component dependent on its parent or classes passed in?

And if I want a canvas next to a message thread, I can just place them side by side and have them fill the parent div with maybe the message thread having a default max-width

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
akhileshrangani4: it is basically for managing the variant of messages in the thread. will add better documentation.

cli/src/registry/canvas-space/canvas-space.tsx
akhileshrangani4: yes, dropped the `isAlone` prop and mentioned somethings about `position` prop above.

cli/src/types/components.d.ts
akhileshrangani4: Removed the "Root" suffix from all places.

cli/src/registry/message-thread-full/message-thread-full.tsx
michaelmagan: hey sorry just seeing this.

what i mean is that we want features to be in their own primitives for a message:
```
<Message>
  <MessageInput>What can we do for you?</MessageInput>
  <MessageActions>
    <Feedback/>
    <Edit/>
    <SubmitButton><Icon.ArrowUp/></SubmitButton>
  </MessageActions>
</Message>
```

cli/src/types/components.d.ts
michaelmagan: This is just adding the suffix for Root.

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
michaelmagan: Great much better.

cli/src/registry/canvas-space/canvas-space.tsx
akhileshrangani4: I changed the logic so that now its in className.

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
akhileshrangani4: you're right, did this!

cli/src/registry/message-thread-full/message-thread-full.tsx
akhileshrangani4: done!

cli/src/registry/message-thread-panel/message-thread-panel.tsx
akhileshrangani4: done!

cli/src/registry/message-thread-collapsible/message-thread-collapsible.tsx
akhileshrangani4: added a config which is pretty obvious for the user to customize:
```
  /**
   * Configuration for the MessageThreadCollapsible component
   */
  const THREAD_CONFIG = {
    labels: {
      openState: "Conversations",
      closedState: "Start chatting with tambo",
    }
  };
  ```

File: cli/src/registry/message-input/message-input.tsx [hunk 14]
```diff
@@ -30,138 +30,364 @@ const messageInputVariants = cva("w-full", {
...
+ * It establishes the context for its children and handles the form submission.
+ * @component MessageInput.Root
  * @example
  * ```tsx
- * <MessageInput
- *   contextKey="my-thread"
- *   variant="solid"
- *   className="custom-styles"
- * />
+ * <MessageInput.Root contextKey="my-thread" variant="solid">
```
michaelmagan: Should we remove .Root?

MessageInputContainer
MessageInputTextArea (needs to be inside a container)

cli/src/registry/message-suggestions/message-suggestions.tsx
```diff
@@ -66,7 +124,6 @@ export const MessageSuggestions = React.forwardRef<
     if (lastAiMessage && lastAiMessage.id !== lastAiMessageIdRef.current) {
```
michaelmagan: i wonder if this would be a good time to replace this with a better hook that only generates suggestions when a decision loop ends. 

@MichaelMilstead do we have a good hook for this yet?

right now its called every time the a message from the ai is returned.

cli/src/registry/message-thread-panel/message-thread-panel.tsx
```diff
@@ -22,99 +48,222 @@ export interface MessageThreadPanelProps
...
+              <ThreadHistoryNewButton />
+              <ThreadHistorySearch />
+              <ThreadHistoryList />
+            </ThreadHistory>
+          </div>
+        )}
+
+        <div className="flex flex-col h-full flex-grow">
+          <ScrollableMessageContainer>
+            <ThreadContent variant={variant}>
```
michaelmagan, line 252: @akhileshrangani4 

here is what i was hoping for with messages:
ThreadContentMessages Takes a messages array and a renderMessage function (or something like this). You define exactly how each message looks inside our blocks thread-panel, thread-full, etc.

Message, MessageContent, MessageActions
- Full control over layout and styling.
- No hardcoded assistant or user behavior.
- Easy to add custom actions (like copy, like, dislike, etc.), or new tool call components, etc.


```
<ScrollableMessageContainer>
  <ThreadContentMessages
    messages={messages}
    renderMessage={(message, idx) => (
      <Message
        key={idx}
        className={message.role === "assistant" ? "flex justify-start" : "flex justify-end"}
      >
        <div className="flex flex-col">
          <MessageContent
            className={message.role === "assistant" ? "bg-muted p-4 rounded-lg" : "bg-primary text-primary-foreground p-4 rounded-lg"}
          >
            {message.content}
          </MessageContent>

          {message.role === "assistant" && enableCanvasSpace && (
            <MessageRenderedComponentArea enableCanvasSpace={enableCanvasSpace} />
          )}

          {message.role === "assistant" && (
            <MessageActions className="self-end">
              <MessageAction tooltip="Copy to clipboard">
                <Button size="icon" onClick={() => copy(message.content)}>
                  <CopyIcon className="h-4 w-4" />
                </Button>
              </MessageAction>
            </MessageActions>
          )}

          {message.role === "user" && (
            <MessageActions className="self-end">
              <MessageAction tooltip="Edit message">
                <Button size="icon" onClick={() => editMessage(idx)}>
                  <EditIcon className="h-4 w-4" />
                </Button>
              </MessageAction>
            </MessageActions>
          )}
        </div>
      </Message>
    )}
  />
</ScrollableMessageContainer>

```

michaelmagan (dismissed):
hey here are key comments.

cli/src/registry/message-thread-panel/message-thread-panel.tsx
akhileshrangani4: okay, I got the idea, working on this now.

michaelmagan: awesome!

PR #365, opened 2025-05-02
feat: Allow MCP urls passed into tambo provider

MichaelMilstead:
A list of mcp urls can be passed into the TamboProvider, and a request will be made to each to fetch tool definitions, which are then registered with Tambo.

When Tambo wants to use one of the tools, it makes a post request to the same route with the tool name and args.

### DIFF
--- a/react-sdk/src/providers/tambo-registry-provider.tsx
+++ b/react-sdk/src/providers/tambo-registry-provider.tsx
@@ -7,7 +7,7 @@ import React, {
   useEffect,
   useState,
 } from "react";
-import { ZodSchema } from "zod";
+import { z, ZodSchema } from "zod";
 import zodToJsonSchema from "zod-to-json-schema";
 import {
   ComponentRegistry,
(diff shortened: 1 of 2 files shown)


react-sdk/src/providers/tambo-registry-provider.tsx
```diff
@@ -178,6 +195,14 @@ export const TamboRegistryProvider: React.FC<
     }
   }, [registerTools, userTools]);
 
+  useEffect(() => {
+    if (mcpServers) {
+      fetchToolsFromMCPServers(mcpServers).then((tools) => {
```
alecf, line 200: Use async here rather than then() - it's harder to maintain 

(Use an inner async function that you call from inside useEffect)

alecf: This can also get you cancellation via an abortcontroller

michaelmagan:
@MichaelMilstead any reason to not push this?

MichaelMilstead:
> @MichaelMilstead any reason to not push this?

Hmm thought I closed this one already. Anyway I did this in a way I don't think we want to do it, I believe @alecf will have a separate PR to do it the right way

PR #443, opened 2025-05-20
fix(cli): update npx command to use create-app without version specifier

michaelmagan:


create-tambo-app/src/index.ts
```diff
@@ -3,7 +3,7 @@
 import { spawn } from "child_process";
 
 // Use npx to ensure we get the latest version of tambo
-const args = ["-y", "tambo", "create-app@latest", ...process.argv.slice(2)];
+const args = ["-y", "tambo", "create-app", ...process.argv.slice(2)];
```
martinnormark, line 6: Shouldn't `@latest` be at the `tambo` command?

```suggestion
const args = ["-y", "tambo@latest", "create-app", ...process.argv.slice(2)];
```

MichaelMilstead: yep that looks right, pr here: https://github.com/tambo-ai/tambo/pull/445

michaelmagan: Nice catch.

PR #468, opened 2025-05-28
feat: useStreamingState hook

michaelmagan:


MichaelMilstead:
I'm a little confused about what it looks like for a dev to use this, can we add a couple examples in the pr description?

michaelmagan:
@CharlieHelps  can you add some example of how to use this?

michaelmagan:
Honestly I don't like this. I'm going to write up a better doc explaining this and give it another go.

PR #654, opened 2025-07-10
feat(sdk): Add TamboPropStreamProvider

aarmful:
Here's the `TamboStreamProvider` proof of concept you asked for. It provides per-key state management with compound components (`Loading`, `Empty`, `Content`) that work with both streaming and static data, offering a more declarative alternative to manual prop status checking.

This was ported from another provider I wrote for Prisma Pulse, but I haven't been able to do proper testing yet. Would love to chat with someone about testing strategies so we can ensure we cover the edge cases you encounter in real usage.

### Example usage:
```tsx
<TamboStreamProvider data={{ title, content, createdAt }}>
  <div>
    <h3>{title}</h3>
    <p>{content}</p>
    
    <TamboStreamProvider.Loading key="createdAt">
      <p>Loading creation date...</p>
    </TamboStreamProvider.Loading>
    
    <TamboStreamProvider.Content key="createdAt">
      <p>Created: {createdAt}</p>
    </TamboStreamProvider.Content>
    
    <TamboStreamProvider.Empty key="createdAt">
      <p>No creation date available</p>
    </TamboStreamProvider.Empty>
  </div>
</TamboStreamProvider>
```

This is just a POC to see if this pattern feels good and could be useful. The idea is to replace stuff like `propStatus["createdAt"]?.isSuccess && <p>{createdAt}</p>` with something more declarative.

Would love to get your thoughts on whether this direction feels right! 🤔

### DIFF
--- a/react-sdk/src/providers/__tests__/tambo-prop-stream-provider.test.tsx
+++ b/react-sdk/src/providers/__tests__/tambo-prop-stream-provider.test.tsx
@@ -1,0 +1,441 @@
+import { render, screen } from "@testing-library/react";
+import React from "react";
+import { StreamStatus } from "../../hooks/use-tambo-stream-status";
+import {
+  TamboPropStreamProvider,
+  useTamboStream,
+} from "../tambo-prop-stream-provider";
+
+// Helper component to test hook usage
+const TestHookComponent: React.FC<{ testKey?: string }> = ({
+  testKey = "default",
+}) => {
+  const { data, streamStatus, getStatusForKey } = useTamboStream();
+  const status = getStatusForKey(testKey);
+
+  return (
+    <div>
+      <div data-testid="data">{JSON.stringify(data)}</div>
+      <div data-testid="stream-status">{JSON.stringify(streamStatus)}</div>
+      <div data-testid="key-status">{JSON.stringify(status)}</div>
+    </div>
+  );
+};
+
+describe("TamboPropStreamProvider", () => {
+  describe("Hook Error Handling", () => {
+    it("should throw error when useTamboStream is used outside provider", () => {
+      // Suppress console.error for this test
+      const originalError = console.error;
+      console.error = jest.fn();
+
+      expect(() => {
+        render(<TestHookComponent />);
+      }).toThrow(
+        "useTamboStream must be used within a TamboPropStreamProvider",
+      );
+
+      console.error = originalError;
+    });
+  });
+
+  describe("Basic Functionality", () => {
+    it("should provide data and stream status through context", () => {
+      const testData = { message: "Hello World" };
+      const testStreamStatus: StreamStatus = {
+        isPending: false,
+        isStreaming: false,
+        isSuccess: true,
+        isError: false,
+        streamError: undefined,
+      };
+
+      render(
+        <TamboPropStreamProvider
+          data={testData}
+          streamStatus={testStreamStatus}
+        >
+          <TestHookComponent />
+        </TamboPropStreamProvider>,
+      );
+
+      expect(screen.getByTestId("data")).toHaveTextContent(
+        JSON.stringify(testData),
+      );
+      expect(screen.getByTestId("stream-status")).toHaveTextContent(
+        JSON.stringify(testStreamStatus),
+      );
+    });
+
+    it("should use default stream status when none provided", () => {
+      const testData = { message: "Hello World" };
+      const expectedDefaultStatus = {
+        isPending: false,
+        isStreaming: false,
+        isSuccess: true,
+        isError: false,
+        streamError: undefined,
+      };
+
+      render(
+        <TamboPropStreamProvider data={testData}>
+          <TestHookComponent />
+        </TamboPropStreamProvider>,
+      );
+
+      expect(screen.getByTestId("stream-status")).toHaveTextContent(
+        JSON.stringify(expectedDefaultStatus),
+      );
+    });
+  });
+
+  describe("Compound Components", () => {
+    describe("Loading Component", () => {
+      it("should render loading when isPending is true", () => {
+        const streamStatus: StreamStatus = {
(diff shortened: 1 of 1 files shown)


MichaelMilstead:
Nice, looks like `Content` will render when the field is done streaming in, can we change the name to `Complete` or something similar?

Overall looks good!

react-sdk/src/providers/tambo-stream-provider.tsx
```diff
@@ -0,0 +1,298 @@
...
+export interface TamboStreamProviderProps {
+  /** The stream data */
+  data: any;
+  /** Optional stream status for more granular control */
+  streamStatus?: StreamStatus;
+}
+
+export interface LoadingProps {
+  /** The key to identify this loading state */
+  key?: string;
```
alecf: At a very practical level, you can't use `key` here
1) I'm pretty sure that the value passed for `key` will not actually show up in your props, as it is reserved for react's reconciliation algorithm 
2) Your sample code passes multiple children with the same `key` prop, which is definitely not allowed. 

At the very lease this needs a new name, like `propKey` or something
https://legacy.reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings

alecf (commented):
Overall I do like this approach! I'm a little wary of the generic name `TamboStreamContext` etc - I feel like this is very specifically for streaming in props and there are other aspects to streaming. 

What about `TamboStreamPropsProvider` or something? I know it is kind of a mouthful...

react-sdk/src/providers/tambo-stream-provider.tsx
aarmful: totally, this was a very quick proof of concept for the team.

i missed that oversight while making the example.

i'm expecting a lot to change according to Tambo requirements

aarmful:
> Nice, looks like `Content` will render when the field is done streaming in, can we change the name to `Complete` or something similar?
> 
> Overall looks good!

`Complete` makes sense as well!

> Overall I do like this approach! I'm a little wary of the generic name `TamboStreamContext` etc - I feel like this is very specifically for streaming in props and there are other aspects to streaming.
> 
> What about `TamboStreamPropsProvider` or something? I know it is kind of a mouthful...

i understand your concern, we can tweak this to make more sense. this was mostly to get the idea across to Michael as i showed him a quick example during todays stream.

michaelmagan:
@alecf this is good feedback on naming. 

I like `TamboPropStreamProvider`.


My feedback is mainly around naming. I think we should keep naming to closely match the hook.

I'm also wondering if we then provide a flexible `When` for when you want to show something during multiple state.

Gate/Wrapper | Renders when status matches… | Typical “OR” combos
-- | -- | --
Pending | isPending | –
Streaming | isStreaming | –
Success | isSuccess and data present | –
Empty | not isPending, isStreaming, isSuccess, isError and data empty | –
Error | isError | –
When any={[…]} | Current state ∈ passed array | e.g. ["streaming","success"], ["pending","streaming","success"]


**I no longer like this:**
```
import { TamboPropStreamProvider } from "@tambo/react";

export default function Note({ title, content, createdAt }) {
  return (
    <TamboPropStreamProvider data={{ title, content, createdAt }}>
      <h3>{title}</h3>
      <p>{content}</p>

      {/* single-state helper */}
      <TamboPropStreamProvider.Pending   prop="createdAt">
        <p>Thinking…</p>
      </TamboPropStream.Pending>

      <TamboPropStreamProvider.Streaming prop="createdAt">
        <p>⏳ streaming …</p>
      </TamboPropStreamProvider.Streaming>
      
      {/* multi-state gate */}
      <TamboPropStreamProvider.When prop="createdAt" any={["pending","streaming","success"]}>
        Show while pending through success: {createdAt}
      </TamboPropStreamProvider.When>

      <TamboPropStreamProvider.Success   prop="createdAt">
        <p>Only Show after Finished Streaming: {createdAt}</p>
      </TamboPropStreamProvider.Success>

      <TamboPropStreamProvider.Empty     prop="createdAt">
        <p>No creation date.</p>
      </TamboPropStreamProvider.Empty>

      <TamboPropStreamProvider.Error     prop="createdAt">
        <p>⚠️ failed to generate date.</p>
      </TamboPropStreamProvider.Error>


    </TamboPropStreamProvider>
  );
}
```

On the flip side `When` is a little word but it gives you very clear granular control.

michaelmagan:
i actually kind of like this:

```
// Note.tsx – gates one prop, streams the rest inline
import {
  TamboPropStreamProvider as Stream,
  StreamStatus,
} from "@tambo/react";

export default function Note({ title, content, createdAt }: NoteProps) {

  return (
    <Stream data={{ title, content, createdAt }}>
      <h3>{title}</h3>
      <p>{content}</p>

      {/* gate createdAt with OR logic */}
      <Stream.When
        prop="createdAt"
        is={[StreamStatus.Pending, StreamStatus.Streaming]}
      >
        <p>⏳ fetching date…</p>
      </Stream.When>

      <Stream.When prop="createdAt" is={StreamStatus.Success}>
        <p>Created: {createdAt}</p>
      </Stream.When>

      <Stream.When prop="createdAt" is={StreamStatus.Empty}>
        <p>No creation date.</p>
      </Stream.When>

      <Stream.When prop="createdAt" is={StreamStatus.Error}>
        <p>⚠️ failed to generate date.</p>
      </Stream.When>
    </Stream>
  );
}

```

michaelmagan:
Hey we synced and this is all we decided on:

`TamboPropStreamProvider`

`content` renamed to `complete`.

Then we can merge.

Thanks @armfuldev!

michaelmagan:
@armfuldev went to test.  this new provider is not exported in the package so it can be used by a developer who imported the packaged. you will need to export it in ` react-sdk/src/providers/index.ts`

Issue #675, opened 2025-07-17
Interactive Map component with react‑leaflet


michaelmagan:
### Overview

Create a Generative UI component for an Interactive Map in tambo with the following features:

* **Base tiles** with pan and zoom functionality.
* Accepts:
  * `markers[]` with `{lat, lng, label}`
  * Optional `heatData[]`
* Tooltip displays on marker hover.
* Clustering of markers when zoomed out.
* Built with **react-leaflet**.

Requirements

* Use zod for defining the components props.
* Export the zod object from the component.
* Use `useTamboComponentState` for anything interactive.
* UseTamboStreamingProps or a useEffect for updating the props while rendering.
* useTamboStreamStatus and TamboPropStreamProvider

Here is an example: [https://github.com/tambo-ai/tambo/blob/main/cli/src/registry/graph/graph.tsx](https://github.com/tambo-ai/tambo/blob/main/cli/src/registry/graph/graph.tsx)

<sub>From [SyncLinear.com](https://synclinear.com) | [TAM-322](https://linear.app/tambo/issue/TAM-322/interactive-map-component-with-react‑leaflet)</sub>

MichaelMilstead:
completed with:
https://github.com/tambo-ai/tambo/pull/678

michaelmagan:
Congrats @revogabe!!!

PR #678, opened 2025-07-21
feat: Add InteractiveMap component using react‑leaflet

revogabe:
### Description

Adds a new `InteractiveMap` component to address issue #675. The component:

* Renders an interactive map with pan and zoom via react‑leaflet
* Accepts props:
  * `markers: Array<{ lat: number; lng: number; label: string }>`
  * Optional `heatData: Array<[number, number, number]>`
* Displays tooltips on marker hover
* Supports marker clustering when zoomed out
* Uses `zod` schema for props validation
* Integrates Tambo-specific streaming via `useTamboComponentState`, `useTamboStreamingProps`, `useTamboStreamStatus`, and `TamboPropStreamProvider`
* Structured similarly to the existing `graph.tsx` example
* Style Variations

---

### Related Issue

* \#675 – interactive-map component

---

### Preview
https://github.com/user-attachments/assets/e5a2191e-a6b1-4871-865b-dd796a15308d

---

### Next Steps

* Fix re-render component
* Allow custom cluster icons
* Optimize performance for large datasets

### DIFF
--- a/showcase/src/components/generative/MapChatInterface.tsx
+++ b/showcase/src/components/generative/MapChatInterface.tsx
@@ -5,28 +5,16 @@ import { useUserContextKey } from "@/lib/useUserContextKey";
 import { useTambo } from "@tambo-ai/react";
 import { useEffect } from "react";
 
-/*
- * We use `dynamic` to load the Map component without server-side rendering (SSR)
- * because `react-leaflet` and `leaflet` access the `window` object at import time.
- */
-
-import dynamic from "next/dynamic";
-
-const MapWithNoSSR = dynamic(
-  async () => await import("@/components/ui/map").then((mod) => mod.Map),
-  {
-    ssr: false,
-  },
-);
-
 export const MapChatInterface = () => {
   const userContextKey = useUserContextKey("map-thread");
   const { registerComponent } = useTambo();
 
   useEffect(() => {
     const register = async () => {
+      /* Dynamically import the Map component and its schema */
       const mod = await import("@/components/ui/map");
       const mapSchema = mod.mapSchema;
+      const Map = mod.Map;
 
       registerComponent({
         name: "Map",
@@ -50,7 +38,7 @@ export const MapChatInterface = () => {
         - Choose zoom level appropriate for your data
         
       Example use cases: store locations, real estate, events, analytics, travel, business intelligence.`,
-        component: MapWithNoSSR,
+        component: Map,
         propsSchema: mapSchema,
       });
     };


michaelmagan:
The demo looks great! I wonder if we can show the map after the first pin is generated, and then upon each complete of a full pin we show it to to the graph?

I'm working on something similar for our graph component right now. We an swap notes tomorrow.

cli/src/registry/map/map.tsx
```diff
@@ -0,0 +1,413 @@
...
+    click: (e: { latlng: L.LatLng; target: L.Map }) => {
+      const map: L.Map = e.target;
+      map.setView(e.latlng, map.getZoom(), { animate: animateRef.current });
+    },
+  });
+  return null;
+}
+
+// --- Map Component ---
+export const MapComponent = React.forwardRef<HTMLDivElement, MapProps>(
```
MichaelMilstead: maybe we rename this to just `Map` instead of `MapComponent`

showcase/src/components/generative/MapChatInterface.tsx
```diff
@@ -0,0 +1,369 @@
+"use client";
+
+import { MessageThreadFull } from "@/components/ui/message-thread-full";
+import { useUserContextKey } from "@/lib/useUserContextKey";
+import { useTambo } from "@tambo-ai/react";
+import { useEffect } from "react";
+
+import dynamic from "next/dynamic";
+
+const MapWithNoSSR = dynamic(
```
MichaelMilstead: I think we can remove this and just import the map component directly

showcase/src/components/generative/MapChatInterface.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+Best practices:
+- Use valid coordinates (lat: -90 to 90, lng: -180 to 180)
+- Provide clear labels for markers
+- Choose zoom level appropriate for your data
+
+Example use cases: store locations, real estate, events, analytics, travel, business intelligence.
+`,
+      component: MapWithNoSSR,
+      propsDefinition: {
```
MichaelMilstead: instead of `propsDefinition`, let's use `propsSchema: mapSchema` so we don't have to define this JSON object. This `propsDefinition` is an older pattern we were using (noticing now that the other components are using `propsDefinition`, we can separately update the other components' registrations)

cli/src/registry/map/map.tsx
```diff
@@ -0,0 +1,413 @@
...
+  (
+    {
+      center,
+      zoom = 10,
```
MichaelMilstead, line 278: maybe something for a later PR: Wonder if we can calculate and auto-set the zoom to fit all the markers. Seems like sometimes Tambo does this on its own, but other times shows the map too zoomed-in to see all the markers

revogabe: yes, maybe I can do that, but I think I should address this issue in another PR

showcase/src/components/generative/MapChatInterface.tsx
revogabe: react-leaflet doesn’t have good support for Next.js, and since it accesses the browser’s `window` object, we need to tell Next not to use SSR for this component. 

curious!, why don’t you want to use `dynamic`? 👀

revogabe: I noticed that right now even `dynamic` doesn’t work because I’m importing `mapSchema` to use in `propsSchema`, and when I import it from inside the file, it runs into the `window` error again.

revogabe: we need to find a solution. either without using `dynamic`, or using `dynamic` but making sure that importing `mapSchema` doesn’t trigger the same window error.

alecf: I can vouch here that leaflet + nextjs don't play together - I had to use `import()` (lower-level `dynamic()`) to get leaflet to work in a recent project - I think @MichaelMilstead probably didn't know about this issue. (maybe just add a comment explaining this, e.g. "We need to use dynamic() because leaflet doesn't play well with RSC in nextjs" or something

revogabe: yeah, yes, exactly but now it runs into another problem even using `dynamic`.

> I noticed that right now even `dynamic` doesn’t work because I’m importing `mapSchema` to use in `propsSchema`, and when I import it from inside the file, it runs into the `window` error again.

MichaelMilstead: hmm, I thought the map component would be ok since we have "use client". What's the error you're seeing?

revogabe: `use client` only marks the file as a Client Component, but Next still processes the imports during SSR. Since `react-leaflet` uses `window` at import time, it breaks on the server. That’s why `dynamic` is needed to make sure it only runs on the client.

revogabe:
In my last commits I made some adjustments and we’re no longer seeing any errors. This is a temporary solution to handle Next.js, and we can look for a better way in the future to avoid this split between server-side and client-side.

showcase/src/components/generative/MapChatInterface.tsx
MichaelMilstead: gotcha!

MichaelMilstead:
looks like we have a `ts-node` dependency in package-lock that's needed for running tests, but it was removed from a package.json a while ago, so the new package-lock here doesn't contain it.

To get this PR through, how about we remove the package-lock changes and we can clean that up later

showcase/package.json
```diff
@@ -20,12 +21,17 @@
...
+    "leaflet": "^1.9.4",
+    "leaflet.heat": "^0.2.0",
+    "leaflet.markercluster": "^1.5.3",
     "lucide-react": "^0.525.0",
     "next": "^15.4.1",
     "next-themes": "^0.4.6",
     "radix-ui": "^1.4.2",
-    "react": "^18.3.1",
-    "react-dom": "^18.3.1",
+    "react": "^19.1.0",
```
alecf: this PR is bumping react to 19. Can you leave that at 18 please?

alecf: Looking further: it looks like react-leaflet v5 requires react 19, so lets use react-leaflet v4 for now

If we upgrade this to react 19, it has a cascading effect on a number of other projects that we'd like to avoid handling right now

cli/CHANGELOG.md
```diff
@@ -2,45 +2,38 @@
 
 ## [0.22.0](https://github.com/tambo-ai/tambo/compare/tambo-v0.21.1...tambo-v0.22.0) (2025-07-15)
 
-
```
alecf: can you also avoid editing this file? it is managed by a script and I'm concerned that this is going to break this script

showcase/package.json
revogabe: That’s exactly why I had upgraded to React 19, but it makes sense. let’s go back to React 18 and use react-leaflet v4.

alecf:
so in order to seperate out these issues, I just landed https://github.com/tambo-ai/tambo/pull/703 which adds back ts-node. It *seems* like someone deleted the package-lock.json and then re-ran `npm install` which is generally not a good pattern (it can cause all sorts of sub-packages to update that you didn't intend)

My request for the above is:
1) restore the original package-lock.json + various package.json files
2) `npm ci` to make sure you get a clean package install
3) `npm install` all the various leaflet libraries at the correct versions

that should make sure only the minimal changes are made to the various package*.json files.

revogabe:
I'm running into this same issue again after reverting back to react-leaflet v4. I remember this was the main reason I originally moved to v5. I'm trying to find a workaround, but so far nothing seems to work consistently.


image

https://github.com/PaulLeCam/react-leaflet/issues/936
https://github.com/PaulLeCam/react-leaflet/pull/1073

cc: @alecf

revogabe:
> so in order to seperate out these issues, I just landed #703 which adds back ts-node. It _seems_ like someone deleted the package-lock.json and then re-ran `npm install` which is generally not a good pattern (it can cause all sorts of sub-packages to update that you didn't intend)
> 
> My request for the above is:
> 
> 1. restore the original package-lock.json + various package.json files
> 2. `npm ci` to make sure you get a clean package install
> 3. `npm install` all the various leaflet libraries at the correct versions
> 
> that should make sure only the minimal changes are made to the various package*.json files.


done, but I’m still working on the issue I mentioned above.

alecf:
@revogabe I found a few fixes - and I turned off strict mode for the showcase to let the map run - turns out strict mode for development is what is causing that map container error. As much as I'd like to use strict mode, turning it off is preferred to upgrading to react 19.

Please see my changes here:
https://github.com/tambo-ai/tambo/commits/interactive-map/

revogabe:
Nice! this works well and is stable for merged, we can work on improvements later. Thanks for the help @alecf

alecf (approved):

PR #702, opened 2025-07-24
feat: add additionalContext support

akhileshrangani4:


react-sdk/src/providers/tambo-thread-provider.tsx
```diff
@@ -255,20 +256,26 @@ export const TamboThreadProvider: React.FC<
...
+
+      // Combine system context with user context
+      const combinedContext = {
+        system: message.role === "user" ? getSystemContext() : undefined,
+        custom: message.additionalContext,
+      };
+
       const chatMessage: TamboThreadMessage & {
-        additionalContext?: string;
+        additionalContext?: object;
```
alecf: similar to the other pr, `Record<string, any>` is generally preferred

https://www.reddit.com/r/typescript/comments/tq3m4f/the_difference_between_object_and_recordstring/

The TL;DR is that functions and arrays are also of type `object` so generally `object` is considered overly broad in almost every context

akhileshrangani4: done with this [commit](https://github.com/tambo-ai/tambo/pull/702/commits/dc608191b7c116e9168ac129f918b95935c4a8af)

PR #816, opened 2025-08-11
fix: split out provider values to prevent re-renders

MichaelMilstead:
Splits `generationStage` out of the ThreadContext into its own context/provider so a consumer can depend on it without re-rendering everytime the ThreadProvider's `thread` object updates.

Updates `useCurrentMessage` to avoid depending on the ThreadProvider's `thread` state by updating its provider props to take in a message rather than a messageId which it then needs to search the thread for.

Updates `useTamboThreadInput` to keep its state local, rather than updating a value within the ThreadProvider, causing re-renders of any ThreadProvider consumers on each typed character.


react-sdk/src/providers/tambo-thread-provider.tsx
```diff
@@ -28,6 +28,44 @@ import { useTamboClient } from "./tambo-client-provider";
...
+
+interface GenerationStageProviderProps {
+  generationStage: GenerationStage;
+  statusMessage: string;
+}
+
+const GenerationStageProvider: React.FC<
+  PropsWithChildren<GenerationStageProviderProps>
+> = ({ children, generationStage, statusMessage }) => {
+  const isIdle = useMemo(() => isIdleStage(generationStage), [generationStage]);
```
alecf: this useMemo is unnecessary and will cause extra renders - isIdle is a primitive, this should just be:
```suggestion
  const isIdle = isIdleStage(generationStage);
```

react-sdk/src/providers/tambo-thread-provider.tsx
```diff
@@ -143,11 +175,11 @@ export const TamboThreadContext = createContext<TamboThreadContextProps>({
    *
    */
   sendThreadMessage: () => {
-    throw new Error("advance not implemented");
+    throw new Error("sendThreadMessage not implemented");
   },
-  generationStage: GenerationStage.IDLE,
-  generationStatusMessage: "",
-  isIdle: true,
+  /**
```
alecf, line 181: I don't see how leaving this out of this provider is going to change the re-rendering?

thanks to the final useMemo() in GenerationStageProvider, these values won't change

PR #868, opened 2025-08-20
feat(api): stop using contextKey returned from API

alecf:
- **bump to latest typescript-sdk**
- **stop using contextKey from return value**

### DIFF
--- a/cli/src/registry/map/map.tsx
+++ b/cli/src/registry/map/map.tsx
@@ -292,7 +292,7 @@ export const Map = React.forwardRef<HTMLDivElement, MapProps>(
 
     const message = thread?.messages[thread?.messages.length - 1];
 
-    const isLatestMessage = message?.id === currentMessage.id;
+    const isLatestMessage = message?.id && message.id === currentMessage?.id;
 
     const generationStage = thread?.generationStage;
     const isGenerating =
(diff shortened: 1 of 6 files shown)


cli/src/registry/map/map.tsx
```diff
@@ -292,7 +292,7 @@ export const Map = React.forwardRef<HTMLDivElement, MapProps>(
 
     const message = thread?.messages[thread?.messages.length - 1];
 
-    const isLatestMessage = message?.id === currentMessage.id;
+    const isLatestMessage = message?.id && message.id === currentMessage?.id;
```
alecf, line 295: this was throwing a type error, not sure why it wasn't blowing up in CI ?

react-sdk/src/providers/tambo-thread-provider.tsx
```diff
@@ -743,7 +741,7 @@ export const TamboThreadProvider: React.FC<
         threadId = currentThreadId ?? PLACEHOLDER_THREAD.id,
         streamResponse = streaming,
         forceToolChoice,
-        contextKey = currentThreadContextKey,
+        contextKey,
```
alecf, line 744: this basically means the user has to pass in contextKey themselves, but that was kind of the design of the API to begin with anyway

PR #873, opened 2025-08-20
docs(hackathon): template ideas

michaelmagan:


michaelmagan:
@CharlieHelps  fix the review issues

PR #876, opened 2025-08-20
fix: create context/provider for threadInput so all consumers share state

MichaelMilstead:
The MessageInput component and useTamboSuggestions hook previously had two separate instances of input state, so accepted suggestions would not be seen in the MessageInput.

Creates a context/provider for thread input so that all consumers of `useTamboThreadInput` within the TamboProvider share the same state

### DIFF
--- a/react-sdk/src/providers/tambo-thread-input-provider.tsx
+++ b/react-sdk/src/providers/tambo-thread-input-provider.tsx
@@ -58,6 +58,8 @@ export interface TamboThreadInputProviderProps {
  * Provider that manages shared thread input state across all components
  * This ensures that useTamboThreadInput, useTamboSuggestions, and components
  * all share the same input state
+ * @param contextKey - Optional context key.
+ * @returns The thread input context
  */
 export const TamboThreadInputProvider: React.FC<
   PropsWithChildren<TamboThreadInputProviderProps>
@@ -125,6 +127,7 @@ export const TamboThreadInputProvider: React.FC<
 /**
  * Hook to access the shared thread input state
  * @param contextKey - Optional context key that overrides the provider's contextKey for this specific usage, for backwards compatibility.
+ * @returns The thread input context
  */
 export const useTamboThreadInput = (contextKey?: string) => {
   const context = useContext(TamboThreadInputContext);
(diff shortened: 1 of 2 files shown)


react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -91,20 +96,55 @@ export function useTamboThreadInput(contextKey?: string): UseThreadInput {
...
     },
-    [inputValue, sendThreadMessage, thread.id, contextKey, setInputValue],
+    [inputValue, sendThreadMessage, thread.id, contextKey],
   );
+
   const {
     mutateAsync: submitAsync,
-    mutate: _unusedSubmit,
-    ...mutationState
+    isPending,
```
alecf, line 104: Why did you drop so much of the mutation state? now users can't access the other props

react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -91,20 +96,55 @@ export function useTamboThreadInput(contextKey?: string): UseThreadInput {
...
+    throw new Error(
+      "useTamboThreadInput must be used within a TamboThreadInputProvider",
+    );
+  }
+
+  // If a contextKey is provided to the hook, create a wrapped submit function
+  if (contextKey) {
+    return {
+      ...context,
+      submit: async (options = {}) =>
```
alecf, line 144: the problem here is that this is no longer memoized, so submit will be recreated on every render, and this can cause downstream components to rerender a lot

react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -1,32 +1,27 @@
...
 import { validateInput } from "../model/validate-input";
-import { useTamboThread } from "../providers/tambo-thread-provider";
-import { useTamboMutation } from "./react-query-hooks";
+import { useTamboThread } from "./tambo-thread-provider";
 
 /**
  * Error messages for various input-related error scenarios
- * These messages are used to provide user-friendly error feedback
- * @readonly
  */
```
alecf, line 16: minor, but why remove the docs?

react-sdk/src/providers/tambo-thread-input-provider.tsx
MichaelMilstead: ah whoops! Will add the rest back

react-sdk/src/providers/tambo-thread-input-provider.tsx
MichaelMilstead: Hmm I think I meant to just remove the text and not the jsdoc tag, thinking that the text might have been reduntant with the line above, but I think the text might be helpful so I can add both lines back in

Issue #840, opened 2025-08-15
Persist unsent message drafts across thread switch/refresh


MichaelMilstead:
## Problem

Users of tambo apps with multiple threads may want to draft a message for a thread and finish it after taking some other actions.
Currently, if a user types a message in the chat input field and then:

1. switches to a different thread, the same message draft is retained. When switching to a new thread, the message input should be cleared, but show my draft again when I switch back to the original thread.
2. performs a page refresh, any draft message is lost.

## Solution

Use localStorage to store a local cache of message drafts per thread

## Acceptance Criteria

- Switching threads and returning shows the previously typed draft.

- A page refresh retains any drafts.

- Sending a message removes the draft from localStorage.

- No cross-pollination: Drafts from other threads must not appear in the wrong thread.

michaelmagan:
Closing for now as we focus on 1.0 launch.

PR #934, opened 2025-08-27
feat: persist unsent message drafts across thread switch/refresh

Aniket-508:
Solves #840

### DIFF
--- a/react-sdk/src/providers/tambo-thread-input-provider.tsx
+++ b/react-sdk/src/providers/tambo-thread-input-provider.tsx
@@ -82,7 +82,22 @@ export const TamboThreadInputProvider: React.FC<
   PropsWithChildren<TamboThreadInputProviderProps>
 > = ({ children, contextKey }) => {
   const { thread, sendThreadMessage } = useTamboThread();
-  const [inputValue, setInputValue] = useState("");
+  const { saveDraft, clearDraft, getDraft } = useMessageDraft(
+    thread?.id ?? "null_thread",
+  );
+  const [inputValue, setInputValue] = useState(getDraft());
+
+  React.useEffect(() => {
+    setInputValue(getDraft());
+  }, [getDraft]);
+
+  const handleSetInputValue = useCallback(
+    (value: string) => {
+      setInputValue(value);
+      saveDraft(value);
+    },
+    [saveDraft],
+  );
 
   const submit = useCallback(
     async ({
@@ -112,8 +127,9 @@ export const TamboThreadInputProvider: React.FC<
         additionalContext: additionalContext,
       });
       setInputValue(""); // Clear local state
+      clearDraft();
     },
-    [inputValue, sendThreadMessage, thread.id, contextKey],
+    [inputValue, sendThreadMessage, thread.id, contextKey, clearDraft],
   );
 
   const {
@@ -127,7 +143,7 @@ export const TamboThreadInputProvider: React.FC<
   const value = {
     ...mutationState,
     value: inputValue,
-    setValue: setInputValue,
+    setValue: handleSetInputValue,
     submit: submitAsync,
   };
 
@@ -153,3 +169,68 @@ export const useTamboThreadInput = () => {
 
   return context;
 };
+
+const DRAFT_KEY = "tambo-user-input";
+
+interface Draft {
+  id: string;
+  content: string;
+  timestamp: number;
+}
+
+/**
+ * Custom hook to manage the draft of a message input for a thread.
+ * Persists the draft to localStorage.
+ * @param threadId The ID of the thread for this draft
+ * @returns An object with `saveDraft`, `clearDraft`, and `getDraft` functions
+ */
+function useMessageDraft(threadId: string | null) {
+  const getDrafts = React.useCallback((): Draft[] => {
+    if (typeof window === "undefined") {
+      return [];
+    }
+    const drafts = localStorage.getItem(DRAFT_KEY);
+    return drafts ? JSON.parse(drafts) : [];
+  }, []);
+
+  const getDraft = React.useCallback(() => {
+    if (typeof window === "undefined" || !threadId) {
+      return "";
+    }
+    const drafts = getDrafts();
+    const draft = drafts.find((d) => d.id === threadId);
+    return draft ? draft.content : "";
+  }, [getDrafts, threadId]);
+
+  const clearDraft = React.useCallback(() => {
+    if (typeof window === "undefined" || !threadId) {
+      return;
+    }
+    const drafts = getDrafts();
+    const otherDrafts = drafts.filter((d) => d.id !== threadId);
+    localStorage.setItem(DRAFT_KEY, JSON.stringify(otherDrafts));
+  }, [getDrafts, threadId]);
+
+  const saveDraft = React.useCallback(
+    (content: string) => {
+      if (typeof window === "undefined" || !threadId) {
+        return;
+      }
+
+      if (content === "") {
+        clearDraft();
+        return;
+      }
+
+      const drafts = getDrafts();
+      const now = Date.now();
+      const newDraft: Draft = { id: threadId, content, timestamp: now };
+      const otherDrafts = drafts.filter((d) => d.id !== threadId);
+      const updatedDrafts = [...otherDrafts, newDraft];
+      localStorage.setItem(DRAFT_KEY, JSON.stringify(updatedDrafts));
+    },
+    [getDrafts, threadId, clearDraft],
+  );
+
+  return { saveDraft, clearDraft, getDraft };
+}


react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -153,3 +169,68 @@ export const useTamboThreadInput = () => {
...
+ * @param threadId The ID of the thread for this draft
+ * @returns An object with `saveDraft`, `clearDraft`, and `getDraft` functions
+ */
+function useMessageDraft(threadId: string | null) {
+  const getDrafts = React.useCallback((): Draft[] => {
+    if (typeof window === "undefined") {
+      return [];
+    }
+    const drafts = localStorage.getItem(DRAFT_KEY);
+    return drafts ? JSON.parse(drafts) : [];
```
michaelmagan: This should have a try-catch loop.

react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -153,3 +169,68 @@ export const useTamboThreadInput = () => {
...
+    return draft ? draft.content : "";
+  }, [getDrafts, threadId]);
+
+  const clearDraft = React.useCallback(() => {
+    if (typeof window === "undefined" || !threadId) {
+      return;
+    }
+    const drafts = getDrafts();
+    const otherDrafts = drafts.filter((d) => d.id !== threadId);
+    localStorage.setItem(DRAFT_KEY, JSON.stringify(otherDrafts));
```
michaelmagan: setItem can sometimes fail; we should add a try-catch block here too.

react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -153,3 +169,68 @@ export const useTamboThreadInput = () => {
...
+        clearDraft();
+        return;
+      }
+
+      const drafts = getDrafts();
+      const now = Date.now();
+      const newDraft: Draft = { id: threadId, content, timestamp: now };
+      const otherDrafts = drafts.filter((d) => d.id !== threadId);
+      const updatedDrafts = [...otherDrafts, newDraft];
+      localStorage.setItem(DRAFT_KEY, JSON.stringify(updatedDrafts));
```
michaelmagan: Same here.

react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -153,3 +169,68 @@ export const useTamboThreadInput = () => {
 
   return context;
 };
+
+const DRAFT_KEY = "tambo-user-input";
```
michaelmagan: This key should be set such that it's scoped to the developers app; maybe the last 20 letters of the api key.

react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -82,7 +82,22 @@ export const TamboThreadInputProvider: React.FC<
...
+  const { saveDraft, clearDraft, getDraft } = useMessageDraft(
+    thread?.id ?? "null_thread",
+  );
+  const [inputValue, setInputValue] = useState(getDraft());
+
+  React.useEffect(() => {
+    setInputValue(getDraft());
+  }, [getDraft]);
+
+  const handleSetInputValue = useCallback(
```
michaelmagan, line 108: Add a 300ms debounce. localStorage is synchronous.

michaelmagan (changes requested):
Hey, overall looks good. 

The main things that are missing:

1. Update Docs: Short note in the “Threads” or “Input” provider page describing persisted drafts and how to clear them programmatically (e.g., “Clear all drafts” UX). Make sure to add a note this is stored in local storage.
2. Tests: It's missing tests like corrupted JSON, or SSR window guards.
3. Every new React feature should update our component library and showcase site.

Otherwise, this is a clean PR and solves the issue. Great job.

react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -82,7 +82,22 @@ export const TamboThreadInputProvider: React.FC<
   PropsWithChildren<TamboThreadInputProviderProps>
 > = ({ children, contextKey }) => {
   const { thread, sendThreadMessage } = useTamboThread();
-  const [inputValue, setInputValue] = useState("");
+  const { saveDraft, clearDraft, getDraft } = useMessageDraft(
+    thread?.id ?? "null_thread",
+  );
+  const [inputValue, setInputValue] = useState(getDraft());
+
+  React.useEffect(() => {
```
MichaelMilstead: let's import and use `useEffect` directly rather than `React.useEffect` throughout

react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -153,3 +169,68 @@ export const useTamboThreadInput = () => {
...
+  timestamp: number;
+}
+
+/**
+ * Custom hook to manage the draft of a message input for a thread.
+ * Persists the draft to localStorage.
+ * @param threadId The ID of the thread for this draft
+ * @returns An object with `saveDraft`, `clearDraft`, and `getDraft` functions
+ */
+function useMessageDraft(threadId: string | null) {
```
MichaelMilstead: let's move this new hook (including the DRAFT_KEY and Draft interface) under `/providers/hooks/` in a new file like `use-message-draft.ts`  just to try to keep the provider files simple

react-sdk/src/providers/tambo-thread-input-provider.tsx
MichaelMilstead: mm this may be fine since localstorage is scoped to the domain, but maybe we change the name to something that makes it clear it can contain multiple input drafts like "tambo-message-drafts"?

react-sdk/src/providers/tambo-thread-input-provider.tsx
alecf: also if a callback has no dependency, then it should just be a module level function. `getDrafts` does not need to reside inside the component.

react-sdk/src/providers/tambo-thread-input-provider.tsx
alecf: Let me offer an alternate approach:
- save the "current" draft for the current thread in a useState() (e.g. `[currentDraft, setCurrentDraft]`
- add an internal function `persistDraft(threadId, content)` that is debounced (see `useDebouncedCallback`) that wraps the `setItem` call - i.e. its only purpose is to update localStorage for the specified thread.
- use a useEffect() to update local storage off the main event handler (keeping the work off the event handler critical path)
- the saveDraft() function should call both `setCurrentDraft` but *not* `persistDraft`, because that will happen asynchronously
- Rather than exposing a function `getDraft`, instead just return `currentDraft`

The advantage of this approach is not just that it keeps storage APIs off the main "thread" (i.e. synchronous event handler path), but also it exposes the current, non-debounced draft, without the consumer of the hook being required to call a function (e.g. consumer uses `currentDraft` rather than `getDraft()`

michaelmagan: much better.

react-sdk/src/providers/tambo-thread-input-provider.tsx
michaelmagan: good point.  I agree with the more clear name.

react-sdk/src/providers/tambo-thread-input-provider.tsx
Aniket-508: @michaelmagan @alecf i have updated these changes.

react-sdk/src/providers/hooks/use-tambo-message-drafts.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  // Load initial draft for the thread
+  useEffect(() => {
+    isInitialLoad.current = true; // Set flag on threadId change
+    if (typeof window === "undefined" || !threadId) {
+      setCurrentDraft("");
+      return;
+    }
+
+    const drafts = getDrafts();
```
alecf: can we make the DRAFT_KEY parameterizable? pass it to useTamboMessageDrafts (defaulting to DRAFT_KEY) and then down into getDrafts() here

react-sdk/src/providers/hooks/use-tambo-message-drafts.ts
```diff
@@ -0,0 +1,106 @@
...
+    const draft = drafts.find((d) => d.id === threadId);
+    setCurrentDraft(draft ? draft.content : "");
```
alecf: very minor but this is a case for nullish coalescing! 
```suggestion
    setCurrentDraft(draft?.content ?? "");
```

react-sdk/src/providers/hooks/use-tambo-message-drafts.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (threadId) {
+      persistDraft(threadId, currentDraft);
+    }
+  }, [threadId, currentDraft, persistDraft]);
+
+  const saveDraft = useCallback((content: string) => {
+    setCurrentDraft(content);
+  }, []);
+
+  return { currentDraft, saveDraft };
```
alecf: minor but at this point I'd say you can just return setCurrentDraft directly here, no need for extra callbacks:
```suggestion

  return { currentDraft, saveDraft: setCurrentDraft };
```

react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -111,7 +98,7 @@ export const TamboThreadInputProvider: React.FC<
       forceToolChoice?: string;
       additionalContext?: Record<string, any>;
     } = {}) => {
-      const validation = validateInput(inputValue);
+      const validation = validateInput(currentDraft);
```
alecf: we're validating the input, but in a delayed way - the user might have hit "submit" quickly before the debounced update completes.

in my mind the user shouldn't have to use useTamboMessageDrafts  directly unless they're explicitly manipulating drafts - instead this should be transparent with the current hooks, and it's just smart about saving and restoring drafts automatically

react-sdk/src/providers/tambo-thread-input-provider.tsx
```diff
@@ -142,8 +128,8 @@ export const TamboThreadInputProvider: React.FC<
 
   const value = {
     ...mutationState,
-    value: inputValue,
-    setValue: handleSetInputValue,
+    value: currentDraft,
```
alecf: Did you play with this? if you're exposing the currentDraft as the value, then the user's keystrokes are always going to be ahead, no?

alecf (changes requested):
Getting there! The point of separating the draft from the input is to maintain that low-latency interaction by using a `useState` for the actual value that appears in the input, but allowing drafts to save automatically in the background, so you can restore a draft when the input first shows up

\

alecf (changes requested):
Getting there! The point of separating the draft from the input is to maintain that low-latency interaction by using a `useState` for the actual value that appears in the input, but allowing drafts to save automatically in the background, so you can restore a draft when the input first shows up

\

michaelmagan:
@Aniket-508, can you make the final changes to this so we can merge and release it? I want to share it during the hackathon!

michaelmagan (approved):

michaelmagan:
@Aniket-508, we should update/add documentation for this on the doc site. One very release we want the docs to be in sync with the actual package. 

https://docs.tambo.co/api-reference/react-hooks <-- update the hooks
https://docs.tambo.co/concepts/message-threads/sending-messages <-- after adding context we should add `Saving Drafts`

Thanks!

react-sdk/src/providers/hooks/use-tambo-message-drafts.ts
```diff
@@ -0,0 +1,83 @@
...
+    } catch (e) {
+      console.error("Failed to persist draft", e);
+    }
+  }, 300);
+
+  // Persist draft to local storage on change
+  useEffect(() => {
+    persistDraft(currentDraft);
+  }, [persistDraft, currentDraft]);
+
```
MichaelMilstead, line 97: Seems like we're correctly updating the stored draft and `currentDraft` for threads when we change the message input value, but we're not updating the `currentDraft` state when switching threads, so the last updated value is what I see in the input no matter what thread I look at.

We might need a useEffect to update `currentDraft` whenever `threadId` changes:

```
  useEffect(() => {
    setCurrentDraft(
      getDrafts(storageKey).find((d) => d.id === threadId)?.content ?? "",
    );
  }, [threadId]);
```

(and then maybe we can split that draft finding logic out since it's used in two places now)

Aniket-508: @MichaelMilstead Yes, will update.

Aniket-508:
video referencing latest changes:

https://github.com/user-attachments/assets/bdb47dfa-fb4f-4d99-b01f-0e309416724a

alecf:
so the one thing that is possibly a problem is that in the showcase we have some strange behavior because all of the showcase components are sharing the same placeholder thread id (i.e. `"placeholder"`) - try typing in any of these fields: https://showcase-eg04c7z4r-tambo-ai.vercel.app/components/message-input - weird stuff happens with focus/scrolling

And if you switch pages, the "draft" value persists to other component showcases 

Can we make each showcase component use its own draft storage key?

michaelmagan:
@CharlieHelps  can you make the change @alecf  suggested?

michaelmagan:
@claude 

Can you fix this? so the one thing that is possibly a problem is that in the showcase we have some strange behavior because all of the showcase components are sharing the same placeholder thread id (i.e. "placeholder") - try typing in any of these fields: [showcase-eg04c7z4r-tambo-ai.vercel.app/components/message-input](https://showcase-eg04c7z4r-tambo-ai.vercel.app/components/message-input) - weird stuff happens with focus/scrolling

And if you switch pages, the "draft" value persists to other component showcases

Can we make each showcase component use its own draft storage key?

Issue #806, opened 2025-08-08
Per-page MDX export (.mdx)


michaelmagan:
## Goal
Allow `.../docs/<path>.mdx` to return MDX content for a page.

**Reference**: [\*.mdx](https://fumadocs.dev/docs/ui/llms#mdx)

## Tasks
- [ ] Add `app/llms.mdx/[[...slug]]/route.ts` to return `getLLMText(page)`
- [ ] Implement `generateStaticParams()`
- [ ] Add rewrite in `next.config.ts` from `/docs/:path*.mdx` → `/llms.mdx/:path*`

## Acceptance
- [ ] Any doc page supports the `.mdx` suffix
- [ ] 404 when page not found; 200 with content when found
- [ ] Works in local dev, build, and on Vercel

Issue #805, opened 2025-08-08
llms-full.txt endpoint


michaelmagan:
## Goal
Provide a single text stream of all docs for AI.

**Reference**: [llms-full.txt](https://fumadocs.dev/docs/ui/llms#llms-fulltxt)

## Tasks
- [ ] Add `app/llms-full.txt/route.ts` that joins `getLLMText(page)` for all pages
- [ ] Set `revalidate = false` and return `text/plain`
- [ ] Add basic error handling; consider streaming if content is large

## Acceptance
- [ ] Endpoint returns full concatenated docs content
- [ ] Reasonable performance on large sets (no timeouts/OOM)
- [ ] Stable across redeploys (cached-forever semantics)

michaelmagan:
Hey we are currently working on this issue in #935

MichaelMilstead:
completed: https://docs.tambo.co/llms-full.txt

PR #935, opened 2025-08-27
feat: added support for llms.txt/llm-full.txt along with .mdx support for individual routes in docs

Aniket-508:


docs/src/app/llms-full.txt/route.ts
```diff
@@ -0,0 +1,12 @@
+import { getLLMText } from "@/lib/get-llm-text";
+import { source } from "@/lib/source";
+
+// cached forever
+export const revalidate = false;
```
alecf: I think for safety (in case caches persist across deployments on vercel, which I think they do?) we can set this to `3600`

docs/src/app/llms.mdx/[[...slug]]/route.ts
```diff
@@ -0,0 +1,21 @@
+import { getLLMText } from "@/lib/get-llm-text";
```
Comment (alecf) [inline docs/src/app/llms.mdx/[[...slug]]/route.ts:1]: this route is neat but Is this referenced somewhere? like who links to `/llms.mdx/foo` ? is this just for debugging?

docs/src/app/llms.txt/route.ts
```diff
@@ -0,0 +1,28 @@
...
+    // First dir returns undefined which we are considering as main docs route
+    map.set(dir ?? "tambo-docs", list);
+  }
+
+  for (const [key, value] of map) {
+    scanned.push(`## ${key}`);
+    scanned.push(value.join("\n"));
+  }
+
+  return new Response(scanned.join("\n\n"));
```
alecf, line 27: Seems like we're joining a lot of potentially really big strings. Can I suggest we add a nice streaming utility somewhere so that we can just stream the array of strings? e.g.

```ts
export function makeReadableStream(chunks: string[]) {
  return new ReadableStream({
    async start(controller) {
      const encoder = new TextEncoder();
      for (const chunk of chunks) {
        controller.enqueue(encoder.encode(chunk));
      }
      controller.close();
    },
  });
}
```

then you can just

```ts
return NextResponse(makeReadableStream(scanned), ...);
```

docs/src/app/llms.mdx/[[...slug]]/route.ts
Comment (Aniket-508) [reply docs/src/app/llms.mdx/[[...slug]]/route.ts:1]: As per [this](https://fumadocs.dev/docs/ui/llms#mdx), people can append `.mdx` to a page to get its Markdown/MDX content. Required config is written in `next.config.mjs`.

docs/src/app/llms-full.txt/route.ts
Aniket-508: I followed the official [docs](https://fumadocs.dev/docs/ui/llms), but will update it for safety.

docs/src/app/llms.txt/route.ts
Aniket-508: I think since we are joining only `title`, `url` and `description` of each page, the resultant txt file is not that big.

MichaelMilstead:
looks good! I'm interested to try out our llms-full.txt to see how well it works as context, and whether we need to care about trimming or ordering in future

docs/src/app/llms.txt/route.ts
alecf: Ah maybe not here then, but elsewhere we are joining the content of every single page into a single string

docs/src/app/llms-full.txt/route.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export async function GET() {
+  const scan = source.getPages().map(getLLMText);
+  const scanned = await Promise.all(scan);
+
+  return new Response(scanned.join("\n\n"));
```
alecf: @Aniket-508 this is where we're joining the entire file into a single string

Aniket-508: @alecf i have added the streaming utility in this response.

alecf (approved):

docs/src/app/llms-full.txt/route.ts
MichaelMilstead: hmm seems like this may have broken the generated llms-full.txt 
image

michaelmagan:
@Aniket-508 looks like there might be some conflict between this PR and this PR #943

docs/src/app/llms-full.txt/route.ts
Aniket-508: @MichaelMilstead this looks fine to me as this is the content present in the docs. are you specifically asking about formatting?

Ref. Screenshot of `llms-full.txt`:
Screenshot 2025-08-29 at 8 32 15 AM

Aniket-508:
> @Aniket-508 looks like there might be some conflict between this PR and this PR #943

@michaelmagan yep, I think there might be conflict only in the `lib/get-llm-text.ts` in this pr and `lib/llm.ts` file in @nermalcat69's pr. we can merge this and then maybe he can remove that in his pr or i can match my file contents to his as well?

i also think packages `remark`, `remark-gfm`, `remark-mdx` in his pr are already transitive dependencies of `fumadocs-core` package so we need not add them separately to `package.json` file.

nermalcat69:
Hey @Aniket-508, Sureee I don't mind feel free to get your pr merged i'll fix the conflicts :)

docs/src/app/llms-full.txt/route.ts
michaelmagan: @Aniket-508, we think this issue is the response isn't returning a content type. Can you fix that?

Aniket-508: @michaelmagan added the content-type as `text/plain;charset=UTF-8` in response headers.

michaelmagan:
@Aniket-508, the change looks good. I'll merge it.

Aniket-508:
@michaelmagan @alecf just realized that this pr also addresses #805 & #806 issues, updated the same in description.

PR #959, opened 2025-08-30
chore: update upgrade command to filter known safe packages

akhileshrangani4:


akhileshrangani4 (commented):
updated the upgrade command to only update the packages that does not break things, @MichaelMilstead - could you please approve this so that I can merge this- I am not able to do so.

PR #993, opened 2025-09-05
feat: add ask to tambo in the ai actions dropdown

akhileshrangani4:


docs/src/app/layout.tsx
```diff
@@ -1,13 +1,12 @@
 import "@/app/global.css";
-import { TamboProvider } from "@tambo-ai/react";
 import { RootProvider } from "fumadocs-ui/provider";
 import { ThemeProvider } from "next-themes";
 import { Inter } from "next/font/google";
-import { components } from "@/lib/tambo";
-import { PostHogPageview, PHProvider } from "@/app/providers";
+import { PostHogPageview, PHProvider } from "@/providers/php-provider";
```
alecf: minor but can we call this `posthog-provider`? It really took me a second to figure out we were not adding PHP to our code :)

docs/src/providers/tambo-provider.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import {
+  currentPageContextHelper,
+  TamboProvider as TamboProviderBase,
+} from "@tambo-ai/react";
+import { components } from "@/lib/tambo";
+
+export function TamboProvider({ children }: { children: React.ReactNode }) {
+  return (
+    <TamboProviderBase
```
alecf: Naming also threw me off here - can we call this something like `TamboAppProvider` or `TamboRootProvider` or something, and then leave the name `TamboProvider` alone? (this also affects developer experience in an IDE  where you might want to auto-complete to `TamboProvider` and not know which one is the right one when they have the same name!)

docs/src/components/ai-actions/open-dropdown.tsx
```diff
@@ -3,24 +3,53 @@
...
+type LinkItem = {
+  name: string;
+  url: string;
+  icon: React.ComponentType;
+  description: string;
+  internal?: boolean;
+};
+
+const TamboLogo = () => (
+  <Image
```
alecf, line 23: when possible, use static imports:
https://nextjs.org/docs/pages/api-reference/components/image#responsive-images-with-a-static-export

(this also ensures if the image is small enough it will get bundled directly into the source)

docs/src/components/tambo/message-input.tsx
```diff
@@ -129,12 +131,16 @@ export interface MessageInputProps
...
+  (
+    { children, className, contextKey, initialQuery, variant, ...props },
+    ref,
+  ) => {
     const { value, setValue, submit, isPending, error } = useTamboThreadInput();
     const { cancel } = useTamboThread();
     const [displayValue, setDisplayValue] = React.useState("");
     const [submitError, setSubmitError] = React.useState<string | null>(null);
     const [isSubmitting, setIsSubmitting] = React.useState(false);
+    const [hasSetInitialQuery, setHasSetInitialQuery] = React.useState(false);
```
alecf, line 143: this again goes back to places where we're modifying our own components - is this a feature we should be adding to the open-source component?

docs/src/components/ai-actions/open-dropdown.tsx
```diff
@@ -71,8 +100,8 @@ export function OpenDropdown({ markdownUrl, githubUrl }: OpenDropdownProps) {
                   <a
```
alecf, line 101: so we're not using the nextjs `<Link>` element, which means this will cause a full reload even if we're just loading the current page. What happens if we use `<Link>`? does it avoid a reload?

docs/src/app/layout.tsx
akhileshrangani4: aah yes, that is a little bit confusing indeed, changed it [here](https://github.com/tambo-ai/tambo/pull/993/commits/e8f421202345a838e2718d94c9298285c55deb9c)

docs/src/providers/tambo-provider.tsx
akhileshrangani4: done, [here](https://github.com/tambo-ai/tambo/pull/993/commits/a576cd1bb04965251904b496647dbd3b82c3a770)

docs/src/components/ai-actions/open-dropdown.tsx
akhileshrangani4: done, [here](https://github.com/tambo-ai/tambo/pull/993/commits/fada6294ddb4c6a07ecc3a2942e7900091da8d45)

docs/src/components/tambo/message-input.tsx
akhileshrangani4: as we have already talked about integrating some feature where we would want different ai actions- basically opening tambo chats throughout our dashboard, my ans to this would be yes, we can add this to our open-source component- I will add this in the recent PR(#996)

docs/src/components/ai-actions/open-dropdown.tsx
akhileshrangani4: it does avoid a reload but the MessageThreadCollapsible component only reads the initialQuery on initial render. so when I navigate client-side, the component doesn't re-read the query parameter because next Link does client-side navigation. hence, it does not trigger the component to auto-open the chat.

alecf (approved):
No

PR #997, opened 2025-09-06
Feat/voice input

ananddtyagi:


ananddtyagi:
I have to review this PR once more before it's ready for review, but the complete functionality is now done!

react-sdk/src/hooks/use-voice-input.tsx
```diff
@@ -0,0 +1,276 @@
...
+      try {
+        // Create FormData for multipart upload
+        const formData = new FormData();
+        formData.append("audio", audioBlob, "recording.webm");
+        formData.append("model", "gpt-4o-mini-transcribe");
+        formData.append("response_format", "text");
+
+        // Send to backend for transcription
+        // Use the current window origin for the API route
+        const baseUrl =
```
MichaelMilstead, line 127: I wonder if we should just add a route for this to our `tambo-cloud` repo and have this default point there, so a dev can just turn this on without needing to implement the route themselves. Maybe we can make it so the dev can override this though in case they want to send to a custom transcription service

ananddtyagi: Updated this to use either the client's baseUrl that is set or the local once in case that's what the user is testing on.

showcase/src/components/ui/tambo-voice-input-provider.tsx
MichaelMilstead, line 1: I think we don't need this file, anywhere that needs the voice hook should import it from the react sdk

docs/src/components/tambo/message-input.tsx
```diff
@@ -1,18 +1,17 @@
 "use client";
 
-import { McpConfigModal } from "@/components/tambo/mcp-config-modal";
-import {
-  Tooltip,
-  TooltipProvider,
-} from "@/components/tambo/suggestions-tooltip";
+import { McpConfigModal } from "@/components/ui/mcp-config-modal";
```
MichaelMilstead: I think we can leave these imports as they were

docs/src/components/ui/markdown-components.tsx
MichaelMilstead, line 1: this might be duplicating a file that's already under `/components/tambo`

docs/src/components/ui/mcp-config-modal.tsx
MichaelMilstead, line 1: same here about duplicating a file that's already under `/components/tambo`

docs/src/components/ui/suggestions-tooltip.tsx
MichaelMilstead, line 1: same here about duplicating a file that's already under `/components/tambo`

docs/content/docs/getting-started/integrate.mdx
```diff
@@ -58,6 +58,27 @@ Note that the `TamboProvider` only works in the browser. On Next.js, specify
 `"use client"` at the top of your file to ensure that the `TamboProvider` is
 rendered in the browser.
 
+#### Voice Input Configuration
```
MichaelMilstead: I think info about voice input can have its own docs page under `/concepts/message-threads/` similar to image attachments. Just want to keep this "getting started" info as simple as possible

showcase/src/components/ui/message-input.tsx
```diff
@@ -442,13 +442,196 @@ const MessageInputMcpConfigButton = React.forwardRef<
...
+    </TooltipProvider>
+  );
+});
+MessageInputVoiceButton.displayName = "MessageInput.VoiceButton";
+
+/**
+ * Real-time transcription indicator component
+ * Shows a pulsing wave animation when real-time transcription is active
+ */
+const MessageInputRealTimeIndicator = React.forwardRef<
```
MichaelMilstead: Looks like we can copy this `MessageInputRealTimeIndicator` into the message-input component under `/cli` (we try to keep them in sync, and for now it's manual)

react-sdk/src/hooks/use-voice-input.tsx
```diff
@@ -0,0 +1,365 @@
...
+      try {
+        // Create FormData for multipart upload
+        const formData = new FormData();
+        formData.append("audio", audioBlob, "recording.webm");
+        formData.append("model", "gpt-4o-mini-transcribe");
+        formData.append("response_format", "text");
+
+        // Send to backend for transcription
+        const baseUrl =
+          client?.baseURL ??
```
MichaelMilstead: I think client.baseURL will always exist, and will always point to the tambo API (hosted or local, wherever the app's messages are being processed), so there's not a way for a developer to point their own transcription URL separate from the tambo api.

Maybe we can add an optional prop to the TamboVoiceInputContext called `transcriptionUrl` or something, which, if set, is used here

MichaelMilstead (changes requested):
Nice! Everything is working, just left a few comments for small changes.

One thing though, I think we should wait to merge this until the API has a route to perform transcription, so it works out of the box without the developer needing to customize anything (while still letting them customize if they want.)

Then we can also remove all the api routes here in the docs and showcase

react-sdk/src/hooks/use-voice-input.tsx
```diff
@@ -0,0 +1,437 @@
...
+ * const { startRecording, stopRecording, isRecording } = useVoiceInput();
+ *
+ * return (
+ *   <button onClick={isRecording ? stopRecording : startRecording}>
+ *     {isRecording ? 'Stop' : 'Start'} Recording
+ *   </button>
+ * );
+ * ```
+ */
+export const useVoiceInput = (
```
alecf, line 50: This is a rather large function with a lot of interrlated state - can you break it up into a few internal hooks to help untangle the states here?

showcase/src/components/ui/use-voice-input.tsx
```diff
@@ -0,0 +1,232 @@
...
+        formData.append("model", "gpt-4o-mini-transcribe");
+        formData.append("response_format", "text");
+
+        // Send to backend for transcription
+        // Use the local API route for development, or configured baseURL for production
+        const baseUrl =
+          client.baseURL ||
+          (typeof window !== "undefined"
+            ? window.location.origin
+            : "https://api.tambo.ai");
```
alecf, line 92: our domain is tambo.co not tambo.ai

alecf (changes requested):
So after taking a step back, I think there are a few architectural problems with this approach:
1) by using `window.location` as the base url, you're requiring the developer to add an endpoint on whatever site they're on, rather than having a clean fallback to something on the tambo side. This also means that you've had to add an api at `/api/v3/...` on the showcase to match
2) everything else in the project is set up via the project in the tambo-cloud dashboard, and it's not clear why this particular configuration would be different.
3) fundamentally this should just be an API call to the Tambo API server by default using user's openai key, and that's set up the `tambo-cloud` repository.
4) a lot of the "isTranscribing" etc lines up complete with react-query, and the existing patterns we're using around isLoading/etc. 

So I think a lot of whats here can be used for the above, but a few changes are necessary:
1) instead of requiring an endpoint that matches the OpenAI endpoint, I think it would be better if the abstraction was around a function callback - i.e. rather than calling `fetch()` and having the url be somewhat pluggable, I think instead this just be a simple function, and folks can replace the function if they want. By default it can still fall through to calling an API on the Tambo API server, and re-using the user's OpenAI key if they have one. this also lets us continue to use the existing API keys and not have to require the user to have an open/unauthenticated endpoint for transcription
2) in order to support an overridable-function, I think we need an optional voice provider, i.e. `<TamboVoiceProvider handler={(value) => {....}}>{.....}</TamboVoiceProvider>` - then useTamboVoice() can just try to get that value and if its not there fallback to an internal function that calls the tambo API with the right API key, e.g.

```ts
const { transcribe: providerTranscribe } = useContext(TamboVoiceContext);
const transcribe = providerTranscribed ?? tamboTranscribe;
...
```

3) The actual callback should be wrapped in useTamboQuery() so the usage would be something like


```ts
const { mutate: transcribe, isLoading: isTranscribing } = useTamboVoice();
```

MichaelMilstead:
Continued and merged in a [separate PR](https://github.com/tambo-ai/tambo/pull/1234) to get this through. Thanks @ananddtyagi !

PR #1036, opened 2025-09-11
feat(sdk): partial updates for interactables + auto tools + new docs

### LINKED ISSUE #990: TamboHack: Enable Partial Updates to Interactables
**Problem:**
Currently tambo registers a tool for each interactable component to allow updates to those components (see [here](https://docs.tambo.co/concepts/components/interactable-components)). These tools require tambo to completely regenerate the list of props for an interactable component, which sometimes takes longer than needed, if the update is a small change.

We should enable "Partial updates" to interactables.

**Proposed Solution:**

- update the update_interactable_component definition so it allows partial parameters.

- update the updateInteractableComponentProps function so it applies partial parameter updates to what exists.

**Acceptance Criteria:**

- If I ask for a small change to an interactable, tambo returns only the required change rather than an entire new set of props

- Tests covering the behavior

- Documentation about interactable components is updated to describe this behavior

shivam096:
### What's changed
- Added partial prop updates for interactable components with clear success/error messages.
- Auto-registered interactable tools for discoverability and CRUD.
- Introduced `useCurrentInteractablesSnapshot` hook to safely read interactables.
- Added new docs page covering partial updates, tools, and snapshot usage.

### Implementation
- **Partial updates**: Shallow-merge behavior in `updateInteractableComponentProps` with no-op detection and warnings for empty payloads.
- **Auto tools** (registered when interactables exist):
  - `get_all_interactable_components`
  - `get_interactable_component_by_id`
  - `remove_interactable_component`
  - `update_interactable_component_<id>` (partial schema from `propsSchema.partial()`)
- **Snapshot hook**: `useCurrentInteractablesSnapshot` returns a cloned, mutation-safe array for context helpers and UIs.
- **Docs**:
  - New page: `tambo/docs/content/docs/concepts/components/interactable-partial-updates.mdx`
  - Sidebar updated: `tambo/docs/content/docs/concepts/components/meta.json`

### Video

https://github.com/user-attachments/assets/4b149b9d-a905-4b90-aaa4-497579caf801

### Testing
- Added comprehensive tests for partial updates, nested behavior, arrays, null/undefined, no-op, and performance:
  - `tambo/react-sdk/src/providers/__tests__/tambo-interactable-provider-partial-updates.test.tsx`
- All existing tests pass.

### Next steps
- Consider an opt-in deep-merge helper for nested updates if needed.
- Expand examples in showcase to demonstrate partial vs full nested updates.

Closes #990 

### Checklist
- [x] Semantic PR title using Conventional Commits
- [x] Linked issue
- [ ] Backend PR needed: N/A (SDK behavior + docs only)
- [x] Docs site updated (new page added)
- [ ] Visual demo video attached (≤ 2 min) (To be attached)
- [x] Tests added/updated and all CI checks green

### DIFF
--- a/react-sdk/eslint.config.mjs
+++ b/react-sdk/eslint.config.mjs
@@ -8,7 +8,15 @@ import eslintPluginReactHooks from "eslint-plugin-react-hooks";
 import tseslint from "typescript-eslint";
 
 export default tseslint.config(
-  { ignores: ["dist/**/*", "esm/**/*", "jest.config.ts", "eslint.config.mjs", "coverage/**/*"] },
+  {
+    ignores: [
+      "dist/**/*",
+      "esm/**/*",
+      "jest.config.ts",
+      "eslint.config.mjs",
+      "coverage/**/*",
+    ],
+  },
   config,
   eslint.configs.recommended,
   jsdoc.configs["flat/recommended"],


react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -174,59 +174,68 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
...
-      let updateResult = "Updated successfully";
+    (id: string, newProps: Record<string, any>): string => {
+      if (!newProps || Object.keys(newProps).length === 0) {
+        return `Warning: No props provided for component with ID ${id}.`;
+      }
 
-      setInteractableComponents((prev) => {
-        const componentExists = prev.some((c) => c.id === id);
+      // Check existence synchronously
+      const exists = interactableComponents.some((c) => c.id === id);
```
MichaelMilstead: I believe we have a stale state here so interactableComponents will always seem empty, let's keep the check within the funcitonal `setInteractableComponents` for now

shivam096: Updated the code to reflect this

docs/content/docs/concepts/components/meta.json
```diff
@@ -5,6 +5,7 @@
     "registering-with-tambo-provider",
     "registering-components-dynamically",
     "component-state",
-    "interactable-components"
+    "interactable-components",
+    "interactable-partial-updates"
```
akhileshrangani4: instead of adding a separate page here, lets just add all of this in the interactable-components page.

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -174,59 +174,68 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
...
-          return prev;
-        }
+        const updatedComponents = [...prev];
+        updatedComponents[idx] = updated;
 
         return updatedComponents;
       });
 
-      return updateResult;
+      return resultMessage;
```
akhileshrangani4: The interface in `tambo-interactable.ts` declares `updateInteractableComponentProps` returns `void`, but the implementation returns a `string`. I believe this was already broken, but since you are already updating this, could you also please update the interface in `tambo-interactable.ts` to reflect this?
```typescript
updateInteractableComponentProps: (id: string, newProps: Record<string, any>) => string;
```

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -174,59 +174,68 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
...
+        // Compare props shallowly
+        const propsChanged = Object.entries(newProps).some(([key, value]) => {
+          return original.props[key] !== value;
+        });
 
-        if (!originalComponent || !updatedComponent) {
-          updateResult = `Error: Failed to update component with ID ${id}`;
-          return prev;
+        if (!propsChanged) {
+          resultMessage = `Updated successfully`;
```
akhileshrangani4: This result message is confusing, it only checks if new values differ from existing, but still returns "Updated successfully", it should return a different message like "No changes needed" when values are identical.

docs/content/docs/concepts/components/interactable-partial-updates.mdx
```diff
@@ -0,0 +1,118 @@
+---
+title: Interactable Components — Partial Updates and Tools
+description: How partial prop updates work for interactable components, the auto-registered tools, and the snapshot hook.
+---
+
+When using interactable components, Tambo can modify props on components you have pre-placed in your UI. This page explains the partial update behavior, the tools that get auto-registered, and how to expose or customize interactable context for the AI.
+
+## Partial Updates (Shallow Merge)
```
akhileshrangani4: We say here that it is "Shallow merge" but the implementation is actually "property replacement". Shallow merge would preserve `config.language` when updating `config.theme` in the example you showed below. Hence, I feel like "shallow merge" is not the right word to use here or anywhere in this implementation.

react-sdk/src/providers/tambo-interactable-provider.tsx
akhileshrangani4: was this updated? @shivam096

docs/content/docs/concepts/components/interactable-partial-updates.mdx
```diff
@@ -0,0 +1,118 @@
...
+When there are interactable components present, the following tools are registered automatically to help the AI reason about and modify your UI:
+
+- `get_all_interactable_components` — Returns all interactable components with their current props.
+- `get_interactable_component_by_id` — Returns a specific interactable component by id.
+- `remove_interactable_component` — Removes a component from the interactables list.
+- `update_interactable_component_<id>` — Updates the props for a specific component id using partial props. The argument schema is derived from the component's `propsSchema` and accepts partials.
+
+These tools enable the AI to discover what's on the page and perform targeted updates.
+
+## Snapshot Hook: useCurrentInteractablesSnapshot
```
akhileshrangani4: can you add this to `docs/content/docs/api-reference/react-hooks.mdx` too? I think we missed that.

react-sdk/src/providers/tambo-interactable-provider.tsx
shivam096: @akhileshrangani4 yes. Added in the latest push. Somehow my last push did not have the changes

docs/content/docs/api-reference/react-hooks.mdx
```diff
@@ -319,3 +319,49 @@ Adds or replaces a helper at the given key.
...
+
+**Returns:** Array of interactable components with the following structure:
+
+- `id`: Unique identifier for the component
+- `name`: Component name
+- `description`: Component description
+- `props`: Cloned props object
+- `propsSchema`: Optional Zod schema for props validation
+
+**Example:**
```
akhileshrangani4: we actually do not need a detailed explanation here, see how other hooks are written here and do the same for this one.

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -174,59 +174,73 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
...
+    (id: string, newProps: Record<string, any>): string => {
+      if (!newProps || Object.keys(newProps).length === 0) {
+        return `Warning: No props provided for component with ID ${id}.`;
+      }
+
+      // Initialize result message
+      let resultMessage = "Updated successfully";
+
+      // Find the component from current state
+      const component = interactableComponents.find((c) => c.id === id);
```
akhileshrangani4: this can still become stale between render cycles. My suggestion would be to move all the logic inside the `setInteractableComponents` callback to make sure we are working with latest state.

shivam096: Oh ok.

I did not think that through. Would update it accordingly. Will update the PR by today EOD.

akhileshrangani4 (approved):
Looks good to go now, I just updated the react hook doc fix- where we did not need the examples and detailed explaination.

MichaelMilstead (approved):
Awesome!

PR #1059, opened 2025-09-16
fix(react-sdk): refetch thread list on new thread creation

effe-megna:
Fix: refetch thread list on new thread creation

Changes:
- Updated `handleAdvanceStream` and `sendThreadMessage` in `TamboThreadProvider` to conditionally refetch the threads list when transitioning from a placeholder thread to a real thread. The code performs an optimistic update first, then refetches the threads query.
- Removed the redundant `refetch` call in ThreadHistoryNewButton's `handleNewThread` (see my comment below for the why)
- Added tests to cover thread list refetch scenarios.

Demo:

As you can see, the thread in the sidebar appears as soon as the server respond with the first message, making possible the switch from the placeholder thread to the real one.

https://www.loom.com/share/ad3a68e1b2a9426b99f9781a61340de3

### DIFF
--- a/react-sdk/src/providers/__tests__/tambo-thread-provider.test.tsx
+++ b/react-sdk/src/providers/__tests__/tambo-thread-provider.test.tsx
@@ -114,7 +114,7 @@ describe("TamboThreadProvider", () => {
   } satisfies PartialTamboAI as unknown as TamboAI;
 
   let mockQueryClient: {
-    refetchQueries: jest.Mock;
+    invalidateQueries: jest.Mock;
     setQueryData: jest.Mock;
   };
 
@@ -159,7 +159,7 @@ describe("TamboThreadProvider", () => {
 
     // Setup mock query client
     mockQueryClient = {
-      refetchQueries: jest.fn().mockResolvedValue(undefined),
+      invalidateQueries: jest.fn().mockResolvedValue(undefined),
       setQueryData: jest.fn(),
     };
     jest
@@ -1090,9 +1090,8 @@ describe("TamboThreadProvider", () => {
       );
 
       // Verify that refetchQueries was called when the new thread was created
-      expect(mockQueryClient.refetchQueries).toHaveBeenCalledWith({
+      expect(mockQueryClient.invalidateQueries).toHaveBeenCalledWith({
         queryKey: ["threads"],
-        type: "active",
       });
     });
 
@@ -1126,7 +1125,7 @@ describe("TamboThreadProvider", () => {
 
       // Verify that neither setQueryData nor refetchQueries were called
       expect(mockQueryClient.setQueryData).not.toHaveBeenCalled();
-      expect(mockQueryClient.refetchQueries).not.toHaveBeenCalled();
+      expect(mockQueryClient.invalidateQueries).not.toHaveBeenCalled();
 
       // Verify the thread was switched correctly
       expect(result.current.thread.id).toBe("existing-thread-123");


showcase/src/components/ui/thread-history.tsx
```diff
@@ -240,7 +240,6 @@ const ThreadHistoryNewButton = React.forwardRef<
 
       try {
         await startNewThread();
-        await refetch();
```
effe-megna, line 243: I assume the refetch call in the handleNewThread function was an attempt to retrieve the newly created thread and display it in the history list.

However, creating a new thread on the client does not mean it has been created on the server, because the new thread is just a placeholder.

This refetch call does nothing, so I am removing it.

react-sdk/src/providers/tambo-thread-provider.tsx
```diff
@@ -246,6 +247,39 @@ export const TamboThreadProvider: React.FC<
...
+          ["threads", currentProject.id, contextKey],
+          (old: any) => {
+            return {
+              ...old,
+              items: [optimisticThread, ...old.items],
+            };
+          },
+        );
+
+        await queryClient.refetchQueries({
```
alecf: instead of manually fetching, can we simply invalidate the query? hopefully that would mean that if nobody wants the thread list, then it is never fetched.

react-sdk/src/providers/tambo-thread-provider.tsx
```diff
@@ -246,6 +247,39 @@ export const TamboThreadProvider: React.FC<
...
+        const optimisticThread = {
+          ...PLACEHOLDER_THREAD,
+          id: threadId,
+          createdAt: new Date().toISOString(),
+          updatedAt: new Date().toISOString(),
+        };
+
+        queryClient.setQueryData(
+          ["threads", currentProject.id, contextKey],
+          (old: any) => {
```
alecf: can you more strongly type `old`?

react-sdk/src/providers/tambo-thread-provider.tsx
```diff
@@ -677,7 +715,18 @@ export const TamboThreadProvider: React.FC<
...
           ) {
             hasSetThreadId = true;
+            const wasPlaceholderThread =
+              currentThreadId === PLACEHOLDER_THREAD.id;
             await switchCurrentThread(chunk.responseMessageDto.threadId, false);
+
+            // If we're switching from placeholder to a real thread
+            // this means a new thread was created, so refetch the threads list
+            if (wasPlaceholderThread) {
+              await refetchThreadsList(
```
alecf, line 724: I'd like to avoid passing the contextKey along here - again can we invalidate rather than refetch?

react-sdk/src/providers/tambo-thread-provider.tsx
```diff
@@ -931,7 +983,17 @@ export const TamboThreadProvider: React.FC<
             componentList,
           )
         : advanceResponse.responseMessageDto;
+      const wasPlaceholderThread = currentThreadId === PLACEHOLDER_THREAD.id;
       await switchCurrentThread(advanceResponse.responseMessageDto.threadId);
+
+      // If we're switching from placeholder to a real thread
+      // this means a new thread was created, so refetch the threads list
+      if (wasPlaceholderThread) {
+        await refetchThreadsList(
```
alecf, line 991: (ditto on invalidate)

showcase/src/components/ui/thread-history.tsx
alecf: good call!

alecf (changes requested):
a few requests around invalidation... I am assuming that invalidation will just work as well as the refetch, but please let me know if it doesn't work for some reason.

It is possible that we're calling refetch elsewhere (may even have been me that wrote that!) when we should be invalidating, but I'd rather move forward with invalidation that keep the refetch pattern

react-sdk/src/providers/tambo-thread-provider.tsx
effe-megna: Oh yes, since there is an optimistic update, invalidating the query is more appropriate. I’ve made the change here f16fb26213a29df59f7e271368f9240c4057b176

react-sdk/src/providers/tambo-thread-provider.tsx
effe-megna: Done here bb829fd23a55b4ec8db2da304904caf0127d7d33

react-sdk/src/providers/tambo-thread-provider.tsx
effe-megna: The `contextKey` is used to reconstruct the Query Key [here](https://github.com/tambo-ai/tambo/pull/1059/files/85c500d6842a298d210806b437f95bf713f410bc#diff-d30c2e9c4fcf05004582e87680dc72758dc6135a049226bb72cefe676ce070e3R263) for updating the cache and performing the optimistic update before invalidating the query.

If I omit the context key, the TanStack Query library won’t know which object to update in the cache. This is because [useTamboThreadList](https://github.dev/tambo-ai/tambo/blob/64b2da187dbbc1fda98953bedb3cfe84c6e83d66/react-sdk/src/hooks/use-tambo-threads.ts#L62) also relies on the context key.

Now, I'm not really sure what's the contextKey is for in Tambo, I see it is marked as optional in many function signature but at runtime, at least locally, is always present 🤔

alecf: ok that's fine for now. It is one of two mechanisms for identifying the user but I now see why we need it here, all good

effe-megna: cool, thanks for all the feedback!

alecf (approved):
Looks great! Thanks!

alecf:
I'm going to hold off merging this because we had a change to our underlying SDK that I need to verify before merging more stuff... hoping to land this tomorrow (18-Sep)

PR #1071, opened 2025-09-19
feat(docs): add documentation for custom llm parameters

akhileshrangani4:


docs/content/docs/models/custom-llm-parameters.mdx
```diff
@@ -0,0 +1,165 @@
+---
+title: Custom LLM Parameters
+description: Configure advanced model parameters for fine-tuned AI responses and behavior control.
+---
+
```
MichaelMilstead, line 5: can we just add a little intro section (maybe one or two sentence) that describes: Tambo uses an LLM behind the scenes to process user messages. You can change what model Tambo uses. Tambo uses certain parameters by default when calling the LLM, but you can override those.

MichaelMilstead (approved):

docs/content/docs/models/custom-llm-parameters.mdx
akhileshrangani4: just pushed! [here](https://github.com/tambo-ai/tambo/pull/1071/commits/f880ef94fdd7678451c5b13125615fb512b54a1b)

PR #1079, opened 2025-09-19
fix: add validation to prevent spaces in component and tool names

akhileshrangani4:


react-sdk/src/util/validate-component-name.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Validates that a name contains no spaces
+ * @param name The name to validate
+ * @param contextName The context name (e.g., "component", "tool")
+ */
+export function assertNoSpacesInName(name: string, contextName: string): void {
+  if (name.includes(" ")) {
+    throw new Error(
+      `${contextName} "${name}" cannot contain spaces. Use underscores, hyphens, or camelCase instead.`,
+    );
+  }
```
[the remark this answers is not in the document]
alecf: I get this from the openai api:
Invalid 'tools[0].function.name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.

lets just use that regex directly

akhileshrangani4:
@alecf updated everything

react-sdk/src/util/validate-component-name.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Validates that a name matches OpenAI function name requirements
+ * @param name The name to validate
+ * @param contextName The context name (e.g., "component", "tool")
+ */
+export function assertNoSpacesInName(name: string, contextName: string): void {
```
MichaelMilstead: I think we can update the name of this function since it now checks for more than just spaces

MichaelMilstead: also since we use it for tools too, maybe we change the filename too

akhileshrangani4: done!

alecf (approved):

PR #1083, opened 2025-09-22
chore: Add next-sitemap to docs sitemap to see

michaelmagan:
Our current sitemap is giving us issues in Google Console. It might be because it's a dynamic route. 

Trying to use `next-sitemap` to generate a static, deterministic sitemap for docs at build time.


michaelmagan:
@CharlieHelps can you fix the formatting issue?

PR #1096, opened 2025-09-23
refactor: update showcase site to use updated structure, and updated the thread history component so its not fixed.

michaelmagan:
- Update showcase to use new component structure.
- Update thread history to not use fixed positions


michaelmagan:
@akhileshrangani4 one thing I ran into on this pr was that the imports in the registry don't match whats actually installed since everything moved into tambo/.

registry:
```
import { ScrollableMessageContainer } from "@/components/ui/scrollable-message-container";
```

actual:
```
import { ScrollableMessageContainer } from "@/components/tambo/scrollable-message-container";
```

Is this intentional?

akhileshrangani4:
@michaelmagan it was intentional at the time we changed the directory to be /tambo, but I think we are fine now. I already have an incoming PR for updating the components with reasoning UI so I will update the paths too with that.

PR #1101, opened 2025-09-26
feat: add reasoning UI, smart autoscroll with UI improvements and update component paths to use /tambo

akhileshrangani4:
This branch introduces reasoning display capabilities for AI messages along with 
intelligent scrolling behavior and various UI/UX enhancements across the codebase.

### Reasoning Display
- Add ReasoningInfo component to display AI reasoning in collapsible panels
- Implement auto-scroll within reasoning panel during streaming
- Show reasoning with markdown formatting support

### Smart Autoscroll
- Implement user-aware autoscroll that detects scroll position
- Auto-scroll only when user is at bottom of container
- Disable autoscroll when user manually scrolls up
- Re-enable when user returns to bottom

### Performance Optimizations
- Create singleton markdownComponents instance to avoid re-creating on each render
- Refactor createMarkdownComponents to return Record type for better type safety
- Simplify content validation with hasContentInItem helper

### Message Display Enhancements
- Add MessageImages component for better image display
- Improve formatToolResult to handle arrays and JSON parsing
- Enhanced content parsing with better type safety
- Add animated gradient effect for thinking/loading states

### Developer Experience
- Update import paths from @/components/ui to @/components/tambo for consistency
- Add CSS configuration check and setup in upgrade command
- Improve message-input handlers for paste and file upload
- Better TypeScript types across components

### Files Changed
- Components: message.tsx, message-input.tsx, scrollable-message-container.tsx
- Utils: thread-hooks.ts, markdown-components.tsx
- Styles: globals-v3.css, globals-v4.css, global.css
- Documentation and showcase examples updated

## Breaking Changes
- getMessageImages now only accepts arrays instead of mixed content types
- createMarkdownComponents returns Record type instead of Components type


cli/src/registry/message/message.tsx
```diff
@@ -439,6 +439,115 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
...
+
+    // Auto-collapse when content arrives and reasoning is not loading
+    React.useEffect(() => {
+      if (checkHasContent(message.content) && !isLoading) {
+        setIsExpanded(false);
+      }
+    }, [message.content, isLoading]);
+
+    // Only show if there's reasoning data
+    if (!message.reasoning || message.reasoning.length === 0) {
```
alecf: FYI things like this can be simplified:

```suggestion
    if (!message.reasoning?.length) {
```

You do NOT need to change that for this PR, just a ProTip™️

docs/src/components/tambo/message-input.tsx
```diff
@@ -297,12 +388,33 @@ const MessageInputTextarea = ({
     }
   };
 
+  const handlePaste = async (e: React.ClipboardEvent<HTMLTextAreaElement>) => {
+    const items = Array.from(e.clipboardData.items);
+    const imageItems = items.filter((item) => item.type.startsWith("image/"));
+
+    if (imageItems.length > 0) {
```
alecf: minor but I don't see a reason to add the if() here - seems like you still want to preventDefault(), and iterating an empty array is free (i.e. `for (const item of imageItems) {}` is a no-op if the array is empty)

docs/src/components/tambo/message-input.tsx
```diff
@@ -505,6 +619,155 @@ const MessageInputError = React.forwardRef<
...
+  const { addImages } = useTamboThreadInput();
+  const fileInputRef = React.useRef<HTMLInputElement>(null);
+
+  const handleClick = () => {
+    fileInputRef.current?.click();
+  };
+
+  const handleFileChange = async (e: React.ChangeEvent<HTMLInputElement>) => {
+    const files = Array.from(e.target.files ?? []);
+    if (files.length > 0) {
```
alecf: similar check: why check the length if it's free it iterate - and wouldn't you want to reset the input anyway?

docs/src/components/tambo/message.tsx
```diff
@@ -346,7 +405,7 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
           <div
             id={toolDetailsId}
             className={cn(
-              "flex flex-col gap-1 p-3 overflow-auto transition-[max-height,opacity,padding] duration-300 w-full",
+              "flex flex-col gap-1 p-3 overflow-hidden transition-[max-height,opacity,padding] duration-300 w-full",
```
alecf: if you're adding overflow-hidden, probably also want `text-ellipsis` or just `truncate`

cli/src/registry/message/message.tsx
```diff
@@ -405,7 +405,7 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
           <div
             id={toolDetailsId}
             className={cn(
-              "flex flex-col gap-1 p-3 overflow-auto transition-[max-height,opacity,padding] duration-300 w-full",
+              "flex flex-col gap-1 p-3 overflow-hidden transition-[max-height,opacity,padding] duration-300 w-full",
```
michaelmagan: Overflow-hidden, this looks like the tool response?

A better solution is to add a horizontal scroll, but only for the tool response (not the entire message thread) which is what we had before.

docs/src/components/tambo/message.tsx
```diff
@@ -380,6 +439,115 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
...
+                "w-3 h-3 transition-transform duration-200",
+                !isExpanded && "-rotate-90",
+              )}
+            />
+          </button>
+          <div
+            id={reasoningDetailsId}
+            className={cn(
+              "flex flex-col gap-1 px-3 py-3 overflow-auto transition-[max-height,opacity,padding] duration-300 w-full",
+              "[&::-webkit-scrollbar]:w-[6px]",
```
alecf: I feel like we should not be messing with scrollbar look & feel in our components...can we leave them alone? (I also wish we weren't hardcoding the max-h here, but I suppose it works for now)

docs/src/components/tambo/message.tsx
```diff
@@ -404,14 +573,33 @@ function formatToolResult(
   try {
     const parsed = JSON.parse(safeContent);
     return (
-      <pre className="bg-muted/50 rounded-md p-3 text-xs overflow-x-auto overflow-y-auto max-w-full max-h-64">
+      <pre
+        className={cn(
+          "bg-muted/50 rounded-md p-3 text-xs overflow-x-auto overflow-y-auto max-w-full max-h-64",
+          "[&::-webkit-scrollbar]:w-[6px]",
```
alecf: (ditto on scrollbars)

docs/src/lib/thread-hooks.ts
```diff
@@ -139,10 +139,26 @@ export function checkHasContent(
     return content.some(
       (item) =>
         item &&
-        item.type === "text" &&
-        typeof item.text === "string" &&
-        item.text.trim().length > 0,
+        ((item.type === "text" &&
```
alecf: this is now incredibly complicated and I have absolutely no idea what it means or what its trying to do. Can you break this out into a named function with if/else and early returns?

also in general really try to stay away from `typeof` - we know based on types that if `item.type === "text"` then `item.text` is a string, we don't need to check that. if you're trying to compare against undefined, you can do stuff like `if (!item.text?.trim())`

docs/src/lib/thread-hooks.ts
```diff
@@ -139,10 +139,26 @@ export function checkHasContent(
...
   }
   return false; // Default for unknown types
 }
+
+/**
+ * Extracts image URLs from message content.
+ * @param content - The message content
+ * @returns Array of image URLs
+ */
+export function getMessageImages(
```
alecf, line 178: I don't really get this function - at least why it is so permissive in accepting ReactNode/etc - it looks like the only way it is used is for content arrays, so lets just let that be the only kind of parameter accepted and avoid the Array.isArray stuff.

showcase/src/components/ui/message.tsx
```diff
@@ -443,13 +439,170 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
...
+                  <span className="text-muted-foreground text-xs font-medium">
+                    Step {index + 1}:
+                  </span>
+                )}
+                {reasoningStep ? (
+                  <div className="bg-muted/50 rounded-md p-3 text-xs overflow-x-auto overflow-y-auto max-w-full">
+                    <div className="whitespace-pre-wrap break-words">
+                      <Streamdown
+                        components={
+                          createMarkdownComponents() as Record<
```
alecf: This PR casts the result to this function being cast `Record<string, ...>` 8 separate times! It feels like the original function just needs to return the correct type so we don't have to cast it so much.

I also wonder if we should just be calling this once and importing the result and/or calling it once per component at the module level?

showcase/src/components/ui/message.tsx
```diff
@@ -443,13 +439,170 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
...
+ * @param content - The content to check and format
+ * @returns Formatted content or original content if not JSON
+ */
+function formatToolResult(
+  content: TamboThreadMessage["content"],
+  enableMarkdown = true,
+): React.ReactNode {
+  if (!content) return content;
+
+  const safeContent = getSafeContent(content);
```
alecf: it looks like we're only using safeContent if the JSON.parse fails - can we delay calling this until that case?

alecf (commented):
lots of small stuff but highlevel:
- it is very often not necessary to check array lengths - just allow arrays to be empty and then iterate the empty array as a no-op. The check is just more code to maintain! Also when you *do* need to check array length, the shorthand `!array?.length` is a quick way to handle null/undefined/etc
- really really try to avoid `typeof` and `Array.isArray()` in code or to make code too permissive of a variety of types - much better to have very strict types to start and only be more permissive when some unusual requirement comes through that requires the runtime type checks. If you have two cases for different types, might be better to have 2 functions with 2 distinct signatures for the different types (which also often means having one call the other onces the types have been further resolved)
- any time you have more than 1 level of ternaries or `&&`+`||` it is better to break it into a separate  function so a human can interpret the code - most of us slow way down reading code with long complicated chains of boolean logic.

alecf (approved):

PR #1133, opened 2025-10-07
docs: add prop syncing guidance for interactable components

michaelmagan:
## Summary
- add local state and `useEffect` synchronization to the interactable component examples
- clarify that the examples rely on effect hooks to sync local state with AI-driven prop updates

## Testing
- not run (docs only)

------
https://chatgpt.com/codex/tasks/task_e_68e550f6fba08332886aa1517bc2256e

### DIFF
--- a/docs/content/docs/concepts/components/syncing-editable-interactables.mdx
+++ b/docs/content/docs/concepts/components/syncing-editable-interactables.mdx
@@ -77,10 +77,7 @@ function EditableNote(initial: NoteProps) {
 
   return (
     <section>
-      <InteractableNote
-        {...note}
-        onInteractableReady={setInteractableId}
-      />
+      <InteractableNote {...note} onInteractableReady={setInteractableId} />
       <label>
         Edit content
         <textarea
@@ -169,10 +166,7 @@ function EditableNote(initial: NoteProps) {
 
   return (
     <section>
-      <InteractableNote
-        {...note}
-        onInteractableReady={setInteractableId}
-      />
+      <InteractableNote {...note} onInteractableReady={setInteractableId} />
       <label>
         Edit content
         <textarea


docs/content/docs/concepts/components/interactable-components.mdx
```diff
@@ -12,16 +12,52 @@ Unlike regular registered components that Tambo generates and renders from scrat
...
+          value={currentColor}
+          onChange={(event) => setCurrentColor(event.target.value as NoteProps["color"])}
+        >
+          <option value="white">White</option>
+          <option value="yellow">Yellow</option>
+          <option value="blue">Blue</option>
+          <option value="green">Green</option>
+        </select>
+      </div>
+    </section>
```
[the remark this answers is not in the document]
michaelmagan: @CharlieHelps yes please

docs/content/docs/concepts/components/interactable-components.mdx
```diff
@@ -60,19 +121,19 @@ const InteractableCounter = withInteractable(Counter, {
...
       />
       <InteractableCounter count={42} label="Items" />
-    </div>
+    </main>
   );
 }
 ```
 
-Now Tambo is able to read and update these components in-place when responding to messages. Users can ask things like:
+Both components track user-editable state and rely on `useEffect` hooks to synchronize their local state whenever Tambo updates the props. Now Tambo is able to read and update these components in-place when responding to messages. Users can ask things like:
```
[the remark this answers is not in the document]
michaelmagan: @CharlieHelps fix the example to handle two-way synchronization correctly.

docs/content/docs/concepts/components/interactable-components.mdx
```diff
@@ -57,21 +72,125 @@ const InteractableCounter = withInteractable(Counter, {
   }),
 });
 
+// Controlled wrappers to achieve two-way synchronization
+function ControlledInteractableNote(props: NoteProps) {
```
MichaelMilstead: I think this should be a separate page or some kind of sub-page. It adds a huge example that feels like it confuses the basic use case/idea of interactables

docs/content/docs/concepts/components/interactable-components.mdx
```diff
@@ -57,21 +72,125 @@ const InteractableCounter = withInteractable(Counter, {
...
       />
-      <InteractableCounter count={42} label="Items" />
-    </div>
+      <ControlledInteractableCounter count={42} label="Items" />
+    </main>
   );
 }
 ```
 
+The controlled wrappers ensure two-way synchronization:
```
MichaelMilstead: can we update this with a small example of some component interaction that requires the wrapper pattern?

docs/content/docs/concepts/components/interactable-components.mdx
michaelmagan: Yes. but the existing example will not work as expected. The props won't update the state and the component will not show updated state. I think its important that code blocks work.

michaelmagan:
@CharlieHelps  run npm run format and then commit.

PR #1152, opened 2025-10-10
refactor(message): simplify tool call request retrieval and enhance status message handling

akhileshrangani4:


cli/src/registry/message/message.tsx
```diff
@@ -308,12 +320,13 @@ function getToolStatusMessage(
...
-  const isToolCall = !!message.toolCallRequest;
-  if (!isToolCall) return null;
+  if (message.role !== "assistant" || !getToolCallRequest(message)) {
+    return null;
+  }
 
   const toolCallMessage = isLoading
-    ? `Calling ${message.toolCallRequest?.toolName ?? "tool"}`
-    : `Called ${message.toolCallRequest?.toolName ?? "tool"}`;
+    ? `Calling ${getToolCallRequest(message)?.toolName ?? "tool"}`
```
alecf, line 328: this unfortunately will break certain things - right now I'm pretty sure if if you're streaming a server-side mcp call, I believe the tool call request is only inside `message.component.toolCallRequest` - its a subtle thing that we need to fix, but no here

alecf: oh wait sorry I misread this diff, I thought it was *removing* the component.toolCallRequest

PR #1167, opened 2025-10-17
feat: (ui) add ui for displaying MCP sampling requests

MichaelMilstead:
Updates message component to show an MCP server's sampling messages within a toolcall.

Updates the threadcontent to filter out any messages that have a parentId, since these are "sub messages" and should not be shown in the usual thread 

image

### DIFF
--- a/cli/src/registry/message/message.tsx
+++ b/cli/src/registry/message/message.tsx
@@ -418,18 +418,16 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
               isExpanded ? "max-h-auto opacity-100" : "max-h-0 opacity-0 p-0",
             )}
           >
-            <span className="whitespace-pre-wrap">
+            <span className="whitespace-pre-wrap pl-2">
               tool: {toolCallRequest?.toolName}
             </span>
-            <span className="whitespace-pre-wrap">
+            <span className="whitespace-pre-wrap pl-2">
               parameters:{"\n"}
               {stringify(keyifyParameters(toolCallRequest?.parameters))}
             </span>
-            <div className="-ml-2">
-              <SamplingSubThread parentMessageId={message.id} />
-            </div>
+            <SamplingSubThread parentMessageId={message.id} />
             {associatedToolResponse && (
-              <div>
+              <div className="pl-2">
                 <span className="whitespace-pre-wrap">result:</span>
                 <div>
                   {!associatedToolResponse.content ? (
@@ -504,10 +502,10 @@ const SamplingSubThread = ({
         )}
         aria-hidden={!isExpanded}
       >
-        <div className="pl-8">
+        <div className="pl-2">
           <div className="border-l-2 border-muted-foreground p-2 flex flex-col gap-4">
             {childMessages?.map((m: TamboThreadMessage) => (
-              <div key={m.id} className={`${m.role === "assistant" && "pl-2"}`}>
+              <div key={m.id} className={`${m.role === "user" && "pl-2"}`}>
                 <span
                   className={cn(
                     "whitespace-pre-wrap",


cli/src/registry/message/message.tsx
```diff
@@ -448,6 +449,88 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
...
 ToolcallInfo.displayName = "ToolcallInfo";
 
+/**
+ * Displays a message's child messages in a collapsible dropdown.
+ * Used for MCP sampling sub-threads.
+ * @component SamplingSubThread
+ */
+const SamplingSubThread = ({
+  parentMessageId,
+  titleText = "requested the LLM",
```
alecf: I do think this needs to be better - this isn't correct or really a complete phrase, other options:

1) request to the LLM (noun)
2) requested from the LLM
3) MCP server made an LLM request
4) MCP server requested from the LLM

tool calls use past  tense:

Image

MichaelMilstead: Starting to feel like it should be something extremely general since the end user will see it and maybe doesn't care or know about LLMs (until we figure out a way to dynamically generate the string based on the request)

Maybe default to "completed processing" or "finished extra processing" or something?

alecf: oh yeah good point! What about "finished additional work" ? "additional" feels less extraneous than extra, "work" feels more friendly than "processing"

MichaelMilstead: ok agreed on both points will update!

cli/src/registry/message/message.tsx
```diff
@@ -418,15 +418,16 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
               isExpanded ? "max-h-auto opacity-100" : "max-h-0 opacity-0 p-0",
             )}
           >
-            <span className="whitespace-pre-wrap">
+            <span className="whitespace-pre-wrap pl-2">
```
alecf: so generally we should be using gap rather than padding here - the parent has `gap-1` so maybe the answer is just to bump that to `gap-2`?

--test--

cli/src/registry/message/message.tsx
```diff
@@ -448,6 +449,88 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
...
+            : "max-h-0 opacity-0 overflow-hidden",
+        )}
+        aria-hidden={!isExpanded}
+      >
+        <div className="pl-8">
+          <div className="border-l-2 border-muted-foreground p-2">
+            {childMessages?.map((m: TamboThreadMessage) => (
+              <div
+                key={m.id}
+                className={`flex flex-col gap-4 pb-4 ${
```
alecf: similar: since this is the only child, you're basically adding extra bottom padding to the last child - `gap-4` on the parent would generally serve us better

cli/src/registry/message/message.tsx
```diff
@@ -448,6 +449,88 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
...
+                  m.role === "assistant" ? "pl-2" : ""
+                }`}
+              >
+                <span
+                  className={cn(
+                    "whitespace-pre-wrap",
+                    m.role === "assistant"
+                      ? " bg-muted/50 rounded-md p-2 inline-block w-fit"
```
alecf: very minor but you don't need the leading space because `cn` handles that for you

alecf (commented):
a few minor things, otherwise looks good!

cli/src/registry/message/message.tsx
MichaelMilstead: was adding `pl-2` to the items in this column so the text aligns with the text on the 'expand sampling' button, which has `p-2` to show a 'button background' (but the background is only visible on hover, so until hover it just looked indented compared to the other rows.)

Removed those and just added `-ml-2` around the button to move itself back in line

cli/src/registry/message/message.tsx
```diff
@@ -425,8 +425,11 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
               parameters:{"\n"}
               {stringify(keyifyParameters(toolCallRequest?.parameters))}
             </span>
+            <div className="-ml-2">
+              <SamplingSubThread parentMessageId={message.id} />
```
alecf: more minor stuff but:
1) no need to wrap SamplingSubThread with a div - instead let it accept `className` and pass in classes - keeps from creating unused wrapper divs that may be styled inadvertently (though extra gaps, or whatever)
2) the negative margin seems even worse than the padding (margins should generally be avoided, and negative margins even more) - I guess use left padding after all?

cli/src/registry/message/message.tsx
```diff
@@ -448,6 +451,82 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
...
+          isExpanded
+            ? "max-h-96 opacity-100 overflow-auto"
+            : "max-h-0 opacity-0 overflow-hidden",
+        )}
+        aria-hidden={!isExpanded}
+      >
+        <div className="pl-8">
+          <div className="border-l-2 border-muted-foreground p-2 flex flex-col gap-4">
+            {childMessages?.map((m: TamboThreadMessage) => (
+              <div key={m.id} className={`${m.role === "assistant" && "pl-2"}`}>
```
alecf: wait, shouldn't the user messages stay to the right and the assistant messages stay on the left? it feels like this extra padding is kind of doing the reverse?

cli/src/registry/message/message.tsx
```diff
@@ -448,6 +451,82 @@ const ToolcallInfo = React.forwardRef<HTMLDivElement, ToolcallInfoProps>(
...
+        id={samplingDetailsId}
+        className={cn(
+          "transition-[max-height,opacity] duration-300",
+          isExpanded
+            ? "max-h-96 opacity-100 overflow-auto"
+            : "max-h-0 opacity-0 overflow-hidden",
+        )}
+        aria-hidden={!isExpanded}
+      >
+        <div className="pl-8">
```
alecf: I feel like the left border should  stay flush, it looks kinda odd indented IMO

cli/src/registry/message/message.tsx
MichaelMilstead: ah true,  will flip to match the 'main' thread's pattern

Issue #983, opened 2025-09-05
TamboHack: Enable automatic thread name generation


MichaelMilstead:
**Problem:**

Currently, threads created in Tambo are not given a "name" by default. We expose functions that developers can call to either manually rename a thread or to allow Tambo to generate the name of the thread based on the thread's content.

Developers, and end users, expect the thread name generation to happen automatically, but we currently don't have an easy way for a developer to enable this.

**Proposed Solution:**

- update TamboThreadProvider to include state values with reasonable defaults that determine: how many messages in a new thread triggers a name generation, and how many messages since last generation to trigger a name generation as an update

- Update TamboThreadProvider and TamboProvider props to allow a developer to set custom values for the thresholds

- update sendThreadMessage so that when a message is sent, the threshold is checked and, if passed, triggers a call to generateThreadName

**Acceptance Criteria:**

- The number of thread messages to trigger name generation for a new thread is configurable with reasonable default

- The number of messages since last name generation to trigger name generation (as a way to auto-update the name when the conversation topic has shifted) is configurable with reasonable default

- The name of a  thread in the thread history UI component automatically updates

**Bonus:**

- Update api threads/:id/generate-name route so it streams back the name, and update the sdk generateThreadName function to update local state as chunks return

effe-megna:
Hey @MichaelMilstead, I'd like to contribute to this one.

MichaelMilstead:
completed

PR #1185, opened 2025-10-22
feat(react-sdk): auto generate thread name

effe-megna:
## Summary
issue #983 

This PR adds automatic thread name generation functionality to the Tambo React SDK:

- Adds `autoGenerateThreadName` prop (default: `true`) toenable/disable the feature
- Adds `autoGenerateNameThreshold` prop (default: `3`) to configure when names are auto-generated
- Automatically generates thread names when message count reaches the threshold
- Only generates names for threads without existing names
- Updates both local state and threads cache when names are generated
- Includes comprehensive test coverage for all scenarios

## Default behaviour
By default, the auto generate thread name functionality is enabled and the threshold it sets to 3.

## Showcase
https://www.loom.com/share/2b6c8781395a4d96825f8dcc9ee12ff6

### DIFF
--- a/react-sdk/src/providers/__tests__/tambo-thread-provider.test.tsx
+++ b/react-sdk/src/providers/__tests__/tambo-thread-provider.test.tsx
@@ -1659,12 +1659,14 @@ describe("TamboThreadProvider", () => {
               currentPageContextHelper: () => null,
             }}
           >
-            <TamboThreadProvider
-              streaming={false}
-              autoGenerateNameThreshold={2}
-            >
-              {children}
-            </TamboThreadProvider>
+            <TamboMcpTokenProvider>
+              <TamboThreadProvider
+                streaming={false}
+                autoGenerateNameThreshold={2}
+              >
+                {children}
+              </TamboThreadProvider>
+            </TamboMcpTokenProvider>
           </TamboContextHelpersProvider>
         </TamboRegistryProvider>
       );
@@ -1737,13 +1739,15 @@ describe("TamboThreadProvider", () => {
               currentPageContextHelper: () => null,
             }}
           >
-            <TamboThreadProvider
-              streaming={false}
-              autoGenerateThreadName={false}
-              autoGenerateNameThreshold={2}
-            >
-              {children}
-            </TamboThreadProvider>
+            <TamboMcpTokenProvider>
+              <TamboThreadProvider
+                streaming={false}
+                autoGenerateThreadName={false}
+                autoGenerateNameThreshold={2}
+              >
+                {children}
+              </TamboThreadProvider>
+            </TamboMcpTokenProvider>
           </TamboContextHelpersProvider>
         </TamboRegistryProvider>
       );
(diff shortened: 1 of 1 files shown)


react-sdk/src/providers/tambo-thread-provider.tsx
```diff
@@ -462,6 +512,31 @@ export const TamboThreadProvider: React.FC<
...
 
+  const shouldAutoGenerateName = useCallback(
+    (thread: TamboThread): boolean => {
+      // Check if auto-generate is enabled
+      if (!autoGenerateThreadName) {
+        return false;
+      }
+
+      // Check if threshold is valid
+      if (!autoGenerateNameThreshold || thread.id === placeholderThread.id) {
```
MichaelMilstead: let's update this to `if (autoGeneratedNameThreshold == null || thread.id === placeholderThread.id)` so a setting of `0` is allowed

react-sdk/src/providers/tambo-thread-provider.tsx
```diff
@@ -930,12 +1023,23 @@ export const TamboThreadProvider: React.FC<
...
         try {
-          return await handleAdvanceStream(
+          const result = await handleAdvanceStream(
             advanceStreamResponse,
             params,
             threadId,
             contextKey,
           );
+
+          const updatedThread = threadMap[result.threadId];
```
MichaelMilstead: I think we have some weirdness here where the threadMap state hasn't been updated with the messages of this request by this line, so the `updatedThread` contains two fewer messages than it really should, so the name generation runs two messages after the real threshold

effe-megna: Ahh I didn't notice that. Investigating

react-sdk/src/providers/tambo-thread-provider.tsx
effe-megna: Oh nice catch, this has been fixed in the latest commit https://github.com/tambo-ai/tambo/pull/1185/files#diff-d30c2e9c4fcf05004582e87680dc72758dc6135a049226bb72cefe676ce070e3R572

react-sdk/src/providers/tambo-thread-provider.tsx
effe-megna: Okay, to fix it, I had to use `setThreadMap` to ensure the function always has the most up-to-date state. In addition, I added the check in four different places (by calling a new function that contains all the logic):

1. After thread creation from placeholder: For threshold 0, we need to perform the check when the real thread ID is created. Before this point, the threadId is still "placeholder", so the check wouldn’t work.

2. After adding a user message: For existing threads with any threshold (including 0), this check runs after the user adds a message.

3. After streaming completes: This check runs when GenerationStage.COMPLETE is set, after the streaming response finishes.

4. After non-streaming completes: Same as #3, but for the non-streaming code path.

---
What do you think?

MichaelMilstead: It's a little awkward to use set just to access the current state, but I think we need to do a bigger restructure to avoid it, so it can work for now!

MichaelMilstead:
Looks good! Seems like maybe a conflict and linter error to fix but should be all good after

react-sdk/src/providers/tambo-thread-provider.tsx
effe-megna: Yeah it is for sure, but I couldn't find a better way at the moment due to how that state is handled

effe-megna:
Hey @MichaelMilstead thanks for the check, the conflict and the linter error are gone now

MichaelMilstead (approved):
Thank you!

Issue #1189, opened 2025-10-23
Streaming Responses Fail with Failed to parse JSON after multiple chunks.


mikestaub:
I see this when using some OpenAI compatible models.

Likely root cause:

Interactive chat sessions were intermittently aborting when the UI attempted to
stream assistant messages. The React SDK delegated streaming to
@tambo-ai/typescript-sdk’s advanceStream, which tries to JSON.parse each
newline-delimited SSE chunk as it arrives. Under normal network conditions the
server can split a single JSON payload across multiple TCP frames, so the client
occasionally received incomplete JSON and threw Failed to parse JSON after multiple chunks.—terminating the stream and surfacing an error toast.

harshit-seth:
Please Assign me this issue #1189

lachieh:
Hey, thanks for checking, @harshitseth-1. We don't assign issues until someone has at least opened a PR. If you'd like to contribute, just create a PR and so long as it follows the [contribution guidelines](https://github.com/tambo-ai/tambo/blob/main/CONTRIBUTING.md) and meets the requirements of the issue we'll review.

PR #1190, opened 2025-10-23
fix(streaming): add advanceStreamWithBuffer method

mikestaub:
https://github.com/tambo-ai/tambo/issues/1189

# Issue: Streaming Responses Fail with `Failed to parse JSON after multiple chunks.`

## Summary

Interactive chat sessions were intermittently aborting when the UI attempted to
stream assistant messages. The React SDK delegated streaming to
`@tambo-ai/typescript-sdk`’s `advanceStream`, which tries to `JSON.parse` each
newline-delimited SSE chunk as it arrives. Under normal network conditions the
server can split a single JSON payload across multiple TCP frames, so the client
occasionally received incomplete JSON and threw `Failed to parse JSON after
multiple chunks.`—terminating the stream and surfacing an error toast.

## What Changed

- Added a local helper `advanceStreamWithBuffer` (`react-sdk/src/util/advance-stream.ts`)
  that wraps the generated SDK call but buffers `data:` lines until a full SSE
  event is assembled before attempting to parse JSON.
- Updated `tambo-thread-provider` and related tests to use the buffered helper
  so streaming consumers no longer experience partial JSON failures.
- Added unit coverage in `react-sdk/src/__tests__/advance-stream.test.ts` to
  verify we correctly handle split chunks and still surface the original error
  when the payload remains malformed.

## Why It Works

By reassembling the event payload inside the React SDK we ensure `JSON.parse`
only runs on complete SSE messages. This prevents premature failures caused by
network chunking, without waiting for upstream regeneration of the Stainless
SDK. The change is isolated to the client, so existing API contracts remain
unchanged while the UI regains stable streaming behavior.


alecf:
@mikestaub the intention behind *our* api is that we always return fully formed JSON - we're handling the partial JSON response from the llm provider server side, and always (theoretically) emitting correct JSON. 

But that doesn't mean it is always working though! We did have some code in the sdk to just ignore malformed chunks at one point, I wonder if it broke. @MichaelMilstead do you remember some of the issues we were seeing and where the fix to ignore bad chunks was?

alecf:
We never got followup here, I don't think this is still an issue.

Issue #534, opened 2025-06-15
Server Side MCP should allow pass through headers into the MCP server request from client side.


MatthewWilsonThomas:
When you setup a server side MCP Server, the server requires the authentication as normal. 

I am using an MCP server as an OpenAPI spec API endpoint wrapper and I need user authentication for that API based on the react bearer token. My suggestion of a solution is that there are passthrough http headers that are sent to the mcp-server upon request, allowing for this behaviour. 

It would be intuitie to have the MCPProvider allow for this configuration as a way of passing custom headers to server-side MCP servers. 

There is an existing workaround by exposing my MCP endpoint to the browser and using a client side MCP provider to access the url, then passing custom headers to you MCPClient which supports them. 

Ideally there would be a way to support ServerSide MCP endpoints with custom header in the MCPProvider, but it would be equally useful to have a passthrough custom header to client-side provider to support user specific authentication.

PR #1191, opened 2025-10-23
feat(react): add defaultHeaders prop to TamboClientProvider

mikestaub:
Resolves: https://github.com/tambo-ai/tambo/issues/534

### DIFF
--- a/react-sdk/src/providers/tambo-client-provider.tsx
+++ b/react-sdk/src/providers/tambo-client-provider.tsx
@@ -1,12 +1,7 @@
 "use client";
 import TamboAI, { ClientOptions } from "@tambo-ai/typescript-sdk";
 import { QueryClient } from "@tanstack/react-query";
-import React, {
-  createContext,
-  PropsWithChildren,
-  useMemo,
-  useState,
-} from "react";
+import React, { createContext, PropsWithChildren, useMemo } from "react";
 import packageJson from "../../package.json";
 import { useTamboSessionToken } from "./hooks/use-tambo-session-token";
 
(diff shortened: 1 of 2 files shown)


react-sdk/src/providers/tambo-client-provider.tsx
```diff
@@ -133,3 +166,17 @@ export const useIsTamboTokenUpdating = () => {
   }
   return context.isUpdatingToken;
 };
+
+/**
+ * Hook to access the default headers for Tambo API requests
+ * @returns The default headers object
+ */
+export const useTamboDefaultHeaders = () => {
```
alecf: I'd like to avoid adding an extra hook here since we can get it from `useTamboClient()?.defaultHeaders` - is there a particular reason to add a separate hook?

react-sdk/src/providers/tambo-client-provider.tsx
```diff
@@ -36,6 +47,8 @@ export interface TamboClientContextProps {
   queryClient: QueryClient;
   /** Whether the session token is currently being updated */
   isUpdatingToken: boolean;
+  /** Additional headers to include in all requests */
```
alecf, line 45: I think the name `defaultHeaders` implies that it is all the existing header or something, rather than extra headers that we are adding to the existing (default) headers. 

How about `additionalHeaders` or `extraHeaders`

alecf (changes requested):
Thanks for the contribution, this is great! 

I think we need to change the name slightly. Curious @MichaelMilstead or @michaelmagan  if you have any preferences here.

react-sdk/src/providers/tambo-client-provider.tsx
mikestaub: @alecf I addressed both comments

mikestaub:
Can this be merged now?

PR #1192, opened 2025-10-24
feat: add context badge for images

akhileshrangani4:


cli/src/registry/message-input/message-input.tsx
```diff
@@ -415,6 +425,12 @@ const MessageInputTextarea = ({
       const file = item.getAsFile();
       if (file) {
         try {
+          // Mark this file as pasted so we can show "Image 1", "Image 2", etc.
+          Object.defineProperty(file, "wasPasted", {
```
alecf: ah nice - so the way to to do "out of band" properties is is with [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol):

```ts
// define a global symbol
const IS_PASTED_IMAGE = Symbol("is-pasted-image");
```

in your code you can refer to `(file as any)[IS_PASTED_IMAGE]` or optionally define an interface like 
```ts
export interface FileWithPasted {
  [IS_PASTED_IMAGE]?: boolean;
}
```
and cast it

or third option - I *think* you can declare this in this file
```ts
// extend the core type
declare global {
  interface File {
    [IS_PASTED_IMAGE]?: boolean;
  }
}
```

alecf (commented):
nice! One change to use Symbol... but I like the explicitness of this

cli/src/registry/message-input/message-input.tsx
akhileshrangani4: ohh this is nice, updated this.

PR #1217, opened 2025-10-30
feat(mcp): Elicitation support + default UI in showcase

alecf:
image
image


michaelmagan:
Hi @alecf this looks really great. 

Just a few design notes:
1. Submit button should be black like our standard buttons, not green
2. Highlight color on the input should match our primary accent
3. Pull all the cancel/decline buttons to the right side
4. Make the Cancel button red (outline style) so it's clear that's the "no" path
5. The form title  needs to be bigger/bolder i.e. "What are your favorite things?"
6. Is there visual feedback some visual feedback when the pills are selected (maybe fill with black? add a checkmark? can they be multi-select?)
7. I know the numbers can support mins/maxes, I'd be curious to see what that looks like as well.

P.S. We have a well designed form in our component library if you ooint an agent at it to match these patterns better might be a little faster..,

PR #1258, opened 2025-11-05
feat: add tambo context attachment provider

akhileshrangani4:


react-sdk/src/providers/tambo-context-attachment-provider.tsx
```diff
@@ -0,0 +1,281 @@
...
+    };
+  }, [removeContextHelper]);
+
+  const addContextAttachment = useCallback(
+    (context: Omit<ContextAttachment, "id">) => {
+      setAttachments((prev) => {
+        if (prev.some((c) => c.name === context.name)) return prev;
+
+        const newId =
+          typeof crypto !== "undefined" && "randomUUID" in crypto
```
alecf: I think we need to expect crypto to be there - and throwing an error is actually the right move if it isn't.  I did learn that the reason it sometimes isn't present is because you're using an ip address instead of localhost

alecf (approved):
one minor thing...

react-sdk/src/providers/tambo-context-attachment-provider.tsx
akhileshrangani4: ohh good point, I was thinking if I should keep it or not. thanks!

akhileshrangani4: updated it

PR #1259, opened 2025-11-05
feat(showcase/message-input): persist user input via session storage …

mizurex:
### Description
- Added persistence for user input using session storage.
- clears data on submit

### Demo:

https://github.com/user-attachments/assets/a1b49732-401d-4eb1-a453-5bfbc5c7af42

### PR Checklist
- [x] Semantic PR title using Conventional Commits (`feat`, `fix`, `!` for breaking changes)
- [x] Visual change demo video attached (≤ 2 min)
- [x] Showcase updated for component changes

_Note: feel free to close this pr if it's not needed. _

### DIFF
--- a/showcase/src/components/tambo/message-input.tsx
+++ b/showcase/src/components/tambo/message-input.tsx
@@ -405,20 +405,24 @@ const MessageInput = React.forwardRef<HTMLFormElement, MessageInputProps>(
 MessageInput.displayName = "MessageInput";
 
 const STORAGE_KEY = "tambo.components.messageInput.draft";
-const storeValueInSessionStorage = (value?: string) => {
+
+const getStorageKey = (key: string) => `${STORAGE_KEY}.${key}`;
+
+const storeValueInSessionStorage = (key: string, value?: string) => {
+  const storageKey = getStorageKey(key);
   if (value === undefined) {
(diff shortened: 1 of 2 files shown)


michaelmagan:
hey @mizurex . we have another PR tht was close to being merged: https://github.com/tambo-ai/tambo/pull/934#issuecomment-3377759412

If they can't update based on the comment we can work on your PR :)

michaelmagan:
This looks functional but we we would also want doc updates, and updates to our component library but we an wait on that until we are sure if we will go with this PR.

lachieh:
Rebased and updated the integration slightly. Re-requesting review for a second set of eyes.

showcase/src/components/tambo/message-input.tsx
```diff
@@ -404,6 +404,26 @@ const MessageInput = React.forwardRef<HTMLFormElement, MessageInputProps>(
 );
 MessageInput.displayName = "MessageInput";
 
+const STORAGE_KEY = "tambo.components.messageInput.draft";
```
alecf, line 407: I think this needs to be configurable and/or inferred to be unique across different uses. perhaps we could scope the key to the current thread id, so that multiple threads don't share the same text

PR #1261, opened 2025-11-05
feat: Add dictation to showcase

MichaelMilstead:
Updates `message-input.tsx` in showcase to include dictation ui and functionality

Splits `DictationButton` component into its own file so it can be dynamically imported into `MessageInput` on client side to avoid prerendering error with `react-media-recorder`

### DIFF
--- a/cli/src/registry/message-input/dictation-button.tsx
+++ b/cli/src/registry/message-input/dictation-button.tsx
@@ -1,7 +1,7 @@
 import { Tooltip } from "@/components/tambo/suggestions-tooltip";
 import { useTamboThreadInput, useTamboVoice } from "@tambo-ai/react";
 import { Loader2Icon, Mic, Square } from "lucide-react";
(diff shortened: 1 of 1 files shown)


showcase/src/components/ui/dictation-button.tsx
```diff
@@ -0,0 +1,72 @@
...
+  if (isTranscribing) {
+    return (
+      <div className="p-2 rounded-md">
+        <Loader2Icon className="h-5 w-5 animate-spin" />
+      </div>
+    );
+  }
+
+  if (isRecording) {
+    return (
```
alecf: feels strange to do an early return here - because you're switching from a `<button>` to a `<div><span><button>` depending on recording state - layout may be ugly or get formatted differently. 

maybe conditionalize the `<button>` instead of early return?

alecf (approved):
one minor issue with the button (it was there before but I just noticed it this time)

PR #1289, opened 2025-11-11
fix(cli): fix deprecation warning of shell:true

michaelmagan:


michaelmagan:
@CharlieHelps  can you?

Add an error handler to provide a clear message and exit code 127 for ENOENT?

alecf (approved):

PR #1308, opened 2025-11-17
feat(mcp-resources): Handle inline mcp resource references with correct prefix behavior, transforming to resource content nodes

alecf:
Key changes:
 - swapped parent/child order of `<TamboMcpTokenProvider>` and `<TamboRegistryProvider>` so that the registry could be responsible for auto-adding the internal 
 - always prefix the mcp servers, even if there is only 1 (gets rid of some ambiguity when the internal mcp server gets activated, plus just makes the code easier to understand)
 - only parse resource URIs that are prefixed with one of the existing MCP server prefixes (so if you do something like `@https://foo.com` it doesn't interpret that as a resource reference)
 - turn parsed resources into `"resource"` content parts, with prefix properly stripped

### DIFF
--- a/react-sdk/src/mcp/__tests__/elicitation.test.tsx
+++ b/react-sdk/src/mcp/__tests__/elicitation.test.tsx
@@ -1,8 +1,8 @@
 import type { RequestHandlerExtra } from "@modelcontextprotocol/sdk/shared/protocol.js";
 import type {
   ClientNotification,
-  ClientRequest,
   ElicitRequest,
+  Request,
 } from "@modelcontextprotocol/sdk/types.js";
 import { act, renderHook } from "@testing-library/react";
 import {
@@ -12,8 +12,8 @@ import {
 } from "../elicitation";
 
 // Create a mock RequestHandlerExtra for testing
-function createMockExtra(): RequestHandlerExtra<
-  ClientRequest,
+function createMockExtra<T extends Request>(): RequestHandlerExtra<
+  T,
   ClientNotification
 > {
   return {
@@ -67,7 +67,7 @@ describe("useElicitation", () => {
 
       // Start the handler but don't await yet
       let handlerPromise: Promise<TamboElicitationResponse>;
-      const extra = createMockExtra();
+      const extra = createMockExtra<ElicitRequest>();
       act(() => {
         handlerPromise = result.current.defaultElicitationHandler(
           request,
@@ -268,7 +268,7 @@ describe("useElicitation", () => {
       };
 
       let promise2: Promise<TamboElicitationResponse>;
-      const extra2 = createMockExtra();
+      const extra2 = createMockExtra<ElicitRequest>();
       act(() => {
         promise2 = result.current.defaultElicitationHandler(request2, extra2);
       });
--- a/react-sdk/src/mcp/elicitation.ts
+++ b/react-sdk/src/mcp/elicitation.ts
@@ -1,12 +1,11 @@
 import type { RequestHandlerExtra } from "@modelcontextprotocol/sdk/shared/protocol.js";
 import type {
   ClientNotification,
-  ClientRequest,
   ElicitRequest,
   ElicitResult,
 } from "@modelcontextprotocol/sdk/types.js";
 import { useCallback, useState } from "react";
-import { MCPElicitationHandler } from "./mcp-client";
+import { MCPElicitationHandler } from "../model/mcp-server-info";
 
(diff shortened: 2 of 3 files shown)


alecf:
@CharlieHelps a few questions / clarifications on your review: 

> * Inline MCP resource parsing depends on a regex and `serverKey` constraints that aren’t clearly documented; changes to key derivation could silently break parsing.

What would you suggest here?

> Additional notes (13)
> 
> * **Readability** | `react-sdk/src/model/mcp-server-info.ts:52-60`
>   `MCPElicitationHandler`, `MCPHandlers`, and `MCPSamplingHandler` are now defined in `model/mcp-server-info`, but this file still describes `handlers` as being "interpreted as `Partial<MCPHandlers>` in the MCP subpackage". That comment is now misleading because the type is concretely `Partial<MCPHandlers>` everywhere, not an opaque `unknown`.
> 
> Leaving stale documentation here makes the API surface harder to understand and can cause confusion for users trying to pass handlers directly via `McpServerInfo`.

Can you suggest a fix to the comment(s) to reflect this?

> 
> Consider updating the JSDoc to reflect the actual type and semantics (e.g., explicitly reference `MCPHandlers` and note that both the registry and MCP provider expect stable callbacks).

Can you suggest better docs?

> * **Maintainability** | `react-sdk/src/util/message-builder.ts:19-93`
>   `parseResourceMentions` now drives all text splitting and always returns a `text` part for any remaining suffix. For messages like `"@mcp:resource://a"`, this can create an extra empty `text` part if not careful, and for multi-mention text it can create numerous very short `text` fragments.
> 
> Small, empty or near-empty `text` parts are noisy and hard to work with further down the pipeline (rendering, transforms, etc.). It would be cleaner to avoid pushing empty strings and to merge adjacent `text` parts when possible.

Is there an easy fix for this without trimming the conent that gets sent to the server? it's totally fine to have small text fragments, but I'd like to avoid fully empty strings in a content part

> 
> * **Readability** | `react-sdk/src/providers/tambo-registry-provider.tsx:191-191`
>   The MCP tooling no longer injects the internal Tambo MCP server in `TamboMcpProvider`; that logic has moved into `TamboRegistryProvider`. This is a nice separation-of-concerns change, but the inline comment above `TAMBO_INTERNAL_MCP_SERVER_NAME` is currently empty (`/** * */`), and there’s no high-level explanation of how/when the internal server is added.

can you update docs/comments around this?

> 
> * **Maintainability** | `react-sdk/src/mcp/tambo-mcp-provider.tsx:68-68`
>   With the internal MCP server injection moved entirely into `TamboRegistryProvider`, `TamboMcpProvider` no longer has any awareness of `mcpAccessToken` or `tamboBaseUrl`. That’s an improvement in separation of concerns, but you still import the MCP handler types (`MCPElicitationHandler`, `MCPHandlers`, `MCPSamplingHandler`) here even though `normalizeServerInfo` now simply passes through `server.handlers` as-is.
> 
> If the registry is now the single source of truth for MCP server metadata, having type-specific normalization logic in two places (`mcp-server-info` and this provider) is unnecessary and makes the flow harder to trace.

can you propose a consolidation?

> * **Maintainability** | `react-sdk/src/mcp/__tests__/mcp-hooks.test.tsx:967-977`
>   The comment on the resource URI expectation in these tests still talks about "removing prefixes when a server is removed", but you’ve flipped the behavior to always keep prefixes even with a single server.
> 
> Tests that assert new behavior but still describe the old semantics are brittle documentation; they make it harder for maintainers to understand the intention behind the change, especially since this is a behavioural breaking change for consumers.

Can you fix/update the docs/comments for that?

alecf:
@CharlieHelps can you please make all of those proposed changes from your last comment?

react-sdk/src/util/message-builder.ts
```diff
@@ -2,24 +2,111 @@ import type TamboAI from "@tambo-ai/typescript-sdk";
...
+  if (lastIndex < text.length) {
+    parts.push({
       type: "text",
-      text: text.trim(),
+      text: text.slice(lastIndex),
     });
   }
 
+  // If no resources were found, return original text as single part (unless empty string)
+  if (parts.length === 0) {
```
MichaelMilstead, line 112: hmm when would this be true since we add any remaining text above?

MichaelMilstead (approved):

alecf:
@CharlieHelps you did a bunch of work on the docs as requested, but then never committed the changes

PR #1339, opened 2025-11-21
feat(mcp-resources): Fetch resources from MCP servers before sending to the AI SDK

alecf:
> **Note:** This PR is ported from the original tambo-cloud repository: https://github.com/tambo-ai/tambo-cloud/pull/2076

## Summary

This PR implements pre-fetching of MCP resources before sending messages to the AI SDK. Previously, resources were passed as URI references to the LLM, which meant the LLM couldn't access the actual content. Now, resources are fetched from MCP servers and transformed into content that can be consumed by the AI SDK providers.

### Resource Fetching Infrastructure

- **New `resource-transformation.ts` module** (`packages/backend/src/util/resource-transformation.ts`):
  - `prefetchAndCacheResources()` - Fetches all resources from thread messages in parallel before sending to the LLM
  - `extractServerKeyFromResource()` - Extracts the MCP server key from resource URIs (e.g., `github:file.txt` → `github`)
  - Handles fetch errors gracefully without breaking other resource fetches
  - Deduplicates resource URIs to avoid fetching the same resource multiple times

- **Resource Fetcher Map** (`apps/api/src/common/systemTools.ts:createResourceFetcherMap()`):
  - Creates a map of serverKey → fetcher functions from MCP clients
  - Each fetcher validates URIs and delegates to the appropriate MCP client

### AI SDK Integration

- **Enhanced `ai-sdk-client.ts`** (`packages/backend/src/services/llm/ai-sdk-client.ts`):
  - `convertContentPartToUserContent()` - Converts resource content parts to AI SDK-compatible formats
  - Supports multiple formats: text, blob (base64), images, files
  - Falls back to text representation when MIME types aren't supported by the provider
  - Added `getSupportedMimeTypePredicate()` to check which MIME types each provider supports

### Thread Processing

- **Updated thread service** (`apps/api/src/threads/threads.service.ts`):
  - Passes resource fetchers to decision loop and agent loop
  - Fetches MCP clients early in the request lifecycle
  
- **Decision loop integration** (`packages/backend/src/services/decision-loop/`):
  - Both `decision-loop-service.ts` and `agent-loop.ts` now call `prefetchAndCacheResources()` before converting messages
  - Resources are transformed before messages are sent to the LLM

### Type System Updates

- **Enhanced Resource type** (`packages/core/src/threads.ts`):
  - Added `ResourceFetcher` and `ResourceFetchResult` types
  - Resources can now contain either `uri` (for lazy fetching) or inline `text`/`blob` data

- **Content part handling** (`apps/api/src/threads/util/content.ts`):
  - `convertContentDtoToContentPart()` now handles Resource content parts
  - Previously filtered out resource parts; now properly converts them

### MCP Server Updates

- **Resource URI prefixing** (`apps/api/src/mcp-server/resources.ts`):
  - Resources are registered with serverKey-prefixed URIs (e.g., `github:file.txt`)
  - Allows multiple MCP servers to provide resources with the same base URI without conflicts

- **Simplified prompt handling** (`apps/api/src/mcp-server/prompts.ts`):
  - Removed unnecessary JSON schema → Zod conversion
  - Simplified to use plain Zod string schemas for prompt arguments

### Resource Fetching Strategy

The implementation uses a three-phase approach:

1. **Extract**: Scan all messages and collect unique resource URIs that need fetching
2. **Fetch**: Fetch all resources in parallel using appropriate MCP clients
3. **Transform**: Replace URI-based resources with inline content (text/blob)

This approach:
- Handles duplicates efficiently (fetch once, use many times)
- Gracefully handles fetch failures (logs errors, continues with other resources)
- Keeps async work in one phase, then does synchronous message updates

### MIME Type Handling

Resources are converted to different AI SDK formats based on MIME type and provider capabilities:
- Images → `ImagePart` (for supported image formats)
- Binary files → `FilePart` (for supported file types)
- Text → `TextPart` with formatted metadata (fallback for unsupported types)

## Testing

All existing tests updated to pass empty resource fetcher maps (`{}`).

---

### DIFF
--- a/packages/backend/src/util/resource-transformation.ts
+++ b/packages/backend/src/util/resource-transformation.ts
@@ -52,28 +52,24 @@ export function extractServerKeyFromResource(resource: Resource): string {
 }
 
 /**
- * Pre-fetch all resources from thread messages and cache them inline.
- * This converts URI-based resources to cached text/blob resources.
- *
- * Strategy:
- * 1. Extract all unique resource URIs from messages
- * 2. Fetch them in parallel into a map (uri -> content)
- * 3. Apply the map to messages to cache content inline
- *
- * This approach:
- * - Handles duplicates efficiently (fetch once, use many times)
- * - Gracefully handles fetch failures
- * - Keeps async work in one phase, then does synchronous message updates
+ * Phase 1: Extract all unique resource URIs that need fetching from messages.
+ * Identifies resources with URIs but no cached content and maps them to their fetchers.
  *
  * @param messages - Thread messages potentially containing resources
  * @param resourceFetchers - Map of serverKey to fetch functions
- * @returns New messages with fetched resources cached as text/blob
+ * @returns Map of URIs to fetch with their associated metadata
  */
-export async function prefetchAndCacheResources(
+function extractUrisToFetch(
   messages: ThreadMessage[],
   resourceFetchers: ResourceFetcherMap,
-): Promise<ThreadMessage[]> {
-  // Phase 1: Extract all unique resource URIs that need fetching
+): Map<
+  string,
+  {
+    uri: string;
+    serverKey: string;
+    fetchFn: ResourceFetcher;
+  }
+> {
   const urisToFetch = new Map<
     string,
     {
@@ -128,7 +124,26 @@ export async function prefetchAndCacheResources(
     }
   }
 
-  // Phase 2: Fetch all unique URIs in parallel
+  return urisToFetch;
+}
+
+/**
+ * Phase 2: Fetch all unique URIs in parallel.
+ * Gracefully handles fetch failures without breaking other fetches.
+ *
+ * @param urisToFetch - Map of URIs to fetch with their metadata
+ * @returns Map of successfully fetched content indexed by URI
+ */
+async function fetchResourceContents(
+  urisToFetch: Map<
+    string,
+    {
+      uri: string;
+      serverKey: string;
+      fetchFn: ResourceFetcher;
+    }
+  >,
+): Promise<Map<string, ResourceFetchResult>> {
   const fetchedContent = new Map<string, ResourceFetchResult>();
 
   await Promise.all(
@@ -145,7 +160,21 @@ export async function prefetchAndCacheResources(
     }),
   );
 
-  // Phase 3: Apply cached content to messages
+  return fetchedContent;
+}
+
+/**
+ * Phase 3: Apply cached content to messages.
+ * Replaces resource parts with fetched content where available.
+ *
+ * @param messages - Original thread messages
+ * @param fetchedContent - Map of fetched content indexed by URI
+ * @returns New messages with fetched resources cached as text/blob
+ */
+function applyCachedContentToMessages(
+  messages: ThreadMessage[],
+  fetchedContent: Map<string, ResourceFetchResult>,
+): ThreadMessage[] {
   return messages.map((message) => {
     if (!Array.isArray(message.content)) {
       return message;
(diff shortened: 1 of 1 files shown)


packages/backend/src/util/resource-transformation.ts
```diff
@@ -0,0 +1,223 @@
...
+ *
+ * @param messages - Thread messages potentially containing resources
+ * @param resourceFetchers - Map of serverKey to fetch functions
+ * @returns New messages with fetched resources cached as text/blob
+ */
+export async function prefetchAndCacheResources(
+  messages: ThreadMessage[],
+  resourceFetchers: ResourceFetcherMap,
+): Promise<ThreadMessage[]> {
+  // Phase 1: Extract all unique resource URIs that need fetching
```
MichaelMilstead: I wonder if we should split the phases into separate functions

packages/backend/src/util/resource-transformation.ts
```diff
@@ -0,0 +1,223 @@
...
+      }
+
+      // Skip if we've already queued this URI
+      if (urisToFetch.has(resource.uri)) {
+        continue;
+      }
+
+      // Find fetcher for this resource's serverKey
+      const serverKey = extractServerKeyFromResource(resource);
+      const fetchFn = serverKey ? resourceFetchers[serverKey] : undefined;
```
MichaelMilstead, line 114: I'm wondering about the scenario where a message has a URI of a resource from an MCP server that was later disconnected. I guess the server wouldn't exist in the `resourceFetchers` map and so here we set the `fetchFn` to undefined, and eventually this resource part is silently removed from the messages sent to the LLM.

I wonder if that could cause the LLM to become confused in some scenarios.
For example maybe the original conversation was:
```
user: look at this document and tell me if it has any typos <some document resource from an mcp server>
llm: Yes, looks good

``` 

And then later, after the server is not connected to the project:

```
user: look at this document and tell me if it has any typos <blank>
llm: Yes, looks good

user: Somebody said there was a typo, can you look again?
llm: ???
```

Wondering if we should use some default resource content in this situation like "This resource is unavailable, since the MCP server that provides it is no longer connected"

MichaelMilstead (approved):

packages/backend/src/util/resource-transformation.ts
alecf: yeah I think this is where caching is also going to come into play - for now I'd like to punt on this specific case because we'll definitely address it in the next week as we flush out some of the resource edge cases

packages/backend/src/util/resource-transformation.ts
alecf: done

packages/backend/src/services/llm/ai-sdk-client.ts
```diff
@@ -771,6 +768,184 @@ function convertOpenAIMessageToCoreMessage(
...
+        return true;
+      }
+      // Handle 'image/*' wildcard
+      if (pattern.endsWith("*")) {
+        return mimeType.startsWith(pattern.slice(0, -1));
+      }
+      // Handle exact match
+      return mimeType === pattern;
+    });
+  };
```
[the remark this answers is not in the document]
alecf: not worried about "legacy" providers here

packages/backend/src/services/llm/ai-sdk-client.ts
```diff
@@ -771,6 +768,184 @@ function convertOpenAIMessageToCoreMessage(
...
+    .join(" ");
+
+  const result = {
+    type: "text" as const,
+    text: `
+<resource ${resourceProps}>
+${resourceData.text}
+</resource>
+`,
+  };
```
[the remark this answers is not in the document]
alecf: this is written as intended

PR #1340, opened 2025-11-21
refactor(api): simplify MCP authentication endpoints and token detection

alecf:
## Summary

Simplifies MCP authentication by consolidating three endpoints into one unified endpoint with conditional behavior and enhanced response metadata.

### Unified Endpoint
- **Consolidated** `/auth/mcp`, `/auth/mcp/refresh`, and `/auth/mcp/sessionless` into single `/auth/mcp/access-token` endpoint
- **Conditional behavior**:
  - With `threadId`: Creates session-bound token (validates thread exists for project)
  - Without `threadId`: Creates sessionless token using contextKey
- **Flexible auth**: contextKey can be provided in request body OR extracted from Bearer token

### Enhanced Response
- Added `expiresAt` field (milliseconds since epoch) - frontend doesn't need to decode JWT
- Added `hasSession` field (boolean) - indicates if token has thread association
- AuthService now returns `McpAccessTokenResult` interface with full metadata

### Bug Fixes
- **Fixed systemTools.ts**: Pass MCP handlers based on `threadId` presence, not `sessionId`
  - Handlers are needed when creating client for a thread, even if no session exists yet
  - MCP server may create new session and needs handlers registered upfront
- Updated test mocks to match new return types

### Request DTO
```typescript
{
  threadId?: string,      // Optional: if provided, creates session-bound token
  contextKey?: string     // Optional: if provided, used for sessionless token; else extracted from Bearer
}
```

### Response DTO
```typescript
{
  mcpAccessToken?: string,  // JWT token
  expiresAt?: number,       // Expiration in ms since epoch
  hasSession?: boolean      // Whether token is bound to a thread
}
```

## Testing
- ✅ All 247 tests pass
- ✅ TypeScript type checks pass
- ✅ Fixed pre-existing systemTools test failures


apps/api/src/common/services/auth.service.ts
```diff
@@ -64,4 +64,46 @@ export class AuthService {
...
+   *
+   * This token is not tied to a specific thread and cannot use session-specific features
+   * (elicitation, sampling). It is primarily used for accessing resources and prompts.
+   *
+   * Notes on expiry/iat:
+   * - We round down to the nearest 5 minutes and set `iat` to that window
+   *   start so tokens minted within the same 5‑minute window are identical.
+   * - `exp` is set to 15 minutes after the window start.
+   */
+  async generateSessionlessMcpAccessToken(
```
MichaelMilstead: I think a lot of this function is identical to logic in `generateMcpAccessToken`, maybe we can extract the common parts

apps/api/src/mcp-server/server.ts
```diff
@@ -79,9 +81,72 @@ export async function createMcpServer(
...
+    },
+  );
+
+  // Session-less servers don't support elicitation/sampling, so no handlers needed
+  // We create MCP clients without thread context
+  const mcpClients = await getThreadMCPClients(
+    db,
+    projectId,
+    // Use empty string for threadId to indicate session-less mode
+    "",
```
MichaelMilstead: Seems like passing "" here for `threadId` causes an error where in `getThreadMCPClients` we create a new `MCPClient` for an attached server which comes with a new `sessionId`, and then we try to call `operations.updateMcpThreadSession` with "" as the `threadId`

alecf: yep, fixed

apps/api/src/common/services/auth.service.ts
alecf: done

PR #1359, opened 2025-11-22
fix(docker): Fix docker build

alecf:
- **no more subdir for tambo-cloud**
- **fix docker build**


alecf:
oops just saw @lachieh has a fix in #1357

Issue #843, opened 2025-08-15
Allow toolcall limit per tool


MichaelMilstead:
## Problem

Currently a developer can define how many toolcalls are allowed when responding to one user message at the project level through the project dashboard.

Sometimes, certain tools should be allowed to be called many times, while others should be only allowed to be called a few times or even once.

## Acceptance Criteria

- registerTool has an optional parameter to set maxCalls for a tool
- API is updated to handle this parameter, and consider this when calling the tool
- If this value exists on a tool, it overrides the 'project' setting
- Updated documentation to explain how to use this parameter and how it overrides the project setting

sidhuiwnl:
how can i get the tool limit set in dashboard data ?
I have trying to find and i couldn't find the data @MichaelMilstead

MichaelMilstead:
In the tambo-cloud repo, we validate that a toolcall request does not take us above the toolcall limit of the project here in `validateToolCallLimits`: https://github.com/tambo-ai/tambo-cloud/blob/3f1d4057cb4d930fbd8953fecd20838505e76e39/apps/api/src/threads/util/tool-call-tracking.ts#L43

@sidhuiwnl

michaelmagan:
This appears to be fixed by PR #1598 (feat(react-sdk): implement toolcall limit per tool), merged on 2026-01-05.

The PR adds optional `maxCalls` property to `TamboTool` and `ComponentContextToolMetadata` interfaces, enabling per-tool limit configuration from the client side.

https://github.com/tambo-ai/tambo/pull/1598

PR #1360, opened 2025-11-22
feat(api): support per-tool maxCalls override for tool-call limits

rakesh-paulraj1:
Added server-side support for per-tool call limits, allowing a tool’s `maxCalls` to override the global `maxToolCallLimit` (affecting total tool calls only, not identical-call limits).

* Preserved `MAX_IDENTICAL_TOOL_CALLS` protection and validator/wrapper separation.
* Validator now supports optional per-tool counts and a `toolLimits` map.
* Added helper to aggregate per-tool counts from signature counters.
* `ThreadsService` derives and passes a normalized `toolLimits` map from tool metadata for both streaming and non-stream flows.
* Implemented `maxCalls` normalization/validation (handle numeric strings, truncate floats, reject negative/NaN).
* Added unit tests for aggregation, per-tool overrides, and provided `perToolCounts`.
* **Fixes:** [[tambo-ai/tambo#843](https://github.com/tambo-ai/tambo/issues/843)](https://github.com/tambo-ai/tambo/issues/843) — updates the API to accept and enforce per-tool `maxCalls`.


rakesh-paulraj1:
please review @michaelmagan

apps/api/src/threads/threads.service.ts
```diff
@@ -2255,3 +2261,56 @@ async function syncThreadStatus(
     },
   );
 }
+/**
+ * Extracts per-tool maxCalls limits from the unified tool registry.
+ * Returns a map of tool name → { maxCalls?: number }.
+ * MCP tools take precedence over client tools when names collide.
+ */
+function deriveToolLimitsFromRegistry(
```
MichaelMilstead: I don't think the tools in this `ToolRegistry` type have `maxCalls`, but I also am not sure we need to derive the limits this way.

I think we can just derive the limits from the `advanceRequestDto` passed in above

rakesh-paulraj1: My bad i thought convertMetadataToTools handles the maxcalls,now changed to `deriveToolLimitsFromDto` wher it accepts the originalrequest which is of type `AdvanceThreadDto` which has the maxCalls type inside `ComponentContextToolMetadataDto`

rakesh-paulraj1: please review @michaelmagan

apps/api/src/threads/util/tool-call-tracking.ts
```diff
@@ -83,6 +108,28 @@ export function validateToolCallLimits(
   return undefined;
 }
 
+/**
+ * Compute per-tool counts from signature-based counts. Signature format is the
+ * JSON produced by createToolCallSignature, which includes toolName at top
+ * level.
+ */
+export function computePerToolCountsFromSignatures(
```
MichaelMilstead: I believe this function isn't working given the parameter (and might not be needed). The incoming `currentToolCounts` looks like:

```
{
"test-tool": 1
}
```

We might just be able to skip this per-tool count calculation since we already have it in `validateToolCallLimits`

rakesh-paulraj1: `currentToolCounts` already contains the correct per-tool counts passed in from the validator. We now derive limits directly from the DTO, so the registry-based logic and this extra counting step are no longer needed I have removed the `computePerToolCountsFromSignatures`

apps/api/src/threads/util/tool-call-tracking.ts
```diff
@@ -57,26 +57,51 @@ export function validateToolCallLimits(
...
+    if (totalCalls >= maxToolCallLimit) {
+      return `I've reached the maximum number of tool calls (${maxToolCallLimit}). This usually indicates I'm stuck in a loop. Please try a different approach or contact support if this persists.`;
+    }
   }
 
-  const signature = createToolCallSignature(newToolCallRequest);
-  const currentCount = currentToolCounts[signature] || 0;
+  // Signature-level identical-call protection (independent of per-tool/project limits)
+  // Note: this check uses the per-tool count as a proxy for identical calls
+  const currentToolTotal = resolvedPerToolCounts[toolName] || 0;
```
MichaelMilstead: since this value isn't "signature specific" like it was before (with `createToolCallSignature`), the count is increased for every call of the same tool, even if the params are different, so it's impossible to have a tool run more than `MAX_IDENTICAL_TOOL_CALLS` in a row, even if they aren't identical.

I think we should add back the `createToolCallSignature`

rakesh-paulraj1: fixed it by feeding `newToolCallRequest` straight into `createToolCallSignature`. That way we get the exact signature that mirrors what we store in currentToolCounts, so the identical-call check only triggers when the same tool name and parameter set repeats and hits `MAX_IDENTICAL_TOOL_CALLS`

rakesh-paulraj1: any updates @MichaelMilstead

MichaelMilstead (approved):
awesome!

PR #1398, opened 2025-11-24
fix: remove internal displayMessage tool and use 'auto' toolChoice

MichaelMilstead:
- Switches the `toolChoice` parameter on LLM requests from 'required' to 'auto'.
- Removes the internal `displayMessage` tool.
- Removes logic to turn a `displayMessage` tool call into a plain response message..

Previously, rather than letting the LLM decide when to show a message and when to use a tool, we forced a tool choice on every request, and included an internal "displayMessage" tool for scenarios where the LLM just wanted to show a text message to the user.

Originally we used this strategy because we found that there were times the LLM would return a message saying that it would call a tool, but would not actually call one.

Recently, the situation has reversed (at least with GPT-5 on 'minimal' reasoningEffort). The model often uses the `displayMessage` tool to say it will take some action, but does not use any other tool to perform that action. No amount of prompting would get the model to "understand" that using the `displayMessage` tool means it is not using another tool.

This change seems to help avoid the model saying it will do something and then not calling a tool to do that thing. Maybe because of model improvements, or because of ai-sdk usage.

### DIFF
--- a/packages/backend/src/services/decision-loop/decision-loop-service.ts
+++ b/packages/backend/src/services/decision-loop/decision-loop-service.ts
@@ -26,7 +26,6 @@ import {
 } from "../llm/llm-client";
 import {
   addParametersToTools,
-  displayMessageTool,
   filterOutStandardToolParameters,
   standardToolParameters,
   TamboToolParameters,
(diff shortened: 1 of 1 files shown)


alecf (approved):
this is great, glad to see this tool gone!

PR #1399, opened 2025-11-24
chore(conductor): update setup script for merged monorepo

michaelmagan:
- Add .env file setup for all merged packages (apps/web, apps/api, packages/db, apps/docs-mcp)
- Add conductor-run.sh with interactive menu for different dev server options
- Auto-start Supabase when running services that need database
- Remove verbose next steps output from setup script

### DIFF
--- a/conductor-run.sh
+++ b/conductor-run.sh
@@ -5,7 +5,8 @@ set -e
 start_supabase_if_needed() {
   if command -v supabase &> /dev/null; then
     echo "🗄️  Checking Supabase status..."
-    if ! supabase status &> /dev/null; then
+    # Use timeout to prevent hanging if Docker isn't running
+    if ! timeout 10 supabase status &> /dev/null; then
       echo "🗄️  Starting Supabase..."
       if ! supabase start; then
         echo "❌ Failed to start Supabase. Common issues:"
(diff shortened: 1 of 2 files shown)


conductor-run.sh
```diff
@@ -0,0 +1,76 @@
...
+    npm run dev
+    ;;
+  2)
+    echo "Starting framework only (showcase + docs)..."
+    npx turbo dev --filter=@tambo-ai/showcase --filter=@tambo-ai/docs
+    ;;
+  3)
+    echo "Starting cloud only (web + api)..."
+    start_supabase_if_needed
+    npx turbo dev --filter=web --filter=api
```
MichaelMilstead: workspace name was updated so should be
`npx turbo dev --filter=@tambo-ai-cloud/web --filter=@tambo-ai-cloud/api`

(and same in the below commands for web and api)

conductor-run.sh
```diff
@@ -0,0 +1,76 @@
...
+echo "  6) Web only"
+echo "  7) API only"
+echo ""
+read -p "Enter your choice (1-7): " choice
+
+case $choice in
+  1)
+    echo "Starting full dev (everything)..."
+    start_supabase_if_needed
+    npm run dev
```
MichaelMilstead: I think right now this script runs only showcase and docs. We should update that anyway so maybe this PR can include the update to the root `package.json`

conductor-run.sh
michaelmagan: fixed!

conductor-run.sh
michaelmagan: also fixed!

MichaelMilstead (approved):

PR #1401, opened 2025-11-25
refactor: consolidate config packages and improve async error handling

alecf:
## Summary

This PR consolidates duplicate configuration packages (typescript and eslint) and improves async/await error handling patterns across the codebase. (since when we brought the async/await rules into the main tambo repo, it found a bunch of issues in the sdk)

90% of this is pretty rote changes - narrower casts, returning `{}` when previously we returned `null`, etc

But the 2 big changes you'll see are:
1) when we had nested ternaries, I had claude break out the whole logic into a separate component (since they all occurred nested in JSX) so rather than `a ? b : c ? d : e` we have a component in the same file with early returns on each condition
2) there were a number of places that we were either not awaiting promises, or had some odd logic that bypassed awaiting promises. I cleaned up what I could without breaking anything but some of this was in our SDK. The next improvement is a good one but you'll see a lot of inline `.catch()` that we didn't have before. I added guidance to AGENTS.md around this too.

---

### Config Package Consolidation

**Merged TypeScript Config:**
- Consolidated `tambo-cloud/packages/typescript-config` into root `packages/typescript-config`
- Kept modern ES2024 target and added sourceMap option
- Updated 8 package.json files to use `@tambo-ai/typescript-config`
- Updated 7 tsconfig.json files to extend from root config
- Removed duplicate package (-336 lines)

**Merged ESLint Config:**
- Consolidated `tambo-cloud/packages/eslint-config` into root `packages/eslint-config`
- Added stricter rules from tambo-cloud version:
  - `@typescript-eslint/no-unnecessary-*` rules
  - `@typescript-eslint/no-floating-promises`
  - `no-nested-ternary`
- Added MIT license to package.json
- Updated all package.json and eslint.config.mjs files across workspace
- Added `lint:fix` script to 6 packages that were missing it
- Removed duplicate package (-202 lines)

**Moved Supabase:**
- Moved `tambo-cloud/supabase/` to root `supabase/`

### Async Error Handling Improvements (React SDK)

**Replaced void with async/await:**
- Converted floating promises in non-test files to proper async/await patterns
- Made `maybeAutoGenerateThreadName` an async callback
- Updated `setValue` in use-component-state.tsx to maintain API compatibility

**Error handling principles:**
- Use `.catch()` only in useEffect contexts where unavoidable (last resort)
- Use `void` operator for fire-and-forget calls that already have error handling inside
- Make callbacks async when possible to allow proper error propagation
- Avoid double error handling (try/catch inside + .catch() outside)

**Updated AGENTS.md:**
- Documented `.catch()` usage guidelines
- Added async callback pattern preferences

## Test Plan

- ✅ All TypeScript checks pass (`npm run check-types`)
- ✅ All ESLint checks pass (`npm run lint`)
- ✅ All tests pass (`npm test`)
- ✅ Build successful across all packages

## Breaking Changes

None - all changes are backwards compatible.

### DIFF
--- a/cli/src/commands/init.ts
+++ b/cli/src/commands/init.ts
@@ -56,13 +56,10 @@ interface InitOptions {
  */
 async function writeApiKeyToEnv(apiKey: string): Promise<boolean> {
   try {
-    let targetEnvFile: string;
-    if (fs.existsSync(".env.local")) {
-      targetEnvFile = ".env.local";
-    } else if (fs.existsSync(".env")) {
+    // use `.env.local` by default or fall back to .env if it exists and .env.local does not
+    let targetEnvFile = ".env.local";
+    if (fs.existsSync(".env") && !fs.existsSync(".env.local")) {
       targetEnvFile = ".env";
-    } else {
-      targetEnvFile = ".env.local"; // default when neither exists
     }
 
     const envContent = `\nNEXT_PUBLIC_TAMBO_API_KEY=${apiKey.trim()}\n`;
--- a/showcase/src/components/tambo/map.tsx
+++ b/showcase/src/components/tambo/map.tsx
@@ -509,13 +509,11 @@ export const Map = React.forwardRef<HTMLDivElement, MapProps>(
                   medium: "w-10 h-10 text-sm",
                   large: "w-12 h-12 text-base",
                 };
-              let iconSize: number;
+              let iconSize = 48;
               if (size === "small") {
                 iconSize = 32;
               } else if (size === "medium") {
                 iconSize = 40;
-              } else {
-                iconSize = 48;
               }
               return L.divIcon({
                 html: `<div class="flex items-center justify-center ${colorClass} ${sizeClasses[size]} text-white font-bold rounded-xl border-2 border-white shadow-lg transition-all duration-200 hover:scale-110 hover:brightness-90">${count}</div>`,


MichaelMilstead (approved):
love the component splits!

apps/test-mcp-server/eslint.config.mjs
```diff
@@ -1,4 +1,4 @@
-import config from "@tambo-ai-cloud/eslint-config/base";
+import config from "@tambo-ai/eslint-config/base";
 import tseslint from "typescript-eslint";
 
 export default tseslint.config(...config);
```
lachieh, line 4: The config helper is [deprecated](https://typescript-eslint.io/packages/typescript-eslint/#config-deprecated) in favor of eslint's `defineConfig`. Not critical for this PR but might be worthwhile swapping it out while you're in there?


```suggestion
import { defineConfig } from "eslint/config";
import config from "@tambo-ai/eslint-config/base";

export default defineConfig(...config);
```

cli/src/commands/init.ts
```diff
@@ -56,11 +56,14 @@ interface InitOptions {
...
-        ? ".env"
-        : ".env.local"; // default when neither exists
+    let targetEnvFile: string;
+    if (fs.existsSync(".env.local")) {
+      targetEnvFile = ".env.local";
+    } else if (fs.existsSync(".env")) {
+      targetEnvFile = ".env";
+    } else {
+      targetEnvFile = ".env.local"; // default when neither exists
+    }
```
lachieh:
```suggestion
    // use `.env.local` by default or fall back to .env if it exists and .env.local does not
    let targetEnvFile = ".env.local";
    if (fs.existsSync(".env") && !fs.existsSync(".env.local")) {
      targetEnvFile = ".env";
    }
```

cli/src/registry/map/map.tsx
lachieh, line 521: I think I like this whole function refactored:

```ts
(cluster: L.MarkerCluster) => {
  const count = cluster.getChildCount();
  let size: "small" | "medium" | "large" = "small";
  if (count > 100) {
    size = "large";
  } else if (count > 10) {
    size = "medium";
  }
  const { colorClass, sizeClasses, iconSize } = {
    small: {
      colorClass: "bg-blue-500",
      sizeClasses: "w-8 h-8 text-xs",
      iconSize: 32
    },
    medium: {
      colorClass: "bg-orange-500",
      sizeClasses: "w-10 h-10 text-sm",
      iconSize: 40
    },
    large: {
      colorClass: "bg-red-500",
      sizeClasses: "w-12 h-12 text-base",
      iconSize: 48
    },
  }[size];

  return L.divIcon({
    html: `<div class="flex items-center justify-center ${colorClass} ${sizeClasses[size]} text-white font-bold rounded-xl border-2 border-white shadow-lg transition-all duration-200 hover:scale-110 hover:brightness-90">${count}</div>`,
    className: "custom-cluster-icon",
    iconSize: L.point(iconSize, iconSize),
    iconAnchor: L.point(iconSize / 2, iconSize / 2),
  });
}
```

packages/backend/src/services/llm/agent-client.test.ts
```diff
@@ -81,6 +81,7 @@ describe("AgentClient", () => {
 
     it("should create a Mastra agent with name", async () => {
       jest.spyOn(MastraAgent, "getRemoteAgents").mockResolvedValue({
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
lachieh, line 84: I think I might be ok with turning this rule off in `*.test.ts(x)` files. Or maybe in `**/__mock__/*.ts` files?

showcase/src/components/tambo/map.tsx
```diff
@@ -519,8 +509,14 @@ export const Map = React.forwardRef<HTMLDivElement, MapProps>(
...
-              const iconSize =
-                size === "small" ? 32 : size === "medium" ? 40 : 48;
+              let iconSize: number;
+              if (size === "small") {
+                iconSize = 32;
+              } else if (size === "medium") {
+                iconSize = 40;
+              } else {
+                iconSize = 48;
+              }
```
lachieh, line 517: Same here as the other `iconCreateFunction` prop change

lachieh:
Oh, I forgot this bit! The docker build is failing because it previously needed the `tambo-cloud/packages` files for the `turbo prune` step, so now that it's gone, you can delete this line:

https://github.com/tambo-ai/tambo/blob/5e4ecee32f0ff90269fbbf5624053f4af18038ad/apps/web/Dockerfile#L22

apps/test-mcp-server/eslint.config.mjs
alecf: Yeah I agree - there are some subtle differences for how rules are merged when you use defineConfig - I'll do a separate PR so I can get this landed

packages/backend/src/services/llm/agent-client.test.ts
alecf: that's a good call - we've been a little inconsistent across the repos about where tests  go (i.e. inside __tests__ or as *.spec.ts(x) etc) so I'll see if I can address that in a separate PR and normalize there

showcase/src/components/tambo/map.tsx
alecf: @CharlieHelps can you simplify this slightly by just adding 48 as the default value in the `let`?

PR #1410, opened 2025-11-26
feat: return partial (building) toolcall requests to client

MichaelMilstead:
- updates `runDecisionLoop` to allow incomplete toolcall requests yielded (after being "closed off" by using partial-parse with `parse`
- updates `fixStreamedToolCalls` to add extra "internal" fields to hold in-progress toolcall info, while still waiting for the complete toolcallrequest for existing fields
- updates `updateThreadMessageFromLegacyDecision` to add in-progress toolcall info to the `.component` field

React SDK is already setup to check `.component.toolCallRequest` to determine whether a toolcall is building and show loading UI, and will not attempt to run a toolcall until the outer `.toolCallRequest` field is filled

### DIFF
--- a/packages/backend/src/services/decision-loop/decision-loop-service.ts
+++ b/packages/backend/src/services/decision-loop/decision-loop-service.ts
@@ -216,6 +216,10 @@ function buildToolCallRequest(
 ): ToolCallRequest | undefined {
   // "custom" tool calls are not supported
   if (toolCall.type !== "function") {
+    console.warn(
+      "Unsupported tool call type, only 'function' is supported, received: ",
+      toolCall.type,
+    );
     return undefined;
   }
 
@@ -223,18 +227,21 @@ function buildToolCallRequest(
   let parsedArgs: Record<string, unknown> | undefined;
   try {
     // Use partial-json parse to handle incomplete JSON during streaming
-    const parsed = parse(toolCall.function.arguments);
-    if (!parsed || typeof parsed !== "object" || Array.isArray(parsed)) {
-      return undefined;
-    }
-
-    parsedArgs = parsed;
+    parsedArgs = parse(toolCall.function.arguments);
   } catch (_e) {
     // If parsing fails completely, we can't build a request
+    console.warn(
+      "Failed to parse tool call arguments, received: ",
+      toolCall.function.arguments,
+    );
     return undefined;
   }
-
-  if (!parsedArgs) {
+  if (
+    !parsedArgs ||
+    typeof parsedArgs !== "object" ||
+    Array.isArray(parsedArgs)
+  ) {
+    console.warn("Invalid tool call arguments, received: ", parsedArgs);
     return undefined;
   }
 


apps/api/src/threads/util/thread-state.ts
```diff
@@ -346,6 +368,23 @@ export function updateThreadMessageFromLegacyDecision(
...
+  const toolCallId = chunk.toolCallId ?? chunkWithHidden.__toolCallId;
+
+  // Always include tool call info in component field if present
+  const componentWithToolCall: LegacyComponentDecision = {
+    ...simpleDecisionChunk,
+    toolCallRequest,
+    toolCallId,
+  };
+
   const currentThreadMessage: ThreadMessage = {
```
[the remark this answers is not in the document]
MichaelMilstead: will update later

apps/api/src/threads/util/thread-state.ts
```diff
@@ -280,72 +280,116 @@ export async function addAssistantResponse(
...
 /**
- * The purpose if this function is to make sure that we do not stream
- * intermediate tool calls and instead withhold the toolCallRequest and
- * toolCallId until the final message.
+ * Type for LegacyComponentDecision with internal fields used during streaming.
+ * These fields are used to preserve tool call information during streaming
+ * before it's moved to the outer fields on the final chunk.
+ */
+export type DecisionWithInternalToolInfo = LegacyComponentDecision & {
+  __toolCallRequest?: ToolCallRequest;
```
alecf: It feels strange to prepend some underscores to imply "unfinished" - can this just be `unfinishedToolCallRequest` and `unfinishedToolCallId`?

alecf: It also feels really complicated to emit these as separate fields and then clear out the other fields. 

Maybe we can just add a single boolean indicating if it is finished or not? I suspect that will clean up the loop below significantly, since you can just emit every message as-is but with isToolCallFinished: false, and then just emit one with `isToolCallFinished: true`

apps/api/src/threads/util/thread-state.ts
```diff
@@ -280,72 +280,116 @@ export async function addAssistantResponse(
...
   // we explicitly remove certain fields from the component decision to avoid
   // duplication, because they appear in the thread message
   const { reasoning, ...simpleDecisionChunk } = chunk;
+
+  // Get tool call info - check both outer fields and hidden properties (for streaming chunks)
+  const toolCallRequest = chunk.toolCallRequest ?? chunk.__toolCallRequest;
+  const toolCallId = chunk.toolCallId ?? chunk.__toolCallId;
+
+  // Always include tool call info in component field if present
+  const componentWithToolCall: LegacyComponentDecision = {
```
alecf: I'm a little surprised this spread works, it seems like you're introducing the extra fields, `__toolCallRequest`, etc into componentWithTool call, and then emitting that as `component`, leaking these extra fields out to the API?

alecf (commented):
before we go further, I have a suggestion below that I think may simplify a lot of this code...

apps/api/src/threads/util/thread-state.ts
```diff
@@ -280,62 +280,84 @@ export async function addAssistantResponse(
...
-        toolCallId: currentToolCallId,
+        toolCallRequest: finalToolCallRequest,
+        toolCallId: finalToolCallId,
       };
-      // and clear the current tool call request and id
+      // and clear the current state
       currentToolCallRequest = undefined;
       currentToolCallId = undefined;
+      finalToolCallRequest = undefined;
+      finalToolCallId = undefined;
```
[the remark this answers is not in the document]
MichaelMilstead: removed

apps/api/src/threads/util/thread-state.ts
```diff
@@ -280,62 +280,84 @@ export async function addAssistantResponse(
...
+      // Store original values for component field access
+      __toolCallRequest: chunk.toolCallRequest,
+      __toolCallId: chunk.toolCallId,
+    } as LegacyComponentDecision & {
+      __toolCallRequest?: ToolCallRequest;
+      __toolCallId?: string;
+    };
+
+    currentDecision = streamingChunk as LegacyComponentDecision;
+    yield streamingChunk as LegacyComponentDecision;
```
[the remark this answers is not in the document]
MichaelMilstead: created type `DecisionWithInternalToolInfo `

packages/backend/src/services/decision-loop/decision-loop-service.ts
```diff
@@ -212,33 +205,47 @@ export async function* runDecisionLoop(
...
+  );
+
+  if (!filteredArgs) {
+    return undefined;
+  }
+
+  return {
+    toolName: toolCall.function.name,
+    parameters: filteredArgs,
+  };
```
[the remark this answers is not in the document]
MichaelMilstead: removed unused `getLLMResponseToolCallRequest`

packages/backend/src/services/decision-loop/decision-loop-service.ts
```diff
@@ -212,33 +205,47 @@ export async function* runDecisionLoop(
...
+    toolCall,
+    tools,
+    parsedArgs,
+  );
+
+  if (!filteredArgs) {
+    return undefined;
+  }
+
+  return {
```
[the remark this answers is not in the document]
MichaelMilstead: added suggested sanity check

apps/api/src/threads/util/thread-state.ts
MichaelMilstead: ah true, and `updateThreadMessageFromLegacyDecision` can still control when we update the "inner" or "outer" toolcall field. Will update

MichaelMilstead: Removed new type and added optional `isToolCallFinished` field to `LegacyComponentDecision`. 

`fixStreamedToolcalls` now allows incomplete toolcallinfo to pass through, but adds `isToolcallFinished: false`.

Then, `updateThreadMessageFromLegacyDecision` only adds tool call info to the 'outer' field once `isToolCallFinished` is true

apps/api/src/threads/util/thread-state.ts
MichaelMilstead: removed extra fields

apps/api/src/threads/util/thread-state.ts
```diff
@@ -280,62 +280,79 @@ export async function addAssistantResponse(
...
 
   for await (const chunk of stream) {
     if (currentDecision?.id && currentDecisionId !== chunk.id) {
-      // we're on to a new chunk, so if we have a previous tool call request, emit it
+      // we're on to a new message, so emit the previous one with complete tool call info
       yield {
         ...currentDecision,
-        toolCallRequest: currentToolCallRequest,
-        toolCallId: currentToolCallId,
+        toolCallRequest: previousToolCallRequest,
```
alecf: purely a naming thing (I think) - you have `currentDecision` but `previousToolCallRequest` - by name I'd think you were combining two different messages. Can we name them consistently?

apps/api/src/threads/util/thread-state.ts
```diff
@@ -280,62 +280,79 @@ export async function addAssistantResponse(
...
 
-    // now emit the next chunk
-    const { toolCallRequest, ...incompleteChunk } = chunk;
-    currentDecision = incompleteChunk;
+    // Track the tool call info from the incoming chunk
     currentDecisionId = chunk.id;
-    currentToolCallId = chunk.toolCallId;
-    currentToolCallRequest = toolCallRequest;
-    yield incompleteChunk;
+    const incomingToolCallId = chunk.toolCallId;
```
alecf: also naming, now we have `incoming...`?

apps/api/src/threads/util/thread-state.ts
```diff
@@ -280,62 +280,79 @@ export async function addAssistantResponse(
...
+    const incomingToolCallRequest = chunk.toolCallRequest;
+
+    // Store the tool call info for when we need to emit the final chunk
+    // (either when ID changes or at end of stream)
+    if (incomingToolCallRequest) {
```
alecf: I don't think you need to check this, because even if this is undefined, you want to set previousToolCallRequest to undefined (i.e. it's more complicated to check then to just do it all the time)

alecf: Honestly when you do all that simplification it goes back to the original code that was here. I think really the only change to this function that you need is to change `yield incompleteChunk` to `yield { ...chunk, isToolCallFinished: false }`

apps/api/src/threads/util/thread-state.ts
```diff
@@ -280,62 +280,79 @@ export async function addAssistantResponse(
...
+
   if (currentDecision) {
-    yield {
+    const finalChunk: LegacyComponentDecision = {
       ...currentDecision,
-      toolCallRequest: currentToolCallRequest,
-      toolCallId: currentToolCallId,
+      toolCallRequest: previousToolCallRequest,
+      toolCallId: previousToolCallId,
+      isToolCallFinished: true,
```
alecf, line 339: (and this of course)

packages/backend/src/services/decision-loop/decision-loop-service.ts
```diff
@@ -212,33 +205,52 @@ export async function* runDecisionLoop(
...
-  const originalRequest = getLLMResponseToolCallRequest(chunk);
-  // Just means the tool call is still streaming
-  if (!originalRequest) {
-    return;
-  }
+): ToolCallRequest | undefined {
   // "custom" tool calls are not supported
   if (toolCall.type !== "function") {
-    return originalRequest;
+    return undefined;
```
alecf, line 223: maybe we need a console.warn here too just in case?

packages/backend/src/services/decision-loop/decision-loop-service.ts
```diff
@@ -212,33 +205,52 @@ export async function* runDecisionLoop(
...
+      return undefined;
     }
+
+    parsedArgs = parsed;
+  } catch (_e) {
+    // If parsing fails completely, we can't build a request
+    return undefined;
   }
+
+  if (!parsedArgs) {
```
alecf: all of this feels kinda redundant with the `if (!parsed || ...)` above - perhaps just move that down here?

apps/api/src/threads/util/thread-state.ts
MichaelMilstead: Fixed!

alecf (approved):

PR #1446, opened 2025-12-02
feat(react): migrate to Standard Schema for schema conversion

lachieh:
## Summary

This PR updates the React SDK to use [Standard Schema](https://standardschema.dev/) as the primary interface for schema validation. This enables support for **any Standard Schema compliant validator** (Zod, Valibot, ArkType, etc.) instead of being tightly coupled to Zod.

### Key Changes

- **New `schema/` module** - Centralized schema utilities:
  - `schemaToJsonSchema()` - Convert any Standard Schema to JSON Schema
  - `isStandardSchema()` - Detect Standard Schema compliant validators
  - `assertNoRecordSchema()` - Validate schemas don't contain unsupported record types
  - `getZodFunctionArgs/Returns()` - Extract args/returns from Zod function schemas

- **New tool interface** - Tools can now use `inputSchema` and `outputSchema` instead of the deprecated `toolSchema`:
  ```typescript
  const tool: TamboTool = {
    name: "myTool",
    description: "...",
    tool: (input) => result,
    inputSchema: z.object({ query: z.string() }),  // Any Standard Schema
    outputSchema: z.string(),
  };
  ```

- **Zod 3 and Zod 4 support** - Both versions are supported with automatic detection

- **Removed Zod-specific utilities** - Deleted `validate-zod-schema.ts` and `schema-utils.ts` in favor of the new `schema/` module

### Migration

The deprecated `toolSchema` (Zod function schema) interface is still supported for backwards compatibility but will be removed in a future release. Users should migrate to `inputSchema`/`outputSchema`.

## Test plan

- [x] Type check passes
- [x] All 400 tests pass
- [x] Lint passes
- [x] Backwards compatibility with existing `toolSchema` interface
- [x] New `inputSchema`/`outputSchema` interface works correctly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41521,6 +41521,16 @@
         "zod": "^3.25.0 || ^4.0.0"
       }
     },
+    "node_modules/zodInternalAlias": {
+      "name": "zod",
+      "version": "4.1.13",
+      "resolved": "https://registry.npmjs.org/zod/-/zod-4.1.13.tgz",
+      "integrity": "sha512-AvvthqfqrAhNH9dnfmrfKzX5upOdjUVJYFqNSlkmGf64gRaTzlPwz99IHYnVs28qYAybvAlBV+H7pn0saFY4Ig==",
+      "license": "MIT",
+      "funding": {
+        "url": "https://github.com/sponsors/colinhacks"
+      }
+    },
     "node_modules/zwitch": {
       "version": "2.0.4",
       "resolved": "https://registry.npmjs.org/zwitch/-/zwitch-2.0.4.tgz",
@@ -41768,7 +41778,7 @@
         "ts-essentials": "^10.1.1",
         "ts-node": "^10.9.2",
         "use-debounce": "^10.0.6",
-        "zod": "^4.1.0"
+        "zodInternalAlias": "npm:zod@^4.1.0"
       },
       "devDependencies": {
         "@eslint/js": "^9.39.1",
@@ -41799,6 +41809,7 @@
         "typescript": "^5.9.3",
         "typescript-eslint": "^8.48.1",
         "valibot": "^1.2.0",
+        "zod": "^3.25.76 || ^4.1.0",
         "zod-to-json-schema": "^3.25.0"
       },
       "engines": {
@@ -41811,24 +41822,19 @@
         "@types/react-dom": "^18.0.0 || ^19.0.0",
         "react": "^18.0.0 || ^19.0.0",
         "react-dom": "^18.0.0 || ^19.0.0",
+        "zod": "^3.25.76 || ^4.1.0",
         "zod-to-json-schema": "^3.25.0"
       },
       "peerDependenciesMeta": {
         "@modelcontextprotocol/sdk": {
           "optional": true
         },
+        "zod": {
+          "optional": true
+        },
         "zod-to-json-schema": {
           "optional": true
         }
-      }
-    },
-    "react-sdk/node_modules/zod": {
-      "version": "4.1.13",
-      "resolved": "https://registry.npmjs.org/zod/-/zod-4.1.13.tgz",
-      "integrity": "sha512-AvvthqfqrAhNH9dnfmrfKzX5upOdjUVJYFqNSlkmGf64gRaTzlPwz99IHYnVs28qYAybvAlBV+H7pn0saFY4Ig==",
-      "license": "MIT",
-      "funding": {
-        "url": "https://github.com/sponsors/colinhacks"
       }
     },
     "showcase": {
--- a/react-sdk/README.md
+++ b/react-sdk/README.md
@@ -276,23 +276,23 @@ Sometimes you need functions that run in the browser. DOM manipulation, authenti
 
 ```tsx
 import { z } from "zod/v4";
-import { type TamboTool } from "@tambo-ai/react";
+import { defineTool } from "@tambo-ai/react";
 
-const tools: TamboTool[] = [
-  {
+const tools = [
+  defineTool({
     name: "getWeather",
     description: "Fetches weather data for a location",
-    tool: async (location: string) =>
+    tool: async ({ location }) =>
       fetch(`/api/weather?q=${location}`).then((r) => r.json()),
-    toolSchema: z.function({
-      input: z.string(),
-      output: z.object({
-        temperature: z.number(),
-        condition: z.string(),
-        location: z.string(),
-      }),
+    inputSchema: z.object({
+      location: z.string(),
     }),
-  },
+    outputSchema: z.object({
+      temperature: z.number(),
+      condition: z.string(),
+      location: z.string(),
+    }),
+  }),
 ];
 
 <TamboProvider tools={tools} components={components}>
--- a/react-sdk/package.json
+++ b/react-sdk/package.json
@@ -63,10 +63,14 @@
     "@types/react-dom": "^18.0.0 || ^19.0.0",
     "react": "^18.0.0 || ^19.0.0",
     "react-dom": "^18.0.0 || ^19.0.0",
+    "zod": "^3.25.76 || ^4.1.0",
     "zod-to-json-schema": "^3.25.0"
   },
   "peerDependenciesMeta": {
     "@modelcontextprotocol/sdk": {
+      "optional": true
+    },
+    "zod": {
       "optional": true
     },
     "zod-to-json-schema": {
@@ -85,7 +89,7 @@
     "ts-essentials": "^10.1.1",
     "ts-node": "^10.9.2",
     "use-debounce": "^10.0.6",
-    "zod": "^4.1.0"
+    "zodInternalAlias": "npm:zod@^4.1.0"
   },
   "devDependencies": {
     "@eslint/js": "^9.39.1",
@@ -116,6 +120,7 @@
     "typescript": "^5.9.3",
     "typescript-eslint": "^8.48.1",
     "valibot": "^1.2.0",
+    "zod": "^3.25.76 || ^4.1.0",
     "zod-to-json-schema": "^3.25.0"
   }
 }
--- a/react-sdk/src/providers/tambo-interactable-provider-partial-updates.test.tsx
+++ b/react-sdk/src/providers/tambo-interactable-provider-partial-updates.test.tsx
@@ -1,6 +1,6 @@
 import { act, renderHook } from "@testing-library/react";
 import React from "react";
-import { z } from "zod/v4";
+import { z } from "zodInternalAlias";
 import { TamboInteractableComponent } from "../model/tambo-interactable";
 import {
   TamboInteractableProvider,
--- a/react-sdk/src/providers/tambo-interactable-provider.tsx
+++ b/react-sdk/src/providers/tambo-interactable-provider.tsx
@@ -8,7 +8,7 @@ import React, {
   useEffect,
   useState,
 } from "react";
-import { z } from "zod/v4";
+import { z } from "zodInternalAlias";
 import { createInteractablesContextHelper } from "../context-helpers/current-interactables-context-helper";
 import {
   TamboInteractableComponent,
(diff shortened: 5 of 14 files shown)


react-sdk/src/util/schema.ts
```diff
@@ -0,0 +1,198 @@
...
+ */
+export function isJSONSchema(obj: unknown): obj is JSONSchema7 {
+  return (
+    typeof obj === "object" &&
+    obj !== null &&
+    !isStandardSchema(obj) &&
+    "type" in obj &&
+    typeof (obj as { type: unknown }).type === "string"
+  );
+}
```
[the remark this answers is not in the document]
lachieh: @CharlieHelps Update this to use a zod schema (`import {z} from "zod/4"`) to check that the object looks like a JSONSchema object from the top level of keys. We don't need to go deeper than that. Update the name of the function to reflect it's basic nature.

react-sdk/src/util/registry.ts
```diff
@@ -123,22 +122,50 @@ export const mapTamboToolToContextTool = (
...
+ * Extracts the args schema from a Zod function schema (v3 or v4).
+ * - Zod 4: args at `_def.input`
+ * - Zod 3: args at `_def.args`
+ * @returns The args schema, or undefined if not found
+ */
+export function getZodFunctionArgs(schema: unknown): unknown | undefined {
+  const s = schema as Record<string, unknown>;
+  const def = s._def as { input?: unknown; args?: unknown } | undefined;
+  // Zod 4 uses `input`, Zod 3 uses `args`
+  return def?.input ?? def?.args;
```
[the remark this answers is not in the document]
lachieh: @CharlieHelps import the zod type guards from schema.ts and use them inside these functions to check if they are Zod function schemas.

react-sdk/src/util/generate-component.ts
```diff
@@ -27,10 +27,13 @@ export function renderComponentIntoMessage(
     componentList,
   );
 
-  const validatedProps =
-    registeredComponent.props instanceof z.ZodType
-      ? registeredComponent.props.parse(parsedProps)
-      : parsedProps;
+  const validatedProps = isStandardSchema(registeredComponent.props)
+    ? (
+        registeredComponent.props as unknown as {
```
alecf: very minor but surprised this cast is necessary?

react-sdk/src/model/component-metadata.ts
```diff
@@ -37,26 +48,61 @@ export type TamboToolRegistry = Record<string, TamboTool>;
...
+ *    input: z.string(),
+ *    output: z.object({
+ *      temperature: z.number(),
+ *      conditions: z.string()
+ *     })
+ *   })
+ * };
+ * ```
+ */
+export interface TamboTool {
```
alecf: sad to see that there's no easy way to keep `Args` and `Returns` - I assume this just gets incredibly hard when you're dealing with a combination of zod and StandardSchema?

react-sdk/src/util/registry.ts
```diff
@@ -123,22 +122,50 @@ export const mapTamboToolToContextTool = (
...
-  schema: z.ZodFunction<any, any> | JSONSchemaLite,
+/**
+ * Extracts the args schema from a Zod function schema (v3 or v4).
+ * - Zod 4: args at `_def.input`
+ * - Zod 3: args at `_def.args`
+ * @returns The args schema, or undefined if not found
+ */
+export function getZodFunctionArgs(schema: unknown): unknown | undefined {
+  const s = schema as Record<string, unknown>;
+  const def = s._def as { input?: unknown; args?: unknown } | undefined;
```
alecf: seems like it is worth using `isZod3FunctionSchema` / `isZod4FunctionSchema` explicitly here rather than `def?.input ?? def?.args` so you don't have this dectection sprinkled around

react-sdk/src/util/schema.ts
```diff
@@ -0,0 +1,198 @@
...
+  if (!schema) {
+    return undefined;
+  }
+
+  try {
+    return schemaToJsonSchema(schema);
+  } catch (error) {
+    console.error("Error converting schema to JSON Schema:", error);
+    return undefined;
+  }
```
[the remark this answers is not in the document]
lachieh: @CharlieHelps have the safe function return `{ schema: JSONSchema, success: true } | { schema: undefined, success: false, error: string }`

react-sdk/src/schema/validate.ts
```diff
@@ -0,0 +1,196 @@
...
+      return;
+    }
+  } catch {
+    // If schema conversion fails, skip validation
+    // This can happen for function schemas or other special types
+    return;
+  }
+
+  assertNoRecordInJsonSchema(jsonSchema, [], contextName);
+}
```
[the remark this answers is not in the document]
lachieh: @CharlieHelps yes please

react-sdk/src/util/registry.ts
```diff
@@ -125,32 +130,50 @@ export const mapTamboToolToContextTool = (
...
-  const def = s._def as { input?: unknown; args?: unknown } | undefined;
+  if (!isZodSchema(schema)) {
+    return undefined;
+  }
+
+  const s = schema as { _def?: { input?: unknown; args?: unknown } };
+  const def = s._def;
   // Zod 4 uses `input`, Zod 3 uses `args`
   return def?.input ?? def?.args;
 }
```
lachieh: @CharlieHelps have this use the versioned `isZodXFunctionSchema` functions since they will also be type guards

react-sdk/src/util/registry.ts
```diff
@@ -125,32 +130,50 @@ export const mapTamboToolToContextTool = (
...
 export function isZod3FunctionSchema(schema: unknown): boolean {
-  const s = schema as Record<string, unknown>;
-  return (s._def as { typeName?: string })?.typeName === "ZodFunction";
+  if (!isZodSchema(schema)) {
+    return false;
+  }
+
+  const s = schema as { _def?: { typeName?: string } };
+  return s._def?.typeName === "ZodFunction";
 }
```
lachieh: @CharlieHelps use the versioned functions (isZod4Schema). Since they are typeguards, the `as` assertions should not be necessary. Adjust the return type to be an assertion on `schema` for these functions too.

lachieh (commented):
@CharlieHelps make these changes too

react-sdk/src/model/component-metadata.ts
lachieh: It's definitely possible, and I think we should extend their usage to the functions that take them as args so that we get type errors inline when registering tools/components/interactables (unless I'm missing something and that already works?)

alecf: well what already worked was you could assign the tool, and then pass in schema, and the type system would complain if things didn't line up:
```
{
  tool: (message: string) => { console.log("I got a message: ", message); return "Hello!" },
  toolSchema: z.function().args(z.number()).returns(z.number())
 }
 ```
and you'd get a type error because the args / return values do not match

Comment (lachieh) [reply react-sdk/src/model/component-metadata.ts]: I see, yeah I wasn't getting those because it wasn't specifying the type args. I'm going to try something to fix this.

File: react-sdk/src/util/generate-component.ts (hunk 3 above)
Comment (lachieh) [reply react-sdk/src/util/generate-component.ts]: Nope! Removed:

```ts
  const validatedProps = isStandardSchema(registeredComponent.props)
    ? registeredComponent.props["~standard"].validate(parsedProps)
    : parsedProps;
```

react-sdk/src/model/component-metadata.ts
lachieh: Ok, this is where some chunky changes happen. See the changes in `react-sdk/src/model/component-metadata.ts` and the new `RegisterToolFn` interface in `react-sdk/src/providers/tambo-registry-provider.tsx`. This should add some stronger type checking! With that, there are some opinions though. We should probably give it some use to see how useful or aggressive it is.

react-sdk/src/util/generate-component.ts
```diff
@@ -27,10 +27,13 @@ export function renderComponentIntoMessage(
...
+          parse: (data: unknown) => unknown;
+        }
+      ).parse(parsedProps)
+    : parsedProps;
```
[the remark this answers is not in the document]
lachieh: This is fixed

react-sdk/src/model/component-metadata.ts
```diff
@@ -29,39 +52,225 @@ export interface RegisteredComponent extends TamboAI.AvailableComponent {
...
-  tool: (...args: z.infer<Args>) => z.infer<Returns>;
-  toolSchema: z.ZodFunction<Args, Returns> | JSONSchemaLite;
+  /**
+   * Optional human-readable name of the tool for display purposes.
+   */
+  title?: string;
+  /**
+   * Optional properties describing tool behavior
+   */
+  annotations?: {
```
alecf, line 137: where does this come from? (some of the docs are referring to "this resource" but I'm not 100% sure what that refers to - like what does it mean when the audience for a tool is "user" ?

alecf (approved):

react-sdk/src/model/component-metadata.ts
lachieh: Yeah this comes from the MCP spec on resources (specifically here: [5.1 - Annotations](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#annotations)).

> Clients can use these annotations to:
> - Filter resources based on their intended audience
> - Prioritize which resources to include in context
> - Display modification times or sort by recency

It's not critical to include this, and perhaps given our conversation this morning on context size, perhaps we don't.

alecf: but also these are tools, not resources

alecf: oh I think I see the confusion: tool *responses* themselves can be resources instead of plain text, and *those* resources can have annotations

alecf: https://modelcontextprotocol.io/specification/2025-06-18/server/tools#tool-result

lachieh: Sorry, clarification: I got the fields for tool from [here](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#tool). It's the same field name for "Tool", but I was looking at "Tool Result" which mentions that "content types support optional annotations" and that is where I got the spec for annotations.

That said, it seems that the "Tool" section doesn't say if it specifically matches same structure as the _resource_'s annotation field. How do you interpret that wording? I think either way, I'll add links in the jsdoc comments to the spec.

lachieh: ok, so I should've just gone here to begin with:
- https://modelcontextprotocol.io/specification/2025-06-18/schema#tool
- https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations

lachieh: I'm going to remove annotations for now since it isn't in line with the MCP Spec and we can continue the discussion later.

alecf: yeah - I do actually think *those* annotations are useful, but yeah we can punt that to discussion / new PR

apps/web/lib/tambo/tools/types.ts
```diff
@@ -10,6 +9,8 @@ export interface ToolContext {
 }
 
 /**
- * Tool registration function
+ * Tool registration function.
+ * Uses a permissive signature to be compatible with the SDK's RegisterToolFn
+ * without importing the complex generic types that cause type explosion.
  */
-export type RegisterToolFn = (tool: TamboTool) => void;
+export type RegisterToolFn = (tool: any) => void;
```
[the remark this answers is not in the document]
lachieh: Resolved with `export type { RegisterToolFn } from @tambo-ai/react`

react-sdk/src/util/registry-validators.ts
```diff
@@ -1,21 +1,83 @@
...
+  } else if ("toolSchema" in tool && tool.toolSchema) {
+    // For deprecated toolSchema, only validate if it's a Zod function schema
+    // (JSON Schema toolSchemas don't need validation)
+    if (isZodFunctionSchema(tool.toolSchema)) {
+      const args = getZodFunctionArgs(tool.toolSchema);
+      if (args) {
+        assertNoRecordSchema(args, `toolSchema of tool "${tool.name}"`);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
lachieh: added check for toolSchema w/ jsonschema

PR #1495, opened 2025-12-09
feat: send interactables internal state to tambo

MichaelMilstead:
Previously tambo was able to "see" only the last props sent to an interactable component, and had no way to know the internal state values of the component. If a user interacted with an interactable component to update some internal state value, tambo would not know about the update, and would not be able to act based on the complete "current state."

This PR:
- updates TamboInteractable type to include `state` for storing the current state values of an interactable.
- updates `TamboInteractableProvider` with methods to get and set the state values of an interactable
- updates `interactable context helper` to send the current state of all interactables to tambo on each sent message
- updates `useTamboComponentState` hook to allow usage in an interactable component

In an interactable component, you can now replace usage of `useState` with `useTamboComponentState` and tambo will see those values.

### DIFF
--- a/apps/web/CHANGELOG.md
+++ b/apps/web/CHANGELOG.md
@@ -2,38 +2,44 @@
 
 ## [0.124.0](https://github.com/tambo-ai/tambo/compare/web-v0.123.4...web-v0.124.0) (2025-12-08)
 
+
 ### Features
 
-- **cli:** bring wysiwyg editor into main message-input component ([#1415](https://github.com/tambo-ai/tambo/issues/1415)) ([6d0a89d](https://github.com/tambo-ai/tambo/commit/6d0a89dfa75c953279b56771209c74c4b3bcc58d))
-- **mcp-resources:** Enable @-resource and /-command inline completion ([#1464](https://github.com/tambo-ai/tambo/issues/1464)) ([775ca87](https://github.com/tambo-ai/tambo/commit/775ca8789341de492bd084e1fbede76ffd3d1f8c))
-- **web:** register dashboard components with Tambo ([#1467](https://github.com/tambo-ai/tambo/issues/1467)) ([0cd2f6e](https://github.com/tambo-ai/tambo/commit/0cd2f6eec4587385da8c9ecb725d25d79ead4c21))
+* **cli:** bring wysiwyg editor into main message-input component ([#1415](https://github.com/tambo-ai/tambo/issues/1415)) ([6d0a89d](https://github.com/tambo-ai/tambo/commit/6d0a89dfa75c953279b56771209c74c4b3bcc58d))
+* **mcp-resources:** Enable @-resource and /-command inline completion ([#1464](https://github.com/tambo-ai/tambo/issues/1464)) ([775ca87](https://github.com/tambo-ai/tambo/commit/775ca8789341de492bd084e1fbede76ffd3d1f8c))
+* **web:** register dashboard components with Tambo ([#1467](https://github.com/tambo-ai/tambo/issues/1467)) ([0cd2f6e](https://github.com/tambo-ai/tambo/commit/0cd2f6eec4587385da8c9ecb725d25d79ead4c21))
+
 
 ### Bug Fixes
 
-- **deps:** upgrade to zod v3 subpath imports and MCP SDK 1.24 ([#1465](https://github.com/tambo-ai/tambo/issues/1465)) ([c8b7f07](https://github.com/tambo-ai/tambo/commit/c8b7f079560d423082c005018a103b9eb3cf6993))
-- **web:** clarify OpenAI-compatible base URL hint ([#1455](https://github.com/tambo-ai/tambo/issues/1455)) ([1da60e2](https://github.com/tambo-ai/tambo/commit/1da60e24887c9dfa3eb6dffc0156d11996a61b7d))
+* **deps:** upgrade to zod v3 subpath imports and MCP SDK 1.24 ([#1465](https://github.com/tambo-ai/tambo/issues/1465)) ([c8b7f07](https://github.com/tambo-ai/tambo/commit/c8b7f079560d423082c005018a103b9eb3cf6993))
+* **web:** clarify OpenAI-compatible base URL hint ([#1455](https://github.com/tambo-ai/tambo/issues/1455)) ([1da60e2](https://github.com/tambo-ai/tambo/commit/1da60e24887c9dfa3eb6dffc0156d11996a61b7d))
+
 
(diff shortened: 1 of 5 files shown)


react-sdk/src/hooks/use-component-state.test.tsx
```diff
@@ -56,10 +71,43 @@ describe("useTamboComponentState", () => {
...
+        }
+        return mockMessage;
+      }
+      if (context === InteractableIdContext) {
+        return mockComponentId;
+      }
+      // For other contexts, use the original implementation
+      return originalUseContext(context);
+    });
+
```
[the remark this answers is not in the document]
MichaelMilstead: @CharlieHelps yes please

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -45,19 +47,19 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
   const { registerTool } = useTamboComponent();
   const { addContextHelper, removeContextHelper } = useTamboContextHelpers();
 
-  // Create a stable context helper function
-  const contextHelper = useCallback(() => {
-    return createInteractablesContextHelper(() => interactableComponents)();
+  // Create a stable context helper function for interactable components
+  const interactablesContextHelper = useCallback(() => {
+    return createInteractablesContextHelper(interactableComponents)();
```
alecf: Minor but seems like there's a simplification here where `createInteractablesContextHelper()` doesn't need the double-call - e.g. since the only place it's used is in this function, why not just make `createInteractablesContextHelper` return the proper result to allow:

```suggestion
    return createInteractablesContextHelper(interactableComponents);
```

react-sdk/src/context-helpers/current-interactables-context-helper.ts
```diff
@@ -27,23 +27,21 @@ export const currentInteractablesContextHelper: ContextHelperFn = () => {
...
- * @param getComponents Function to get current interactable components
+ * @param components Array of interactable components
  * @returns Context helper function
  */
 export const createInteractablesContextHelper = (
-  getComponents: () => any[],
+  components: any[],
 ): ContextHelperFn => {
   return () => {
     try {
```
alecf: with this simplification I would leave out the try/catch, if anything it may mask other issues - we should be handling this in the interactables registration, not having specialized code here

react-sdk/src/hooks/use-component-state.tsx
```diff
@@ -81,21 +109,69 @@ export function useTamboComponentState<S>(
...
 
-  // Mirror the thread message's componentState value to the local state
+  // Set initial value in interactable state if we're in an interactable context and there's no existing state
   useEffect(() => {
-    const messageState = message?.componentState?.[keyName];
+    if (!componentId) {
+      return;
+    }
+
+    if (messageState !== undefined) {
```
alecf: is there a possibility of a race condition if messageState hasn't been initialized when this useEffect first fires?

(I think either way add some comments here - it might also be worth adding a separate boolean outside the useEffect so that updating individual values don't cause the useEffect to fire too much, since that might trigger a lot of duplicate state updates)

```ts
const isUnintializedInteractable = !!componentId && const existingInteractableState = isUninitializedInteractable ? 
      getInteractableComponentState(componentId)?.[keyName] : undefined;
const shouldUpdate = existingInteractableState === undefined && initialValue !== undefined;

useEffect(() => {
  if (!shouldUpdate) {
    return;
  }
  setInteractableState(componentId, keyName, initialValue);
}, 
// note how we no longer depend on messageState or getInteractableComponentState (which depends on the whole list of interactable components)
[shouldUpdate, componentId, keyName, initialValue, setInteractableState]);
```

react-sdk/src/hooks/use-component-state.tsx
```diff
@@ -81,21 +109,69 @@ export function useTamboComponentState<S>(
...
+    componentId,
+    keyName,
+    initialValue,
+    messageState,
+    getInteractableComponentState,
+    setInteractableState,
+  ]);
+
+  // Mirror the thread message's componentState value to the local state and interactable state
+  useEffect(() => {
```
alecf, line 157: possibly same strategy here?

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -297,6 +300,42 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
     setInteractableComponents([]);
   }, []);
 
+  const setInteractableStateValue = useCallback(
+    (componentId: string, key: string, value: unknown) => {
+      setInteractableComponents((prev) => {
+        const component = prev.find((c) => c.id === componentId);
+        if (!component) {
+          return prev;
```
alecf, line 319: worth a console.warn that we're updating an interactable that is no (longer? yet?) registered?

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -297,6 +300,42 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
...
+          ...component,
+          state: {
+            ...(component.state ?? {}),
+            [key]: value,
+          },
+        };
+
+        const updatedComponents = [...prev];
+        const idx = prev.findIndex((c) => c.id === componentId);
+        updatedComponents[idx] = updated;
```
alecf: minor pattern improvement (IMO)


```suggestion
        const updatedComponents = prev.map(c => {
          if (c.id === componentId) {
            return updated;
          }
          return c;
        }
```

apps/api/CHANGELOG.md
```diff
@@ -2,83 +2,69 @@
 
 ## [0.126.0](https://github.com/tambo-ai/tambo/compare/api-v0.125.0...api-v0.126.0) (2025-12-08)
 
-
 ### Features
 
-* **backend:** use ThreadMessage directly in AI SDK client (PR 2/3) ([#1424](https://github.com/tambo-ai/tambo/issues/1424)) ([5199869](https://github.com/tambo-ai/tambo/commit/5199869c667c5e3e24b6c47bb8b91d4ed2ba5b80))
-
+- **backend:** use ThreadMessage directly in AI SDK client (PR 2/3) ([#1424](https://github.com/tambo-ai/tambo/issues/1424)) ([5199869](https://github.com/tambo-ai/tambo/commit/5199869c667c5e3e24b6c47bb8b91d4ed2ba5b80))
```
alecf: you still have changelog reformatting in this PR

PR #1502, opened 2025-12-10
fix(editor): Clean up message-input and text-editor to only expose a limited set of capabilities in the editor

alecf:
- **refactor(cli): expose minimal TamboEditor interface instead of TipTap Editor**
- **refactor(web): expose minimal TamboEditor interface instead of TipTap Editor**


akhileshrangani4 (approved):
lgtm

PR #1508, opened 2025-12-11
feat: add EditWithTamboButton page to showcase site

akhileshrangani4:
- Updated EditWithTamboButton to accept custom suggestions and add context attachments when editing components.
- Modified MessageThreadCollapsible to utilize custom suggestions and clear them upon sending a new message.
- Introduced a new showcase page for EditWithTamboButton with usage examples and installation instructions.
- Added a new ContactForm component to demonstrate the EditWithTamboButton in action.
- Implemented a reusable Select component for dropdown functionality in forms.

### DIFF
--- a/apps/web/components/ui/tambo/edit-with-tambo-button.tsx
+++ b/apps/web/components/ui/tambo/edit-with-tambo-button.tsx
@@ -176,15 +176,11 @@ export function EditWithTamboButton({
       // Check if mention already exists to avoid duplicates
       if (editor.hasMention(interactableId)) {
         // If mention exists, just append the user query
-        editor.focus("end");
-        const currentText = editor.getTextWithResourceURIs().text;
-        editor.setContent(currentText + " " + messageToInsert);
+        editor.appendText(" " + messageToInsert);
       } else {
         // Insert @mention (which adds a space after), then append the user query
-        editor.focus();
         editor.insertMention(interactableId, componentName);
-        const currentText = editor.getTextWithResourceURIs().text;
-        editor.setContent(currentText + messageToInsert);
+        editor.appendText(messageToInsert);
       }
     }
   }, [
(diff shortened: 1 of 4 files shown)


apps/web/components/ui/tambo/edit-with-tambo-button.tsx
```diff
@@ -153,26 +174,22 @@ export function EditWithTamboButton({
     const editor = editorRef.current;
     if (editor) {
       // Check if mention already exists to avoid duplicates
-      if (editor.hasMention(component.componentName ?? "Unknown Component")) {
+      if (editor.hasMention(componentName)) {
```
alecf: minor but I'm realizing "hasMention" should really be using componentId here, not the visual name (the component should be updated too) - you might have two mentions with the same name but different ids (might have to tweak the text-editor here)

apps/web/components/ui/tambo/edit-with-tambo-button.tsx
```diff
@@ -153,26 +174,22 @@ export function EditWithTamboButton({
...
         // If mention exists, just append the user query
         editor.focus("end");
         const currentText = editor.getTextWithResourceURIs().text;
         editor.setContent(currentText + " " + messageToInsert);
```
alecf: hmm thinking what we need rather than explicit focus("end") + setcontent() is an `appendText` - the setting of the currentText causes some odd gymnastics in the editor where it kind of re-maps the content into mention nodes/etc. 

I wonder if we added that, we might be able to remove focus("end") and maybe setContent altogether

apps/web/components/ui/tambo/edit-with-tambo-button.tsx
akhileshrangani4: oh yes, you are right, I have updated the text editor and this component.

apps/web/components/ui/tambo/edit-with-tambo-button.tsx
akhileshrangani4: ohh yes, that would be very neat. updated that [here](https://github.com/tambo-ai/tambo/pull/1508/commits/11fcfcb739883d557e235914a31abbaff8036240), now we just have this one line:
```ts
editor.appendText(" " + messageToInsert);
```

alecf (approved):

PR #1513, opened 2025-12-12
fix(react): correct schema parsing logic

lachieh:
- **fix(react-sdk): correct check order for zod schemas**
- **fix(web/showcase): fix warning from optional peer deps**

Essentially, the `looksLikeJSONSchema` was matching too eagerly and was preventing any schema from being converted to a plain JSONSchema object. The valibot schemas are not working, but there isn't enough tests for this either so we'll have to revisit that later.

### DIFF
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -137,6 +137,13 @@ const config = {
       use: ["@svgr/webpack"],
     });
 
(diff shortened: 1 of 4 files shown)


lachieh:
> will consumers of our SDK still see those errors about `effect` and `sury`?

Only if they are using Next. I might try some things with dynamic imports and see if that solves the same issue.

PR #1528, opened 2025-12-13
fix(cli): Fixed cli logs typos on self-hosted flow

andrewckor:
It fixes the outdated logs of the self-hosted instructions, which are pointing to the wrong folder.

### DIFF
--- a/cli/src/commands/init.ts
+++ b/cli/src/commands/init.ts
@@ -165,8 +165,12 @@ function displaySelfHostInstructions(): void {
   console.log(
     chalk.gray("  1. Create .env files (see repo .env.example files)"),
   );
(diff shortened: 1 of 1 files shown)


michaelmagan:
Thanks @andrewckor we have a few issued in our backlog to clean up self-hosted setup. Thanks for finding this and fixing it!

lachieh (approved):

andrewckor:
@lachieh, it cannot merge due to the failed Vercel deployment.

lachieh:
Thanks for the ping, @andrewckor. Looks like it was just some formatting changes so I pushed that and now we're good. Thanks again for the contribution!

PR #1531, opened 2025-12-15
feat: update interactables context with selection tracking

akhileshrangani4:
### Summary

This PR refactors how context attachments communicate with the AI by introducing a unified `isSelectedForInteraction` field on interactable components, replacing the previous approach of creating separate context helpers for each attachment.

### What Changed

**Context Attachment Flow**
- Context attachments no longer create individual context helpers
- Instead, attached components are marked with `isSelectedForInteraction: true` in the unified interactables context
- Added `getCurrentAttachments()` function for synchronous access to attachments (avoids stale closure issues in callbacks)
- Attachments now include `metadata.componentId` for precise component matching

**Context Helper Refactoring**
- `createInteractablesContextHelper` now accepts getter functions (`getComponents`, `getSelectedIds`) instead of direct arrays
- Components include new `isSelectedForInteraction` boolean field indicating explicit user selection
- Uses ref-based access pattern to always get latest component/attachment state

**EditWithTamboButton**
- Now uses context attachments with `componentId` metadata instead of `additionalContext.inlineEdit`
- Clears attachments after sending (one-time edit behavior)
- Added optional `contextKey` prop for thread caching

**Provider Changes**
- Moved `contextKey` generation from `MessageThreadPanel` to `MessageThreadPanelProvider` for shared access
- `TamboInteractableProvider` now reads from `useTamboContextAttachment` to determine selected components

**AI Instructions**
- Added explicit guidance about `isSelectedForInteraction` in decision loop prompts
- AI is instructed to focus on components where `isSelectedForInteraction: true`

### Breaking Changes

- `getContextHelperData` prop on `TamboContextAttachmentProvider`/`TamboProvider` is deprecated (no-op)
- Context attachment structure now requires `metadata.componentId` for precise component matching

### Why

The previous approach created separate context helpers per attachment, which:
1. Could lead to stale closure issues
2. Required complex synchronization between attachment state and context helpers
3. Made it harder for the AI to understand which components were explicitly selected

The new unified approach:
1. Uses refs for always-fresh data in callbacks
2. Centralizes selection state in the interactables context
3. Gives the AI clearer signal about user intent via `isSelectedForInteraction`


apps/web/components/ui/tambo/edit-with-tambo-button.tsx
```diff
@@ -133,20 +140,37 @@ export function EditWithTamboButton({
...
+      await sendThreadMessage(prompt.trim(), {
+        streamResponse: true,
+        contextKey,
+      });
+
+      // Clear the prompt after successful send
+      setPrompt("");
+    } finally {
+      // Clear attachments regardless of success/failure (one-time edit)
+      clearContextAttachments();
```
alecf, line 161: so we don't have to worry about losing stuff like the current url or time? (I remember we had a thing where you registered those separately?)

apps/web/components/ui/tambo/edit-with-tambo-button.tsx
```diff
@@ -161,12 +185,15 @@ export function EditWithTamboButton({
...
-      name: componentName,
-    });
+    // Add the component as a context attachment (only if valid)
+    const interactableId = component?.interactableId;
+    const componentName = component?.componentName;
+    if (interactableId && componentName) {
+      addContextAttachment({
+        name: componentName,
+        metadata: { componentId: interactableId },
+      });
```
alecf, line 195: Is this type defined somewhere, so that we could put `satisfies InteractableComponentContext` or something here and elsewhere?

apps/web/providers/message-thread-panel-provider.tsx
```diff
@@ -21,6 +23,25 @@ export function MessageThreadPanelProvider({
   const [isOpen, setIsOpen] = React.useState(false);
   const editorRef = React.useRef<TamboEditor | null>(null);
 
+  // Generate or retrieve contextKey only once on mount
+  const [contextKey] = React.useState<string>(() => {
+    // Defensive UUID generation for environments where crypto.randomUUID may not be available
+    const generateUUID = () =>
```
alecf: I don't think we should bother with fallback behavior here - just assume crypto.randomUUID() exists. (the only time it doesn't is in dev when you specify an IP address, but it does work fine when you use localhost)

cli/src/registry/edit-with-tambo-button/edit-with-tambo-button.tsx
```diff
@@ -57,6 +57,11 @@ export interface EditWithTamboButtonProps {
    * MessageThreadCollapsible or implements similar clearing logic.
    */
   suggestions?: Suggestion[];
+  /**
+   * Optional context key for thread caching. When provided, threads created via
+   * inline edit will be cached and appear in thread lists for this context key.
+   */
+  contextKey?: string;
```
alecf: it feels strange that we do this here? is this just kind of necessary because much of the rest of the sendMessage/etc does this?  I think we should be hiding contextKey more, not exposing it here..

I had actually pulled contextKey out of a few places because its better if that is managed at the `<TamboProvider>` level (and because the auth stuff under the hood actually swaps out contextKey for oauth context)

apps/web/components/ui/tambo/edit-with-tambo-button.tsx
akhileshrangani4: Nope, we don’t need to worry about losing stuff like the URL or time.

There are **two completely separate systems** here:

* **Context Helpers (`TamboContextHelpersProvider`)**
  Default context helpers (`currentPageContextHelper`, `currentTimeContextHelper`) stay registered. Before every message is sent, `getAdditionalContext()` calls them fresh.
  They are **not affected at all** by `clearContextAttachments()`.

* **Context Attachments (`TamboContextAttachmentProvider`)**
  These are the `@mentions`/component selections. `clearContextAttachments()` **only** clears these temporary user selections.

So calling `clearContextAttachments()` here **only removes** the component selection that was added via `addContextAttachment()` on lines 147–150.

Any other context helpers continues to be there.

apps/web/components/ui/tambo/edit-with-tambo-button.tsx
akhileshrangani4: There is not a defined type for this, but yes, I should add it.

akhileshrangani4: added it [here](https://github.com/tambo-ai/tambo/pull/1531/commits/221f82620db66ab97d4ca7b37d2efa8b700b6cb6)

apps/web/providers/message-thread-panel-provider.tsx
akhileshrangani4: okay, will keep this in mind. removed [it](https://github.com/tambo-ai/tambo/pull/1531/commits/1ce66368a8064474e9d4e91a97914bf7e9164e51).

cli/src/registry/edit-with-tambo-button/edit-with-tambo-button.tsx
akhileshrangani4: the reason why I added this was because in the web app, before "not" using contextkey, the thread that was generated when using the inline editor was with a different context key then what the MessageThreadPanel had, so when the user opens the thread after sending an inline prompt, the next message they send would fail saying "Thread not found". So to sync the context keys, I added that in the web app version.

But I do agree that the web app use case is really specific and the registry component user would not necessarily face the same thing, but I do think if they wanna specify a specific context key, then maybe this should be there? should I keep it or remove it?

akhileshrangani4: Though in the web app, we are not passing the context key at the `TamboProvider` level, so I think I should just do that.

akhileshrangani4: removed contextkey from the cli and showcase components.

apps/web/components/ui/tambo/edit-with-tambo-button.tsx
```diff
@@ -133,20 +141,39 @@ export function EditWithTamboButton({
...
+          componentId: interactableId,
+        } satisfies InteractableComponentMetadata,
+      });
+    }
 
-    // Clear the prompt after successful send
-    setPrompt("");
-  }, [prompt, isGenerating, component, sendThreadMessage]);
+    try {
+      await sendThreadMessage(prompt.trim(), {
```
alecf: I hope that if contextKey is left out here, it will pick it up from the context, so it shouldn't need to be passed?

akhileshrangani4: okay updated it to use `submit` from `useTamboThreadInput` and now passing context key at `TamboProvider` level - so we do not need contextkey anymore at this level - [here](https://github.com/tambo-ai/tambo/pull/1531/commits/bbc5ce6e055786f263a244b639f204e46b39cf20)

apps/web/providers/tambo-provider.tsx
```diff
@@ -3,12 +3,26 @@
...
 import { tamboRegisteredComponents } from "@/lib/tambo/config";
 import { TamboProvider, currentPageContextHelper } from "@tambo-ai/react";
+import { useState } from "react";
 
 type TamboProviderWrapperProps = Readonly<{
   children: React.ReactNode;
 }>;
 
 export function TamboProviderWrapper({ children }: TamboProviderWrapperProps) {
+  // Generate or retrieve contextKey only once on mount
```
alecf: ooooh I see why you wanted this. so for our own app, you should be able to use the user's own id as the contextKey, which you can get from `useSession()` - `session.user.id` (see authed-layout-wrapper.tsx for an example)

So this is hopefully as simple as
```ts
const { session } = useSession();
const contextKey = session?.user?.id;
```
(since this is only ever used in an authed context, `session.user.id` will never be nullish)

alecf (approved):
looks good! one other bit with contextKey, approved with that

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -47,21 +48,54 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
   >([]);
   const { registerTool } = useTamboComponent();
   const { addContextHelper, removeContextHelper } = useTamboContextHelpers();
+  const { getCurrentAttachments } = useTamboContextAttachment();
+
+  // Ref for latest value in callbacks (avoids stale closures)
+  const interactableComponentsRef = useRef<TamboInteractableComponent[]>([]);
```
MichaelMilstead, line 54: we might not need this ref

MichaelMilstead (changes requested):
looks mostly good, will add a couple more comments tomorrow!

apps/web/providers/tambo-provider.tsx
akhileshrangani4: actually, it can be nullish when the user has not yet signed in- as the tambo chat is always available even on the landing page and all the pages where session is not required.

alecf: ah... ok I guess for that a fallback to a UUID is fine but use the user id if available!

akhileshrangani4: yes sir! done [here](https://github.com/tambo-ai/tambo/pull/1531/commits/62c7f239e7ba8ce696ea71fcf8a3a5c411e89c81)

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -47,21 +48,54 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
...
+  interactableComponentsRef.current = interactableComponents;
+
+  // Context helper reads from refs to always get latest data
+  const contextHelper = useCallback(() => {
+    const currentAttachments = getCurrentAttachments();
```
MichaelMilstead, line 59: I think we should update the structure of things a little here.

Rather than the `EditWithTamboButton` updating the list of `ContextAttachments`, and then this provider determining `isSelected` based on the attachments, we should use `ContextAttachments` just to determine the list of badges to show in the UI, and have a function on this provider that the `EditWithTamboButton` can call like `setComponentSelected(componentId)` that just updates the local state of interactables.

Then we could update `EditWithTamboButton` so it separately calls `setComponentSelected` to update the state and then `addContextAttachment` to update the list of badges, but really what we should probably do is let the `setComponentSelected` handle adding the context attachment. I think it makes more sense to let the place that updates the "staged context" also decide whether or not to add a context attachment badge.

react-sdk/src/providers/tambo-context-attachment-provider.tsx
```diff
@@ -40,12 +38,29 @@ export interface ContextAttachment {
  */
 export type ContextHelperData = Record<string, unknown>;
 
+/**
+ * Metadata shape for interactable component context attachments.
+ * Used when attaching an interactable component via `@mention` or EditWithTamboButton.
```
MichaelMilstead, line 43: I think ContextAttachments shouldn't know anything about components or interactables, it should just be for storing and managing a list of "badge data" to show in the UI, regardless of what the badge represents

apps/web/components/ui/tambo/edit-with-tambo-button.tsx
```diff
@@ -75,10 +77,14 @@ export function EditWithTamboButton({
...
+  const { isIdle } = useTambo();
   const { setIsOpen: setThreadPanelOpen, editorRef } = useMessageThreadPanel();
-  const { addContextAttachment, setCustomSuggestions } =
-    useTamboContextAttachment();
+  const {
+    addContextAttachment,
+    clearContextAttachments,
+    setCustomSuggestions,
+  } = useTamboContextAttachment();
+  const { setValue: setThreadInputValue, submit } = useTamboThreadInput();
```
MichaelMilstead, line 87: for a future PR, but I think this component shouldn't have any message sending logic, and should re-use our message-input component (or as much of it as possible)

PR #1536, opened 2025-12-15
fix(react-sdk): inline reused Zod schemas in JSON Schema conversion

### LINKED ISSUE #1537: Bug: zod-to-json reuses Zod objects and emits wrong $ref (e.g. #/properties/data)
### Problem
In the hotel demo (and other apps), the zod-to-JSONSchema conversion appears to re-use Zod schema instances it has already seen within the current root schema.

When the same `ZodObject` instance is used in multiple places, the JSONSchema generated for the later property can incorrectly reference a sibling property path (e.g. `#/properties/data`) instead of inlining the schema (or using a shared definition).

### Repro
```ts
const dataschema = z.object({ name: z.string(), value: z.number() });
const props = z.object({
  data: z.array(dataschema),
  historicalData: z.array(dataschema),
});
```

### Actual
`historicalData` ends up with a `$ref` like `#/properties/data` (pointing at the other property) rather than its own schema.

### Expected
`historicalData` should either:
- inline the schema identical to `data`, or
- reference a shared definition (e.g. under `$defs`) in a way that is not tied to a sibling property path.

### Notes
Alec mentioned he has a potential fix coming up for the SDK; logging this to track + ensure we cover this case.

alecf:
Fixes #1537 
Fixes TAM-860

## Summary

- Fixes an issue where reused Zod schemas produced `$ref` references in JSON Schema output instead of inlining the schema
- For example, when defining `const dataSchema = z.object({...})` and using it in `z.object({ data: z.array(dataSchema), historicalData: z.array(dataSchema) })`, the output would contain cross-references that many consumers don't support

## Changes

- Add `{ reused: "inline" }` option for Zod 4 `toJSONSchema` in `react-sdk/src/schema/zod.ts`
- Add `{ $refStrategy: "none" }` option for Zod 3 `zodToJsonSchema` in `react-sdk/src/schema/zod.ts`
- Fix `packages/backend/src/prompt/component-formatting.ts` to also use `$refStrategy: "none"`
- Add comprehensive test coverage in `react-sdk/src/schema/zod.test.ts` (19 tests)

## Test plan

- [x] All new tests pass (`npm run test -- src/schema/zod.test.ts`)
- [x] All existing schema tests pass (`npm run test -- src/schema/`)
- [x] Type checks pass (`npm run check-types`)
- [x] Lint passes

### DIFF
--- a/react-sdk/src/schema/zod.test.ts
+++ b/react-sdk/src/schema/zod.test.ts
@@ -68,13 +68,15 @@ describe("zod schema utilities", () => {
     it("returns true for Zod 3 schemas", () => {
       const schema = z3.object({ name: z3.string() });
       expect(isZod3Schema(schema)).toBe(true);
+      expect(isZod4Schema(schema)).toBe(false);
     });
 
     it("returns true for Zod 4 schemas (they also have _def)", () => {
       // Note: Zod 4 schemas have both _def and _zod, so isZod3Schema returns true.
       // This is fine because handleZodSchemaToJson checks isZod4Schema first.
       const schema = z4.object({ name: z4.string() });
-      expect(isZod3Schema(schema)).toBe(true);
+      expect(isZod4Schema(schema)).toBe(true);
+      expect(isZod3Schema(schema)).toBe(false);
     });
 
     it("returns false for non-Zod values", () => {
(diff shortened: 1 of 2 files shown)


react-sdk/src/schema/zod.test.ts
```diff
@@ -0,0 +1,389 @@
...
+          }),
+        });
+
+        const result = handleZodSchemaToJson(shapeSchema);
+        const resultStr = JSON.stringify(result);
+
+        // The key assertion: no $ref anywhere
+        expect(resultStr).not.toContain("$ref");
+      });
+    });
```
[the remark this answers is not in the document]
alecf: @CharlieHelps yes please

react-sdk/src/schema/zod.test.ts
```diff
@@ -0,0 +1,474 @@
...
+      const schema = z3.object({ name: z3.string() });
+      expect(isZod3Schema(schema)).toBe(true);
+    });
+
+    it("returns true for Zod 4 schemas (they also have _def)", () => {
+      // Note: Zod 4 schemas have both _def and _zod, so isZod3Schema returns true.
+      // This is fine because handleZodSchemaToJson checks isZod4Schema first.
+      const schema = z4.object({ name: z4.string() });
+      expect(isZod3Schema(schema)).toBe(true);
+    });
```
lachieh, line 80: This probably doesn't have to match true for Zod 4 schemas. The thing to check will be if imports from `zod/v4@^3` and `zod/v4@^3` actually return the same properties. Changing `isZod3Schema` so that it ensures that the schema does include `._def` and _does not_ include `.def` should work. That property should only exist on `/v4` imports.

https://github.com/tambo-ai/tambo/blob/5864c5d20bb5d3675bc3a55dba95d124fa36d128/react-sdk/src/schema/zod.ts#L131-L133

react-sdk/src/schema/zod.test.ts
```diff
@@ -0,0 +1,474 @@
...
+          boundingBox: z4.object({
+            topLeft: pointSchema,
+            bottomRight: pointSchema,
+          }),
+        });
+
+        const result = handleZodSchemaToJson(shapeSchema);
+
+        expect(hasKeyDeep(result, "$ref")).toBe(false);
+      });
```
lachieh, line 432: This is the only test that doesn't include `expect(result).toMatchObject({ /*...*/ })` assuming because it is a large object. Could use `expect(result).toMatchSnapshot()` if you think it is worth including.

alecf: added an `toMatchInlineSnapshot()`

react-sdk/src/schema/zod.test.ts
alecf: done!

PR #1586, opened 2025-12-17
chore: add cspell settings and correct a few spleling mistaeks

lachieh:
I've found the cSpell project helpful. I'd like to get to a point where docs and such are checked for spelling, but this is an opt-in change.

Charlie mentioned that the settings don't explicitly look for the cspell config file, but that is fine since the tool looks for itself.


lachieh:
@MichaelMilstead yeah, there's a few too many in there to enforce it for everything but we can probably start on docs and blogs in the next pr and then expand from there as necessary.

PR #1588, opened 2025-12-18
fix: refactor ContextAttachmentProvider and allow Interactable selection

MichaelMilstead:
Updates `TamboContextAttachmentsProvider` to simplify the way it adds context to the next user message using a context helper, and removes logic for suggestions.

 Previously, the `EditWithTamboButton` depended on using the attachments provider to add context telling tambo which interactables were selected, but the context describing Interactable Components' state should not be separated into context attachments, so this PR also includes updating the `TamboInteractableProvider` with a function to handle setting a new `isSelectedForInteraction` field on an interactable, and updates the `EditWithTamboButton` to use this.

### DIFF
--- a/apps/web/components/ui/tambo/edit-with-tambo-button.tsx
+++ b/apps/web/components/ui/tambo/edit-with-tambo-button.tsx
@@ -58,11 +58,7 @@ export function EditWithTamboButton({
   onOpenThread: onOpenThreadProp,
 }: EditWithTamboButtonProps) {
   const component = useTamboCurrentComponent();
-  const {
-    sendThreadMessage,
-    isIdle,
-    setInteractableSelected: setInteractableSelectedForInteraction,
-  } = useTambo();
+  const { sendThreadMessage, isIdle, setInteractableSelected } = useTambo();
   const { setIsOpen: setThreadPanelOpen, editorRef } = useMessageThreadPanel();
 
   const [prompt, setPrompt] = useState("");
@@ -134,7 +130,7 @@ export function EditWithTamboButton({
     const componentName = component?.componentName ?? "Unknown Component";
     const interactableId = component?.interactableId ?? "";
     if (interactableId) {
-      setInteractableSelectedForInteraction(interactableId, true);
+      setInteractableSelected(interactableId, true);
     }
 
     // Open the thread panel first
@@ -157,13 +153,7 @@ export function EditWithTamboButton({
         editor.appendText(messageToInsert);
       }
     }
-  }, [
-    prompt,
-    component,
-    setInteractableSelectedForInteraction,
-    onOpenThread,
-    editorRef,
-  ]);
+  }, [prompt, component, setInteractableSelected, onOpenThread, editorRef]);
 
   const handleMainAction = useCallback(() => {
     if (sendMode === "thread") {
--- a/cli/src/registry/edit-with-tambo-button/edit-with-tambo-button.tsx
+++ b/cli/src/registry/edit-with-tambo-button/edit-with-tambo-button.tsx
@@ -79,8 +79,7 @@ export function EditWithTamboButton({
   editorRef,
 }: EditWithTamboButtonProps) {
   const component = useTamboCurrentComponent();
-  const { sendThreadMessage, isIdle, setInteractableSelectedForInteraction } =
-    useTambo();
+  const { sendThreadMessage, isIdle, setInteractableSelected } = useTambo();
   const { setValue: setThreadInputValue } = useTamboThreadInput();
 
   const [prompt, setPrompt] = useState("");
@@ -147,7 +146,7 @@ export function EditWithTamboButton({
 
     const interactableId = component?.interactableId ?? "";
     if (interactableId) {
-      setInteractableSelectedForInteraction(interactableId, true);
+      setInteractableSelected(interactableId, true);
     }
 
     // Open the thread panel if callback provided
@@ -174,7 +173,7 @@ export function EditWithTamboButton({
     component,
     onOpenThread,
     editorRef,
-    setInteractableSelectedForInteraction,
+    setInteractableSelected,
     setThreadInputValue,
   ]);
 
--- a/showcase/src/components/tambo/edit-with-tambo-button.tsx
+++ b/showcase/src/components/tambo/edit-with-tambo-button.tsx
@@ -89,8 +89,7 @@ export function EditWithTamboButton({
   editorRef,
 }: EditWithTamboButtonProps) {
   const component = useTamboCurrentComponent();
-  const { sendThreadMessage, isIdle, setInteractableSelectedForInteraction } =
-    useTambo();
+  const { sendThreadMessage, isIdle, setInteractableSelected } = useTambo();
   const { setValue: setThreadInputValue } = useTamboThreadInput();
 
   const [prompt, setPrompt] = useState("");
@@ -155,10 +154,9 @@ export function EditWithTamboButton({
     // Save the message before clearing
     const messageToInsert = prompt.trim();
 
-    // Add the component as a context attachment for the next message
     const interactableId = component?.interactableId ?? "";
     if (interactableId) {
-      setInteractableSelectedForInteraction(interactableId, true);
+      setInteractableSelected(interactableId, true);
     }
 
     // Open the thread panel if callback provided
@@ -185,7 +183,7 @@ export function EditWithTamboButton({
     component,
     onOpenThread,
     editorRef,
-    setInteractableSelectedForInteraction,
+    setInteractableSelected,
     setThreadInputValue,
   ]);
 


michaelmagan:
What's the new way to add context with a message or are we removing this?

MichaelMilstead:
> What's the new way to add context with a message or are we removing this?

Thinking we add a function to the `context-helpers-provider` for this, like `addOneTimeContext(somecontextobject)`, so there's one "place" to manage context

--EDIT--
This PR keeps (and fixes) the ContextAttachment's "one-time message context" behavior

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -350,6 +352,37 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
...
+  );
+
+  const clearInteractableSelections = useCallback(() => {
+    setInteractableComponents((prev) =>
+      prev.map((c) => ({
+        ...c,
+        isSelectedForInteraction: false,
+      })),
+    );
+  }, []);
```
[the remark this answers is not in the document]
MichaelMilstead: done

docs/content/docs/concepts/additional-context/context-attachments.mdx
```diff
@@ -1,18 +1,18 @@
 ---
 title: Context Attachments
-description: Add visual context badges that automatically provide AI with focused information about what users are working on.
+description: Add visual context badges above the message input for user feedback
```
alecf: minor but this feels oddly prescriptive - because I think their primary use *is* for the LLM, and a secondary use is for the user - you might not want to show any badges at all

docs/content/docs/concepts/additional-context/context-attachments.mdx
```diff
@@ -1,18 +1,18 @@
...
 ---
 
-Context attachments enable you to display visual badges above the message input that automatically provide the AI with additional context. This helps the AI understand what the user is focused on and provide more relevant responses.
+Context attachments enable you to display visual badges above the message input.
```
alecf: yeah all of the language around this makes me feel like this is purely a visual thing, which is odd considering it's part of a non-visual sdk

docs/content/docs/concepts/additional-context/index.mdx
```diff
@@ -191,3 +171,11 @@ See [Interactable Components](/concepts/components/interactable-components) for
 
 - Helpers are plain functions. Return a value to include it, or null/undefined to skip.
 - The helper key becomes the context name in messages.
+
+## UI Features
+
+### Context Attachments (Badges)
+
+Context attachments are visual badges that appear above the message input. They provide UI feedback but do not send context to the AI. To send context, use context helpers separately.
```
alecf: wait they're not even sent to the AI then? At that point I'm wondering why they are in `@tambo-ai/react` at all?

packages/backend/src/prompt/decision-loop-prompts.ts
```diff
@@ -43,6 +43,17 @@ A JSON object containing:
 - **Maintain continuity**: Acknowledge what's currently displayed
 - **Make informed choices**: Use state to determine next appropriate actions
 
+### Interactable Components
+
+Tambo allows developers to place components on screen for you to help the user interact with.
```
alecf: I thin you need to call out "Interactable Components" as an explicit term, otherwise it just feels like it's an adjective, like you are talking about components that are interactable vs being read-only.


> Tambo's "Interactable Components", or sometimes just "Interactables" are...

And then in the code you need to be consistent about capitalization just so that it is totally unambiguous what you're referring to.

akhileshrangani4 (approved):
Looks so much cleaner! Thanks for taking this on! LGTM!

react-sdk/src/context-helpers/current-interactables-context-helper.ts
```diff
@@ -40,7 +40,7 @@ export const createInteractablesContextHelper = (
 
     return {
       description:
-        "These are the interactable components currently visible on the page that you can read and modify. Each component has an id, componentName, current props, current state,and optional schema. You can use tools to update these components on behalf of the user. Don't tell the user the ID of the components, only the name, unless they ask for it.",
+        "These are the interactable components currently visible on the page that you can read and modify. Each component has an id, componentName, current props, current state, and optional schema. You can use tools to update these components on behalf of the user. Don't tell the user the ID of the components, only the name, unless they ask for it. Some components may be selected for interaction, meaning Tambo should interact with them when responding to the next message.",
```
alecf: Again, rewrite this whole paragraph using specific terminology and proper capitalization.

alecf: (but I do wonder why we need to keep repeating this for every component, this should just be part of the system prompt)

packages/backend/src/prompt/decision-loop-prompts.ts
MichaelMilstead: done

react-sdk/src/context-helpers/current-interactables-context-helper.ts
MichaelMilstead: removed, and left the description only in the system prompt

apps/web/components/ui/tambo/edit-with-tambo-button.tsx
```diff
@@ -156,17 +133,13 @@ export function EditWithTamboButton({
...
+    // Add the component as a context attachment for the next message
     const componentName = component?.componentName ?? "Unknown Component";
     const interactableId = component?.interactableId ?? "";
-    addContextAttachment({
-      name: componentName,
-    });
+    addContextAttachment(componentName, componentName, "component");
+    if (interactableId) {
+      setInteractableSelectedForInteraction(interactableId, true);
+    }
```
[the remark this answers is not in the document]
MichaelMilstead: removed the call completely since `setInteractableSelectedForInteraction` adds the appropriate context to the message

cli/src/registry/edit-with-tambo-button/edit-with-tambo-button.tsx
```diff
@@ -170,16 +147,13 @@ export function EditWithTamboButton({
...
+    // Add the component as a context attachment for the next message
     const componentName = component?.componentName ?? "Unknown Component";
-    addContextAttachment({
-      name: componentName,
-    });
+    const interactableId = component?.interactableId ?? "";
+    addContextAttachment(componentName, componentName, "component");
+    if (interactableId) {
+      setInteractableSelectedForInteraction(interactableId, true);
+    }
```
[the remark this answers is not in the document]
MichaelMilstead: removed the call completely since `setInteractableSelectedForInteraction` adds the appropriate context to the message

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -350,6 +352,39 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
...
+        const updated = {
+          ...component,
+          isSelectedForInteraction: isSelected,
+        };
+
+        return prev.map((c) => (c.id === componentId ? updated : c));
+      });
+    },
+    [],
+  );
```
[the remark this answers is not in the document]
MichaelMilstead: done

react-sdk/src/providers/tambo-context-attachment-provider.tsx
```diff
@@ -1,225 +1,150 @@
...
+        id,
+        displayName,
+        context: contextValue,
+        type,
+      };
+      setAttachments((prev) => [...prev, attachment]);
+
+      return attachment;
     },
     [],
```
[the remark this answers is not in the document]
MichaelMilstead: we can decide on this later and update multiple uses

docs/content/docs/api-reference/react-hooks.mdx
```diff
@@ -328,43 +328,42 @@ Removes a helper by key so it is no longer included in outgoing messages.
...
-`const { customSuggestions } = useTamboContextAttachment()`
-
-Current custom suggestions that override auto-generated suggestions. Returns `null` if using auto-generated suggestions.
-
-### setCustomSuggestions
-
-`const { setCustomSuggestions } = useTamboContextAttachment()`
-
-Function to set custom suggestions or clear them by passing `null`.
+Function to remove all active context attachments at once. Also deregisters all associated context helpers. Typically called after message submission.
```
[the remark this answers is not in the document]
MichaelMilstead: fixed

packages/backend/src/prompt/decision-loop-prompts.ts
```diff
@@ -65,6 +83,45 @@ You MUST parse and understand these tags to provide contextually appropriate res
...
+      context_attachments_example: `
+        "context_attachments": [
+            {
+                "id": "File-13b",
+                "displayName": "SomeFile",
+                "context": "These are the contents of SomeFile.",
+                "type": "file"
+            },
+            ...more context attachments...
+        ]`,
```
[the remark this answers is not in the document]
MichaelMilstead: fixed

apps/web/components/observability/messages/message-content.test.tsx
```diff
@@ -14,6 +14,7 @@ jest.mock("@/components/ui/tambo/markdown-components", () => ({
 jest.mock("../utils", () => ({
   formatTime: (_date: Date) => "12:00 PM",
   getMessageContexts: () => [],
+  extractSelectedInteractablesFromAdditionalContext: () => [],
```
alecf: it seems really strange to mock this out, as this is a pure function in our own code

apps/web/components/observability/utils.ts
```diff
@@ -434,24 +433,37 @@ export function getMessageContexts(
 
   if (!additionalContext) return [];
 
-  // Look through all additionalContext values for selectedComponent
-  // Context helpers can store data under any key (e.g., context ID)
+  // Interactables are stored under the key "interactables" with a components array
+  const interactablesContext = additionalContext.interactables;
+
+  if (
```
alecf, line 439: I'm seeing a lot of this kind of paranoid type checking - seems like this could just be `if (!interactablesContext?.components)` ? not a huge deal but it's a lot of clutter if the issue is just that the types aren't well defined enough?

(I wonder if there's something we can add to AGENTS.md to clarify this)

apps/web/components/observability/utils.ts
```diff
@@ -434,24 +433,37 @@ export function getMessageContexts(
...
-      if (selectedComponent?.name) {
-        contexts.push({
-          id: key,
-          name: selectedComponent.name,
-          metadata: selectedComponent,
-        });
-      }
+  for (const component of components) {
+    if (
+      component &&
```
alecf, line 456: same here

react-sdk/src/model/tambo-interactable.ts
```diff
@@ -51,4 +53,11 @@ export interface TamboInteractableContext {
   getInteractableComponentState: (
     componentId: string,
   ) => Record<string, unknown> | undefined;
+  /** Set whether an interactable component is selected for interaction */
+  setInteractableSelectedForInteraction: (
```
alecf: this is a mouthful. How about `setSelectedInteractable`?

react-sdk/src/model/tambo-interactable.ts
```diff
@@ -8,6 +8,8 @@ export interface TamboInteractableComponent<
   id: string;
   /** Current props for the component */
   props: Props;
+  /** Whether the component is currently selected for interaction, meaning Tambo should interact with it when responding to the next message*/
+  isSelectedForInteraction?: boolean;
```
alecf: `...ForInteraction` feels very specific and vague at the same time. I don't know if this makes it into the LLM prompt in this form but I feel like we need a better term for this. `isActiveInteractable`? I guess below I also suggest maybe the equivalent of `isSelectedInteractable`? 

I think to me "Selected for interaction" doesn't really mean anything more specific than just "Selected" - like the system prompt should say something like "The Selected Interactable is the component that the LLM should focus on, updating its state and/or props"

alecf (approved):
some minor feedback, but otherwise looks good!

react-sdk/src/model/tambo-interactable.ts
MichaelMilstead: true, updated the field to `isSelected` and updated system prompt to describe what it means.

apps/web/components/observability/utils.ts
MichaelMilstead: I think there are a few places to update with a proper type for additionalContext, I might do that in a followup.

Wondering if we need to update message schema to store the pieces we know about ( contextAttachments, interactables ) as separate fields

PR #1590, opened 2025-12-18
feat(react-sdk): implement toolcall limit per tool

rakesh-paulraj1:
**Metadata & Models**: Added optional maxCalls: number property to TamboTool and ComponentContextToolMetadata interfaces, enabling per-tool limit configuration from the client side.
**Registry Mapping**: Updated mapTamboToolToContextTool to correctly extract and propagate maxCalls from tool definitions into the context metadata payload sent to the backend.


rakesh-paulraj1:
Tested manually by adding a boilerplate application in the monorepo 
1)
Screenshot From 2025-12-18 17-53-59
Setting up the Project level Restrictionn which leads to this 
Screenshot From 2025-12-18 17-54-54

2) Explicitly mentioning Higher maxCalls which overrides the project level tool-call limit per response
Screenshot From 2025-12-18 17-53-42

PR #1599, opened 2025-12-19
feat(cli): implement device authentication flow and session management

akhileshrangani4:
### Summary

Implements OAuth 2.0 Device Authorization Grant (RFC 8628) for CLI authentication with a unified sessions model. Users can authenticate the CLI via browser without manually copying API keys.

---

### Authentication Flow

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                           DEVICE AUTH FLOW                                  │
└─────────────────────────────────────────────────────────────────────────────┘

  ┌─────────┐                    ┌─────────┐                    ┌─────────┐
  │   CLI   │                    │ Server  │                    │ Browser │
  └────┬────┘                    └────┬────┘                    └────┬────┘
       │                              │                              │
       │  1. POST /deviceAuth.initiate│                              │
       │─────────────────────────────>│                              │
       │                              │                              │
       │  { deviceCode, userCode,     │                              │
       │    verificationUri }         │                              │
       │<─────────────────────────────│                              │
       │                              │                              │
       │  2. Display code & open URL  │                              │
       │─────────────────────────────────────────────────────────--->│
       │                              │                              │
       │     "Enter code: WDJB-MJHT"  │    3. User enters code       │
       │                              │         at /device           │
       │                              │<─────────────────────────────│
       │                              │                              │
       │  4. Poll /deviceAuth.poll    │  5. POST /deviceAuth.verify  │
       │─────────────────────────────>│<─────────────────────────────│
       │                              │                              │
       │     { status: "pending" }    │  6. Create session           │
       │<─────────────────────────────│     (source: "cli")          │
       │                              │                              │
       │  7. Poll again               │                              │
       │─────────────────────────────>│                              │
       │                              │                              │
       │  { status: "complete",       │                              │
       │    sessionToken, expiresAt } │                              │
       │<─────────────────────────────│                              │
       │                              │                              │
       │  8. Store token locally      │                              │
       │     (~/.local/share/tambo/)  │                              │
       ▼                              ▼                              ▼
```

---

### `tambo auth` Subcommands

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                         AUTH COMMAND STRUCTURE                              │
└─────────────────────────────────────────────────────────────────────────────┘

  tambo auth [subcommand]
       │
       ├── status [--quiet]          Show authentication status
       │                              • Local token check + server validation
       │                              • Exit code for scripts
       │
       ├── login                      Authenticate via browser
       │                              • Runs device auth flow
       │                              • Stores token locally
       │
       ├── logout [--force]           Clear stored credentials
       │
       ├── sessions                   List active CLI sessions
       │                              • Shows truncated IDs, created/expires
       │
       └── revoke-session             Revoke CLI session(s)
            │
            ├── (no flags)            Interactive session picker
            └── --all                 Revoke all CLI sessions
```

---

### Database Schema (Unified Sessions)

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                           DATABASE TABLES                                   │
└─────────────────────────────────────────────────────────────────────────────┘

  ┌─────────────────────┐
  │    auth.users       │
  │   (Supabase/Next)   │
  └──────────┬──────────┘
             │
             │ userId (cascade)
             ▼
  ┌─────────────────────────────────────────────────────┐
  │                     sessions                        │
  ├─────────────────────────────────────────────────────┤
  │  id              TEXT PK                            │
  │                  • Browser: gen_random_uuid()::text │
  │                  • CLI: 256-bit token (credential)  │
  │  user_id         UUID FK → auth.users               │
  │  source          ENUM('browser', 'cli')             │
  │  expires_at      TIMESTAMP (90 days default)        │
  │  created_at      TIMESTAMP                          │
  │  updated_at      TIMESTAMP                          │
  └──────────────────────────┬──────────────────────────┘
                             │
                             │ sessionId (after verify)
                             │
  ┌──────────────────────────┴──────────────────────────┐
  │                device_auth_codes                    │
  ├─────────────────────────────────────────────────────┤
  │  id              TEXT PK  (dac_xxx)                 │
  │  device_code     UUID     (DB-generated, CLI polls) │
  │  user_code       TEXT     (WDJBMJHT - user enters)  │
  │  user_id         UUID FK  (set after verify)        │
  │  session_id      TEXT FK  (set after verify)        │
  │  is_used         BOOLEAN  (prevent reuse)           │
  │  expires_at      TIMESTAMP (15 min TTL)             │
  └─────────────────────────────────────────────────────┘
```

---

#### `packages/db`
- **Unified `sessions` table** - Both CLI and browser sessions with `source` enum
- **`device_auth_codes` table** - `deviceCode` now `uuid` type with DB default
- Migration: `0082_lowly_valkyrie.sql`

#### `packages/core`
- **`SessionSource` enum** - Shared enum for session types (`browser`, `cli`)

#### `apps/web`
- **`deviceAuth` tRPC router** - initiate, verify, poll, listSessions, revokeSession, revokeAllSessions
- **CLI session validation** via `Authorization: Bearer` header
- **`/device` page** - Code entry with preserved `returnUrl`
- **`/cli-sessions` page** - View/revoke sessions
- **Browser session audit** - Logged to unified sessions table
- **Legal acceptance `returnUrl`** - New CLI users return to `/device` after accepting

#### `cli/`
- **tRPC vanilla client** with `StubRouter` interface for type-safe API calls
- **`tambo auth` command** - status, login, logout, sessions, revoke-session
- **Interactive session picker** for revoke-session (no `--id` flag)
- **Token storage** via `env-paths` (XDG-compliant with explicit env var mapping)
- **25 new auth tests**

---

### Security

| Feature | Implementation |
|---------|----------------|
| Token entropy | 256-bit cryptographically secure |
| User code | 32-char set, no modulo bias |
| Storage permissions | Unix: 0600 (owner only) |
| Device code | UUID, DB-generated |
| Rate limiting | 5s minimum poll interval |
| Expiry | Server-authoritative, 90 days |
| Unified errors | All auth failures → `UNAUTHORIZED` |

---

### Breaking Changes

None.


apps/web/app/(authed)/device/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <div className="w-full max-w-md">
+        <div className="text-center mb-8">
+          <h1 className="text-3xl font-bold tracking-tight">
+            Authorize CLI Access
+          </h1>
+          <p className="text-muted-foreground mt-2">
+            Grant your terminal access to your tambo account
+          </p>
+        </div>
+        <DeviceCodeForm initialCode={initialCode} />
```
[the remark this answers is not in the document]
lachieh: Charlie is wrong here. Next 14 had objects but 15+ uses promises.

apps/web/app/(authed)/device/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+interface DevicePageProps {
+  searchParams: Promise<{ user_code?: string }>;
+}
+
+export default async function DevicePage({ searchParams }: DevicePageProps) {
+  const params = await searchParams;
+  const initialCode = params.user_code ?? "";
+
```
lachieh: You can use the `PageProps` helper type here which shouldn't need to be imported. If it doesn't work, let me know.

Additionally, you can use the `use` hook here

```suggestion
export default async function DevicePage({ searchParams }: PageProps<'/blog/[slug]'>) {
  const initialCode = use(searchParams).initialCode ?? "";
```

cli/src/cli.ts
```diff
@@ -187,6 +196,27 @@ ${chalk.bold("Templates")}
...
+    options: ["quiet", "force", "revoke", "revoke-all"],
+    examples: [
+      `$ ${chalk.cyan("tambo auth")}                       # Show auth status`,
+      `$ ${chalk.cyan("tambo auth login")}                 # Authenticate`,
+      `$ ${chalk.cyan("tambo auth logout --force")}        # Logout without prompt`,
+      `$ ${chalk.cyan("tambo auth sessions")}              # List sessions`,
+      `$ ${chalk.cyan("tambo auth sessions --revoke-all")} # Revoke all sessions`,
+    ],
+    exampleTitle: "Authentication",
+  },
```
lachieh, line 214: Let's implement these as subcommands, similar to how `tambo update` and `tambo update installed` are configured.

```suggestion
  auth: {
    command: "auth",
    syntax: "auth [subcommand]",
    description: "Manage authentication (status, login, logout, sessions)",
    usage: [
      `$ ${chalk.cyan("tambo auth")} [subcommand] [options]`,
      `$ ${chalk.cyan("tambo auth status")}   ${chalk.dim("Show authentication status")}`,
      `$ ${chalk.cyan("tambo auth login")}    ${chalk.dim("Authenticate via browser")}`,
      `$ ${chalk.cyan("tambo auth logout")}   ${chalk.dim("Clear stored credentials")}`,
      `$ ${chalk.cyan("tambo auth sessions")} ${chalk.dim("List/manage CLI sessions")}`,
    ],
    options: ["quiet"],
    examples: [
      `$ ${chalk.cyan("tambo auth")}                       # Show auth status`,
      `$ ${chalk.cyan("tambo auth login")}                 # Authenticate`,
      `$ ${chalk.cyan("tambo auth logout --force")}        # Logout without prompt`,
      `$ ${chalk.cyan("tambo auth sessions")}              # List sessions`,
      `$ ${chalk.cyan("tambo auth sessions --revoke-all")} # Revoke all sessions`,
    ],
    exampleTitle: "Authentication",
  },
  auth-status: {/*...*/}, 
  auth-login: {/*...*/}, 
  auth-logout: {/*...*/}, 
  auth-sessions {
    command: "auth sessions",
    syntax: "auth sessions",
    description: "List auth sessions",
    usage: [
      `$ ${chalk.cyan("tambo auth sessions [options] ${chalk.dim("(list all sessions)")}`,
      // ...
    ],
    options: [],
    examples: [
      // ...
    ]
  },
```

cli/src/cli.ts
```diff
@@ -62,6 +67,10 @@ const OPTION_DOCS: Record<string, string> = {
   template: `${chalk.yellow("--template, -t <name>")}  Template to use: standard, analytics`,
   "init-git": `${chalk.yellow("--init-git")}           Initialize git repository automatically`,
   "dry-run": `${chalk.yellow("--dry-run")}            Preview changes without applying them`,
+  quiet: `${chalk.yellow("--quiet, -q")}          Exit code 0 if authenticated, 1 otherwise`,
+  force: `${chalk.yellow("--force, -f")}          Skip confirmation prompts`,
+  revoke: `${chalk.yellow("--revoke <id>")}        Revoke a specific session`,
+  "revoke-all": `${chalk.yellow("--revoke-all")}         Revoke all CLI sessions`,
```
lachieh: Do these need to be global options?

lachieh (changes requested):
Looking good!

I'm only going to comment on the UI, CLI UX; I'll leave the auth flow stuff to @alecf but from a quick glance it looks good to me.

As general advice on building CLI patterns:
- **Commands** should change the type of action (`auth login`)
- **Options (aka flags)** should change how the command behaves (`auth status --output yaml` — example, we don't need output formatting right now)

You've already got it correct with `auth`, `auth login`, `auth logout`, but I'd move the sessions options into their own commands:
- `auth sessions --revoke` > `auth revoke-session --id <session-id>`
- `auth sessions --revoke-all` > `auth revoke-session --all`
This way, there's no change in what is happening; `auth sessions` will only ever list sessions, and `auth revoke` will only ever revoke sessions, the options just change what is being revoked. I could see an argument being used like `auth revoke-session <session-id>`  instead of `auth revoke-session --id <session-id>` but the general advice on args vs options/flags is that args are harder to alter:
> #### [Arguments and Flags](https://clig.dev/#arguments-and-flags)
> **Prefer flags to args.** It’s a bit more typing, but it makes it much clearer what is going on. It also makes it easier to make changes to how you accept input in the future. Sometimes when using args, it’s impossible to add new input without breaking existing behavior or creating ambiguity.

cli/src/cli.ts
akhileshrangani4: ohh right, these do not need to be global options.

apps/web/app/(authed)/device/page.tsx
akhileshrangani4: aah yes, you are right.

cli/src/cli.ts
akhileshrangani4: yup, will do that.

cli/src/lib/token-storage.ts
```diff
@@ -0,0 +1,198 @@
...
+ * If TAMBO_SESSION_TOKEN is set, it takes precedence over stored token
+ */
+export function getEffectiveSessionToken(): string | null {
+  const envToken = process.env.TAMBO_SESSION_TOKEN;
+  if (envToken) {
+    return envToken;
+  }
+
+  return getSessionToken();
+}
```
lachieh, line 228: Good, love the env var override. I think I prefer `TAMBO_TOKEN` so that it mirrors other tools like `GITHUB_TOKEN` and `TURBO_TOKEN`

cli/src/lib/token-storage.ts
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * Ensure the data directory exists with appropriate permissions
+ */
+function ensureDataDir(): void {
+  const dataDir = paths.data;
+  if (!existsSync(dataDir)) {
+    mkdirSync(dataDir, { recursive: true, mode: 0o700 });
+  }
+}
```
lachieh, line 82: I'm going to be a bit pedantic here and say I want to support XDG_ spec on macos too.


```suggestion
/**
 * Name of directory to store state/config/cache
 */
const DIR_PREFIX = 'tambo';

/**
 * Get OS-specific paths for tambo state storage
 * Uses XDG Base Directory Specification on Linux
 *
 * Data paths (for auth tokens and persistent data):
 * - macOS: ~/Library/Application Support/tambo/ (or $XDG_DATA_HOME/tambo/)
 * - Linux: ~/.local/share/tambo/ (or $XDG_DATA_HOME/tambo/)
 * - Windows: %LOCALAPPDATA%/tambo/
 */
function getDir(type: 'cache' | 'config' | 'data'): string {
  const paths = envPaths(DIR_PREFIX, { suffix: "" });
  const dir = process.env[`TAMBO_${type.toUpperCase()}_DIR`] || process.env[`XDG_${type.toUpperCase()}_HOME`] || paths[type];
  return join(dir, DIR_PREFIX);
}

/**
 * Path to the auth token file
 */
const AUTH_FILE_NAME = "auth.json";

function getAuthFilePath(): string {
  return join(getDir('data'), AUTH_FILE_NAME);
}

/**
 * Ensure the data directory exists with appropriate permissions
 */
function ensureDataDir(): void {
  const dataDir = getDir('data');
  if (!existsSync(dataDir)) {
    mkdirSync(dataDir, { recursive: true, mode: 0o700 });
  }
}
```

lachieh (changes requested):
small change with my opinionated dirs config and some env variable naming but it's close!

apps/web/server/api/routers/device-auth.ts
```diff
@@ -0,0 +1,444 @@
...
+  const proto = headers.get("x-forwarded-proto") ?? "https";
+
+  if (forwardedHost) {
+    return `${proto}://${forwardedHost}`;
+  }
+  if (host) {
+    return `${proto}://${host}`;
+  }
+
+  // Last resort fallback
```
alecf: I think we should throw an exception instead of returning this

apps/web/server/api/routers/device-auth.ts
```diff
@@ -0,0 +1,444 @@
...
+
+/**
+ * Try to get the most recent browser session ID for a user
+ * Used to link CLI sessions to the browser session that authorized them
+ */
+async function getBrowserSessionId(
+  db: HydraDb,
+  userId: string,
+): Promise<string | null> {
+  const [session] = await db
```
alecf: very minor but you can use db.query.sessions.findFirst()

cli/src/lib/token-storage.ts
akhileshrangani4: thanks, I have changed it to `TAMBO_TOKEN` now.

cli/src/lib/token-storage.ts
akhileshrangani4: made these changes too!

akhileshrangani4:
@lachieh Thanks for your review, I have made the changes you suggested.

apps/web/server/api/routers/device-auth.ts
```diff
@@ -0,0 +1,444 @@
...
+ * Generate a random alphanumeric string for user codes
+ * Uses uppercase letters and digits, excluding confusing characters (0, O, I, 1, L)
+ */
+function generateUserCode(length: number = 8): string {
+  const chars = "ABCDEFGHJKMNPQRSTUVWXYZ23456789";
+  let result = "";
+  const randomValues = new Uint8Array(length);
+  crypto.getRandomValues(randomValues);
+  for (let i = 0; i < length; i++) {
+    result += chars[randomValues[i] % chars.length];
```
alecf: this introduces bias into the random values because 30 does not go into 256 evenly. 

Suggestion from chatgpt:
```
function generateUserCode(length = 8): string {
  const chars = "ABCDEFGHJKLMNPQRSTUVWXYZ23456789_"; // 32
  const bytes = new Uint8Array(length);
  crypto.getRandomValues(bytes);
  return Array.from(bytes, b => chars[b & 31]).join("");
}
```
Note it just added 2 more characters to the end of that. not sure if `-` and `_`

alecf: I wonder if a better way to do this is just to let postgres do it, so you don't worry about creating a session with a null id, you can add a default to the column:

`split_part(gen_random_uuid()::text, '-', 1)`

apps/web/server/api/routers/device-auth.ts
```diff
@@ -0,0 +1,444 @@
...
+
+      // Try to get browser session ID for reference (optional)
+      // This links the CLI session to the browser session that authorized it
+      const browserSessionId = ctx.session?.user
+        ? await getBrowserSessionId(ctx.db, ctx.user.id)
+        : null;
+
+      // Create CLI session first (required by FK constraint on device_auth_codes)
+      await ctx.db.insert(schema.cliSessions).values({
+        id: cliSessionToken,
```
alecf: Yeah I think in general any place where we can let the database generate these codes, the better

apps/web/server/api/routers/device-auth.ts
```diff
@@ -0,0 +1,444 @@
...
+      const [existingCode] = await ctx.db
+        .select()
+        .from(schema.deviceAuthCodes)
+        .where(eq(schema.deviceAuthCodes.userCode, input.userCode))
+        .limit(1);
+
+      if (existingCode) {
+        if (existingCode.isUsed) {
+          throw new TRPCError({
+            code: "CONFLICT",
```
alecf: minor but I think UNAUTHORIZED is better for all of the errors here.

packages/db/src/schema.ts
```diff
@@ -104,6 +104,115 @@ export const sessions = authSchema.table(
...
+export const deviceAuthCodes = pgTable(
+  "device_auth_codes",
+  ({ text, timestamp, uuid, boolean }) => ({
+    id: text("id")
+      .primaryKey()
+      .notNull()
+      .unique()
+      .default(sql`generate_custom_id('dac_')`),
+    // Server-generated UUID used by CLI to poll for completion
+    deviceCode: text("device_code").notNull().unique(),
```
alecf: If this is a server-generated UUID, then use the `uuid()` data type for the column and use `gen_random_uuid()` as the default

packages/db/src/schema.ts
```diff
@@ -104,6 +104,115 @@ export const sessions = authSchema.table(
...
+ * References the browser session that authorized it, but independent lifecycle.
+ *
+ * Token stored as `id` is a 256-bit high-entropy credential - NEVER log it.
+ */
+export const cliSessions = pgTable(
+  "cli_sessions",
+  ({ text, timestamp, uuid }) => ({
+    // High-entropy session token (256 bits, base64url encoded)
+    // This IS the credential - treat as secret
+    id: text("id").primaryKey().notNull(),
```
alecf: similarly, primary keys should be generated database-side. I'm not sure where the cliSessions.id is exposed though. It seems like it is only used in joins, so it seems like it should be fine to use `bigserial` or `generate_custom_id` (I'd prefer bigserial) - I see the warnings about not exposing it but I think that's generally true of a lot of primary keys - perhaps a warning saying that the id should never be used publically

I think in general all of these questions are why I'd much rather us use `auth.sessions` and just make this table have a column `auth_session_id` that we use to resolve to a user id.

packages/db/src/schema.ts
```diff
@@ -104,6 +104,115 @@ export const sessions = authSchema.table(
...
+    userId: uuid("user_id")
+      .references(() => authUsers.id, { onDelete: "cascade" })
+      .notNull(),
+    // The browser session that authorized this CLI session (for audit/reference)
+    // Set to null if browser session is deleted - CLI session survives
+    browserSessionId: text("browser_session_id").references(() => sessions.id, {
+      onDelete: "set null",
+    }),
+    // When this CLI session expires (required, typically 90 days)
+    notAfter: timestamp("not_after", { withTimezone: true }).notNull(),
```
alecf: call this `expiresAt` (try to avoid creating columns defined as the negative of something!)

apps/web/server/api/trpc.ts
```diff
@@ -32,6 +32,62 @@ export type Context = {
...
+    })
+    .from(schema.cliSessions)
+    .innerJoin(
+      schema.authUsers,
+      eq(schema.cliSessions.userId, schema.authUsers.id),
+    )
+    .where(
+      and(
+        eq(schema.cliSessions.id, sessionToken),
+        gt(schema.cliSessions.notAfter, now),
```
alecf: `now` should be done on the server, not in javascript, I believe its something like

> sql\`now()`

apps/web/server/api/trpc.ts
```diff
@@ -32,6 +32,62 @@ export type Context = {
...
+      ),
+    )
+    .limit(1);
+
+  if (result.length === 0) {
+    return null;
```
alecf: it seems like if the validation fails, this should throw an exception, so this gets treated like a full error, not just treated as a session that doesn't happen to be logged in

alecf (commented):
There's a lot here... here are a bunch of nits around the database schema and validation but in general I'm concerned with how much manual validation we're doing and the potential for security holes. I know the comments say this lets it have "clean separation" but for auth stuff, I view that as "an additional attack surface"

I would really prefer that we emit a session token that can be validated as an actual session... 

Related, I feel like even in the current design has a lot of tables with `userId` in it - feels like there is opportunity to to exploit something because it's not clear which `userId` is authoritative (and again, if we were using auth.session, we would know that the session.userId is authoritative)

apps/web/server/api/routers/device-auth.ts
akhileshrangani4: I have updated this as you recommended, and I hope after our discussion on friday, you now know why I am generating a human readable code.

apps/web/server/api/routers/device-auth.ts
akhileshrangani4: yup, did it.

apps/web/server/api/routers/device-auth.ts
akhileshrangani4: this is now outdated.

apps/web/server/api/routers/device-auth.ts
akhileshrangani4: yup, using db generated codes wherever possible.

apps/web/server/api/routers/device-auth.ts
akhileshrangani4: yup, did it.

packages/db/src/schema.ts
akhileshrangani4: doing it now.

packages/db/src/schema.ts
akhileshrangani4: done.

apps/web/server/api/trpc.ts
akhileshrangani4: yup, updated it.

packages/db/src/schema.ts
akhileshrangani4: The `sessions.id` for CLI sessions can't be `bigserial` because **the ID IS the credential** - it's the token that the CLI stores locally and sends in the `Authorization` header to authenticate API requests. Using predictable IDs (1, 2, 3...) would be a security issue.

For browser sessions, the ID uses the DB default (`gen_random_uuid()::text`) since it's just an audit trail - browser auth goes through NextAuth/cookies.

To follow your suggestion fully, I'd need to split into:
- `id`: bigserial (internal, safe to expose)
- `token`: text (CLI credential, generated server-side)

That's a bigger refactor. Current approach works and is a common pattern (session token as PK). Happy to refactor if you prefer the split approach though.

apps/web/server/api/trpc.ts
akhileshrangani4: yes, updated it.

apps/web/components/auth/nextauth-layout-wrapper.tsx
```diff
@@ -39,7 +42,7 @@ export const NextAuthLayoutWrapper: FC<NextAuthLayoutWrapperProps> = ({
       // Redirect to legal acceptance if not accepted
       router.push("/legal-acceptance");
     }
-  }, [session, status, router, legalStatus, pathname]);
+  }, [session, status, router, legalStatus, pathname, searchParams]);
```
alecf: searchParams is likely the same object, even if the searchParams change. Just for correctness, calculate `search` outside the effect, and then make the effect dependent on `search`

apps/web/components/device-auth/device-code-form.tsx
```diff
@@ -0,0 +1,167 @@
...
+const deviceCodeSchema = z.object({
+  userCode: z
+    .string()
+    .min(1, "Please enter a device code")
+    .transform((val) => val.replace(/-/g, "").trim())
+    .pipe(
+      z
+        .string()
+        .length(8, "Device code must be 8 characters")
+        .regex(/^[A-Z0-9]+$/i, "Device code must be alphanumeric"),
```
alecf: if you're going to use a regex, might as well use something that validates the exact format, e.g. `/[A-Z0-9]{4}-[A-Z0-9]{4}/i`

apps/web/components/device-auth/device-code-form.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  return (
+    <div className="rounded-lg border bg-card p-8">
+      <div className="text-center mb-6">
+        <p className="text-sm text-muted-foreground">
+          Enter the code displayed in your terminal
+        </p>
+      </div>
+
+      <Form {...form}>
```
alecf, line 109: I don't think you want ...form here, it will pass all sorts of unrelated props to `<form>`

apps/web/components/device-auth/device-code-form.tsx
```diff
@@ -0,0 +1,167 @@
...
+        <form onSubmit={form.handleSubmit(onSubmit)} className="space-y-6">
```
alecf, line 110: wait and then why a `<form>` inside a `<Form>`?

oh.. because `Form` is just `FormProvider`? 

Ok nevermind, shadcn form is wierd. IMO our shadcn form shouldn't export `FormProvider` as `Form`, it should just be called `FormProvider`

apps/web/lib/auth.ts
```diff
@@ -105,27 +110,50 @@ export const authOptions: NextAuthOptions = {
...
-      // Redirect to a generic unauthorized page. We MUST NOT leak the
-      // restricted domain or full incoming email in the response.
-      return "/unauthorized";
+      // Create audit entry for browser session
+      // This is just a log - with JWT strategy, the JWT controls actual auth
+      // ID and expiresAt are generated by Postgres
+      // Note: user.id from OAuth providers (e.g., GitHub) may not be a UUID,
+      // so we look up the database user by email to get the actual UUID
+      try {
+        const db = getDb(env.DATABASE_URL);
```
alecf, line 140: I'm a little confused: seems like `user.id` should be how we're looking stuff up, not by email? is `user.id` not filled in?

apps/web/lib/auth.ts
```diff
@@ -105,27 +110,50 @@ export const authOptions: NextAuthOptions = {
...
+        const dbUser = await db.query.authUsers.findFirst({
+          where: eq(schema.authUsers.email, email),
+          columns: { id: true },
+        });
+        if (dbUser) {
```
alecf: if you don't find the dbuser, you're still letting them log in? I mean maybe there's some way to get here without finding the user? this should either return false, or have a comment explaining..

packages/db/src/schema.ts
```diff
@@ -104,6 +109,128 @@ export const sessions = authSchema.table(
   ],
 );
 
+/**
+ * Session source enum - identifies where a session originated
+ * - browser: Created via NextAuth signIn (audit entry only, not revocable - JWT controls auth)
+ * - cli: Created via device auth flow (revocable, stored credential)
+ */
+export const sessionSourceEnum = pgEnum("session_source", ["browser", "cli"]);
```
alecf: lets stay away from native enums for now - they make future migrations complicated... look elsewhere in the file for how we're handling enums

apps/web/lib/auth.ts
```diff
@@ -105,27 +110,50 @@ export const authOptions: NextAuthOptions = {
...
+          await db.insert(schema.sessions).values({
+            userId: dbUser.id,
+            source: "browser",
+            expiresAt: sql`now() + interval '30 days'`,
```
alecf: Yes, this is the way to do expiresAt - can you just make this the default for this column and then leave out any manual expiresAt setting? There are a few other places in the PR where you're setting expiresAt from JS

apps/web/server/api/trpc.ts
```diff
@@ -32,6 +32,65 @@ export type Context = {
...
+ * - Tokens should NEVER be logged - they grant full account access
+ * - This is enforced here and when creating sessions in device-auth.ts
+ */
+async function validateSession(
+  db: HydraDb,
+  sessionToken: string,
+): Promise<NonNullable<Context["user"]>> {
+  // Query the session and join with user
+  // Sessions must have expiresAt set and not be expired
+  const result = await db
```
alecf: can you move all the db operations into the db package with specific names, so we have `operations.validateSession()` etc?

cli/src/lib/api-client.ts
```diff
@@ -0,0 +1,149 @@
...
+        }
+        return res;
+      },
+    }),
+  ],
+});
+
+// ============================================================================
+// Typed Wrappers
+// ============================================================================
```
alecf, line 135: hold on you don't need to write a typed wrapper if you just pass the right kind  of typed router to createTRPCClient. I think it will look something like this:

```ts
interface StubRouter {
  deviceAuth: {
    initiate: () => Promise<InitiateResponse>;
  }
};

const client = createTRPCClient<StubRouter>(....)
```

cli/src/lib/api-client.ts
```diff
@@ -0,0 +1,149 @@
...
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+// Also export raw client for direct tRPC access if needed
+export const api = client;
+
+export { TRPCClientError };
```
alecf: don't re-export 3rd party types, don't re-export api if you don't need it.. also do you really need to expose all these methods? because every single method you expose via cli is now something we have to be careful about changing. Make sure this is the bare minimum set of methods in the router

cli/src/lib/device-auth.ts
```diff
@@ -0,0 +1,288 @@
...
+    this.name = "DeviceAuthError";
+  }
+}
+
+/**
+ * Options for the device auth flow
+ */
+interface DeviceAuthOptions {
+  /** Whether to automatically open the browser */
+  openBrowser?: boolean;
```
alecf: don't make this stuff optional - either leave it out and hardcode the defaults, or make them required - but nobody else needs this flexibility, it will just complicate things later having to deal with optional settings

cli/src/lib/token-storage.ts
```diff
@@ -0,0 +1,211 @@
...
+ * Uses XDG Base Directory Specification on Linux
+ *
+ * Data paths (for auth tokens and persistent data):
+ * - macOS: ~/Library/Application Support/tambo/ (or $XDG_DATA_HOME/tambo/)
+ * - Linux: ~/.local/share/tambo/ (or $XDG_DATA_HOME/tambo/)
+ * - Windows: %LOCALAPPDATA%/tambo/
+ */
+function getDir(type: "cache" | "config" | "data"): string {
+  const paths = envPaths(DIR_PREFIX, { suffix: "" });
+  const envKey = `XDG_${type.toUpperCase()}_HOME`;
```
alecf: do not mess around with upper/lower casing of values like this, especialy going from one api (`envPaths`) to another (XDG_ environment variables) - do explicit mapping either with a global dictionary or a switch statement... especially for stuff like this, it makes sure some runtime value doesn't get injected in a strange way into this value.

this also lets you manually find the explicit string `XDG_CACHE_HOME` etc in the code

cli/src/lib/token-storage.ts
```diff
@@ -0,0 +1,211 @@
...
+  const xdgDir = process.env[envKey];
+  if (xdgDir) {
+    return join(xdgDir, DIR_PREFIX);
+  }
+  return paths[type];
```
alecf, line 62: minor but don't call `envPaths()` until you need it (meaning, move it after the xdgDir check

cli/src/lib/token-storage.ts
```diff
@@ -0,0 +1,211 @@
...
+  if (!existsSync(filePath)) {
+    return null;
+  }
+
+  try {
+    const data = readFileSync(filePath, { encoding: "utf-8" });
+    const token = JSON.parse(data) as StoredToken;
+
+    // Validate required fields
+    if (!token.sessionToken || !token.expiresAt || !token.user) {
```
alecf, line 127: how about a warning to the console if the token is an invalid shape?

apps/web/components/device-auth/device-code-form.tsx
lachieh: Yeah, this is how just the regular shadcn form works. I was going to say that on the other comment re: `...form` but I agree. It's probably the only component that doesn't fall on either side of "make it transparent" or "make it easy". It feels like shadcn was trying to do the latter by abstracting everything, but then realized that the `<form>` element shouldn't always be at the same level in the DOM as the FormProvider.

All that to say... I'm good with this as is, but we might want to think about a better set of react-hook-form helpers in the future.

cli/src/lib/token-storage.ts
lachieh: This was my suggestion. 100% on explicitness. @akhileshrangani4 something like this:

```ts
const XDG_DIRS = {
  cache: "XDG_CACHE_HOME",
  config: "XDG_CONFIG_HOME",
  data: "XDG_DATA_HOME",
} as const

/**
 * Get OS-specific paths for tambo state storage
 * Uses XDG Base Directory Specification on Linux
 *
 * Data paths (for auth tokens and persistent data):
 * - macOS: ~/Library/Application Support/tambo/ (or $XDG_DATA_HOME/tambo/)
 * - Linux: ~/.local/share/tambo/ (or $XDG_DATA_HOME/tambo/)
 * - Windows: %LOCALAPPDATA%/tambo/
 */
function getDir(type: keyof XDG_DIRS): string {
  const envKey = DIR_PATHS[type];
  const xdgDir = process.env[envKey];
  if (xdgDir) {
    return join(xdgDir, DIR_PREFIX);
  }
  const paths = envPaths(DIR_PREFIX, { suffix: "" });
  return paths[type];
}
```

apps/web/components/device-auth/device-code-form.tsx
akhileshrangani4: yes, this is just because shadcn form is a little weird, but I do agree about the export, I will make a follow up PR to make the export `FormProvider` instead of `Form`.

apps/web/lib/auth.ts
akhileshrangani4: yes, you are right, I will change this.

apps/web/lib/auth.ts
akhileshrangani4: The user is guaranteed to exist here around lines 150-158 (the adapter creates/finds the user before the signIn callback returns). This path (sessions table) is audit-only, so a DB failure shouldn’t block auth. I have added a comment explaining this.

packages/db/src/schema.ts
akhileshrangani4: yup, saw how it has been done for other enums, implemented it in a similar way.

apps/web/lib/auth.ts
akhileshrangani4: ohh yes, that would be clean, will do.

apps/web/server/api/trpc.ts
akhileshrangani4: yes, done

cli/src/lib/api-client.ts
akhileshrangani4: I have removed the export of the error type. And about the methods, I am using all those methods one way or another in the cli, I have actually only added methods in the router that I actually need.

cli/src/lib/api-client.ts
akhileshrangani4: ohh, okay, will do this.

cli/src/lib/device-auth.ts
akhileshrangani4: yeah, you are right, making this required.

cli/src/lib/token-storage.ts
akhileshrangani4: yes, adding it.

akhileshrangani4 (commented):
I have made all the changes based on the review, @alecf and @lachieh, thanks!

apps/web/app/(authed)/cli-sessions/page.tsx
```diff
@@ -0,0 +1,17 @@
+"use client";
```
alecf: totally minor but don'd add "use client" when the page can be rendered server-side - just leave that out to let nextjs decide (may apply elsewhere in the PR)

packages/db/src/schema.ts
```diff
@@ -104,6 +109,125 @@ export const sessions = authSchema.table(
...
+      onDelete: "cascade",
+    }),
+    // The session ID created when verification completes
+    sessionId: text("session_id").references(() => sessions.id, {
+      onDelete: "cascade",
+    }),
+    // Prevent reuse of codes
+    isUsed: boolean("is_used").default(false).notNull(),
+    // Codes expire after 15 minutes
+    expiresAt: timestamp("expires_at", { withTimezone: true }).notNull(),
```
alecf: this can actually be defaulted too:

```suggestion
    expiresAt: timestamp("expires_at", { withTimezone: true })
      .default(sql`now() + interval '15 minutes'`)
      .notNull(),
```

apps/web/server/api/routers/device-auth.ts
```diff
@@ -0,0 +1,426 @@
...
+  /**
+   * Initiate device auth flow (called by CLI)
+   *
+   * Device code is generated by Postgres (gen_random_uuid).
+   * User code is generated in app code for custom human-readable charset.
+   * Returns absolute verification URLs per RFC 8628.
+   */
+  initiate: publicProcedure.mutation(async ({ ctx }) => {
+    const userCode = generateUserCode(8);
+    const expiresAt = new Date(Date.now() + CODE_EXPIRY_MINUTES * 60 * 1000);
```
alecf: once you add the postgres default, then you can also remove this and the `isUsed: false` below

apps/web/app/(authed)/cli-sessions/page.tsx
akhileshrangani4: done

packages/db/src/schema.ts
akhileshrangani4: ohh, okay, doing this.

apps/web/server/api/routers/device-auth.ts
akhileshrangani4: true, doing it.

PR #1619, opened 2025-12-23
chore(repo): update testing and linting config

lachieh:
## Summary

- Simplify VSCode Jest runner config for CLI package (use turbo instead of separate commands)
- Ignore coverage folders in ESLint base config
- Clean up CLI jest coverage collection patterns
- Use `eslint/config` defineConfig instead of typescript-eslint in apps/api


lachieh:
> The biggest issue is the Jest coverage glob change in `cli/jest.config.ts`: `*.(ts|tsx)` is likely an invalid pattern for Jest’s micromatch and can result in no coverage being collected.

This is correct. Coverage stays the same, so simpler rules is nicer.

> In `apps/api/eslint.config.mjs`, the new `defineConfig(...config, …)` call introduces a runtime coupling to the shape of the exported base config; normalizing to an array would make it more resilient.

Not really beneficial. If `config` changes to anything other than an array (which it is now) then the `defineConfig` function will throw a type warning anyway.

alecf (approved):

PR #1647, opened 2026-01-05
feat(db): add anon role and RLS policies for device auth flow

akhileshrangani4:
## Summary

Adds database role configuration and RLS policies to support the device auth flow for CLI authentication, with a **two-step auth flow** that avoids accessing `auth.users` from the anon role.

### Two-Step Device Auth Flow (New)

The poll endpoint runs as `anon` role which **cannot access `auth.users`** (intentional security constraint). This implements a two-step flow:

1. CLI polls until auth completes → receives `sessionToken` only
2. CLI saves temp token (enables authenticated requests)
3. CLI calls `user.getUser` to fetch user info via authenticated endpoint
4. CLI updates token with complete user data

**Files changed:**
- `apps/web/server/api/routers/device-auth.ts` - Remove `auth.users` join from poll endpoint
- `cli/src/lib/api-client.ts` - Add `UserInfo` type and `user.getUser` stub
- `cli/src/lib/device-auth.ts` - Implement two-step flow with proper error handling
- `cli/src/lib/token-storage.ts` - Support bootstrap phase (token valid before user.id populated)

#### `0000_init_setup.sql`
- Create `anon` role for unauthenticated requests (device auth initiation/polling)
- Grant `anon` and `authenticated` roles to `postgres` (required for `SET LOCAL ROLE`)
- **Security fix:** Only `authenticated` role gets auth schema access (not anon)

#### `0083_past_agent_zero.sql`
- Grant `anon` SELECT/INSERT/UPDATE on `public.device_auth_codes` table
- Grant `anon` SELECT on `public.sessions` table (poll endpoint queries session after verification)
- Enable RLS on `device_auth_codes` table
- Add RLS policies for `anon` (insert/select/update) and `authenticated` (all)

### `schema.ts`
- Define `anonRole` with `createRole: false` (role created in migration)
- Add RLS policies for `deviceAuthCodes` table

## Why

1. **Supabase compatibility** - Cannot modify auth schema permissions on Supabase prod
2. **Docker simplicity** - Vanilla PostgreSQL users shouldn't need to replicate complex Supabase roles
3. **Security** - anon role should never have access to user data in `auth.users`

## Production Impact

**No breaking changes for Supabase production:**

- **`0000_init_setup.sql` changes** - Won't run on prod. This migration was already applied, and Drizzle tracks applied migrations.
- **`0083` migration** - Will run on prod but is safe because:
  - Supabase already has `anon` and `authenticated` roles built-in
  - GRANTs are additive (just adds permissions, doesn't remove anything)
  - RLS policies are new additions
  - Migrations run in order (0082 creates tables → 0083 adds grants)

## Testing

- [x] Fresh DB setup works with `./scripts/cloud/init-database.sh`
- [x] Device auth flow (initiate → verify → poll) works end-to-end
- [x] Two-step flow: poll returns token → CLI fetches user info → complete auth

### DIFF
--- a/cli/src/lib/token-storage.ts
+++ b/cli/src/lib/token-storage.ts
@@ -124,7 +124,6 @@ export function loadToken(): StoredToken | null {
     const token = JSON.parse(data) as StoredToken;
 
     // Validate basic structure (sessionToken, expiresAt, user object)
-    // Note: user.id can be empty during the two-step auth flow bootstrap
     if (!token.sessionToken || !token.expiresAt || !token.user) {
       console.warn(
         `Warning: Auth token file has invalid format. Run 'tambo auth login' to re-authenticate.`,


packages/db/src/schema.ts
```diff
@@ -204,8 +209,29 @@ export const deviceAuthCodes = pgTable(
...
+      using: sql`true`,
+    }),
+    // Authenticated role can also manage device auth codes (for verify flow)
+    pgPolicy("device_auth_authenticated_all", {
+      to: authenticatedRole,
+      using: sql`true`,
+    }),
   ],
-);
+).enableRLS();
```
lachieh, line 235: As it stands, the `` using: sql`true` `` effectively grants public permissions here. @alecf, what do you think about adding a `device_auth` role (similar to the `projectApiKeyRole`) here:

https://github.com/tambo-ai/tambo/blob/7d8d8da1d389690eab7ee4952cdd7fc932058613/packages/db/src/schema.ts#L41-L44

Then we can have the `device-auth` route set that role and do the same thing as what the `transactionMiddleware` does here:

https://github.com/tambo-ai/tambo/blob/7d8d8da1d389690eab7ee4952cdd7fc932058613/apps/api/src/common/middleware/db-transaction-middleware.ts#L76-L120

Then we could adjust the policy to check `current_setting('request.deviceCode.code')` against the row and only grant access to the required row in the session table.

packages/db/migrations/0000_init_setup.sql
```diff
@@ -51,11 +60,21 @@ DO $$
...
+GRANT authenticated TO postgres;
+
+-- Grant schema usage to roles
+GRANT USAGE ON SCHEMA public TO anon, authenticated;
+GRANT USAGE ON SCHEMA auth TO authenticated;
+
+-- Grant auth schema access (read-only for user data) - only to authenticated role
+GRANT SELECT ON ALL TABLES IN SCHEMA auth TO authenticated;
+ALTER DEFAULT PRIVILEGES IN SCHEMA auth GRANT SELECT ON TABLES TO authenticated;
+
```
lachieh: I'm not seeing a reason that any of this is required. The `postgres` user is not guaranteed to exist and `set local role` already works here:
https://github.com/tambo-ai/tambo/blob/7d8d8da1d389690eab7ee4952cdd7fc932058613/apps/web/server/api/trpc.ts#L188

Granting SELECT on all tables is way too broad and the authenticated role can already do everything it needs to (or else existing features would not work).

lachieh (commented):
I have some questions that might change how we're thinking about these tables and roles 😬

packages/db/migrations/0000_init_setup.sql
akhileshrangani4: okay yes, this is indeed not required. removing this. my bad. done [here](https://github.com/tambo-ai/tambo/pull/1647/commits/8346f1b779a01d45c49ad85fbd2dceeeb2359a36)

packages/db/src/schema.ts
lachieh: We'll go with the current solution!

packages/db/migrations/0083_needy_puppet_master.sql
```diff
@@ -0,0 +1,8 @@
+CREATE ROLE "anon" WITH CREATEROLE NOINHERIT;--> statement-breakpoint
```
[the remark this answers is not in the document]
akhileshrangani4: @lachieh I think charlie's concern is right here, where creating an `anon` role would fail on prod supabase, to make it idempotent, I think I should add the example charlie suggested to the 0000 migration. what do you think? and ofc update schema.ts to remove `createRole: true`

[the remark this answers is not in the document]
lachieh: Yeah, agreed. Let's do it in an additional custom migration though.

See: https://orm.drizzle.team/docs/kit-custom-migrations

```sh
drizzle-kit generate --custom --name=add-anon-role
```

packages/db/migrations/0083_add-anon-role.sql
```diff
@@ -0,0 +1,8 @@
+CREATE ROLE "anon" WITH CREATEROLE NOINHERIT;--> statement-breakpoint
+GRANT SELECT, INSERT, UPDATE ON public.device_auth_codes TO "anon";--> statement-breakpoint
+GRANT SELECT ON public.sessions TO "anon";--> statement-breakpoint
```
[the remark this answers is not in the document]
lachieh: we can do this later if necessary

packages/db/src/schema.ts
```diff
@@ -204,8 +210,29 @@ export const deviceAuthCodes = pgTable(
...
```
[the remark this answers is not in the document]
lachieh: we can fix this later if we need to

packages/db/migrations/0083_needy_puppet_master.sql
```diff
@@ -0,0 +1,8 @@
...
+GRANT SELECT, INSERT, UPDATE ON public.device_auth_codes TO "anon";--> statement-breakpoint
+GRANT SELECT ON public.sessions TO "anon";--> statement-breakpoint
+ALTER TABLE "device_auth_codes" ENABLE ROW LEVEL SECURITY;--> statement-breakpoint
+CREATE POLICY "device_auth_anon_insert" ON "device_auth_codes" AS PERMISSIVE FOR INSERT TO "anon" WITH CHECK (true);--> statement-breakpoint
+CREATE POLICY "device_auth_anon_select" ON "device_auth_codes" AS PERMISSIVE FOR SELECT TO "anon" USING (true);--> statement-breakpoint
+CREATE POLICY "device_auth_anon_update" ON "device_auth_codes" AS PERMISSIVE FOR UPDATE TO "anon" USING (true);--> statement-breakpoint
+CREATE POLICY "device_auth_authenticated_all" ON "device_auth_codes" AS PERMISSIVE FOR ALL TO "authenticated" USING (true);
```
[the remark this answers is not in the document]
lachieh: this is fine

packages/db/src/schema.ts
```diff
@@ -204,8 +210,29 @@ export const deviceAuthCodes = pgTable(
...
+    pgPolicy("device_auth_anon_update", {
+      to: anonRole,
+      for: "update",
+      using: sql`true`,
+    }),
+    // Authenticated role can also manage device auth codes (for verify flow)
+    pgPolicy("device_auth_authenticated_all", {
+      to: authenticatedRole,
+      using: sql`true`,
+    }),
```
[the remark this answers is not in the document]
lachieh: Yeah, it's generated code, my guy.

packages/db/migrations/0083_needy_puppet_master.sql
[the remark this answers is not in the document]
akhileshrangani4: done [here](https://github.com/tambo-ai/tambo/pull/1647/commits/3d1dbd947f38b38b119e189004d727b0e09dd444)

packages/db/src/schema.ts
```diff
@@ -42,11 +42,11 @@ export const projectApiKeyRole = pgRole("project_api_key", {
...
-/** Anon role for unauthenticated requests (e.g., device auth initiation/polling) */
-export const anonRole = pgRole("anon", {
-  createRole: true,
-  inherit: false,
-});
+/** Anon role for unauthenticated requests (e.g., device auth initiation/polling)
+ * Note: Role is created via custom migration 0083_add-anon-role.sql.
+ * See that migration for details on why we use pgRole() with no options here.
+ */
+export const anonRole = pgRole("anon");
```
lachieh: Not sure it's necessary, but let's mark this as explicitly existing:
```suggestion
export const anonRole = pgRole("anon").existing();
```

apps/web/server/api/routers/device-auth.ts
```diff
@@ -278,35 +278,19 @@ export const deviceAuthRouter = createTRPCRouter({
...
           })
           .from(schema.sessions)
-          .innerJoin(
-            schema.authUsers,
-            eq(schema.sessions.userId, schema.authUsers.id),
-          )
           .where(eq(schema.sessions.id, deviceAuthCode.sessionId))
           .limit(1);
 
         return {
```
alecf, line 297: what guarantee do we have that `session` is valid? we should return/throw an error if `session` is nullish, and then use `session.expiresAt.toIsoString()` so you don't have to fall back to null

cli/src/lib/device-auth.ts
```diff
@@ -125,35 +125,64 @@ export async function runDeviceAuthFlow(): Promise<DeviceAuthResult> {
...
           throw new DeviceAuthError(
             "Authentication completed but no session token received",
           );
         }
 
-        // Step 4: Save token to disk
+        // Step 4a: Save token temporarily (allows authenticated requests)
         // Use server-provided expiry, fallback to 90 days if not provided
         const expiresAt =
           pollResponse.expiresAt ??
```
alecf: Under what circumstances would pollResponse.expiresAt be nullish? I think that if you make the above change then expiresAt shouldn't ever be nullish, so you won't need the fallback

cli/src/lib/device-auth.ts
```diff
@@ -125,35 +125,64 @@ export async function runDeviceAuthFlow(): Promise<DeviceAuthResult> {
...
           storedAt: new Date().toISOString(),
         };
 
-        await saveToken(tokenData);
-
-        return {
-          sessionToken: pollResponse.sessionToken,
-          user: pollResponse.user,
-        };
+        await saveToken(tempTokenData);
```
alecf: I think I've lost the reasoning for why the temp token needs to actually be saved to disk? I feel like if we can avoid saving the temp token data, but still reuse it later to get the user info, we should do that... avoid persisting a half-complete token?

(or if you've got a rational here, add a comment)

cli/src/lib/device-auth.ts
```diff
@@ -125,35 +125,64 @@ export async function runDeviceAuthFlow(): Promise<DeviceAuthResult> {
...
+          await saveToken(completeTokenData);
+
+          return {
+            sessionToken: pollResponse.sessionToken,
+            user: completeTokenData.user,
+          };
+        } catch (userError) {
+          userSpinner.fail(chalk.red("Failed to fetch user info"));
+          // Clean up incomplete token
+          clearToken();
```
alecf: minor but do you need to `await` this or is this synchronous?

cli/src/lib/token-storage.ts
```diff
@@ -123,7 +123,8 @@ export function loadToken(): StoredToken | null {
     const data = readFileSync(filePath, { encoding: "utf-8" });
     const token = JSON.parse(data) as StoredToken;
 
-    // Validate required fields
+    // Validate basic structure (sessionToken, expiresAt, user object)
+    // Note: user.id can be empty during the two-step auth flow bootstrap
```
alecf: but it seems like if user.id is empty, then you start the whole flow anyway, so no need to persist that

alecf (commented):
Just a bunch of little edge cases around error states - lets make sure that expiresAt can never be null - it should be reasonable for the whole app to crash if it gets invalid data like that

packages/db/src/schema.ts
akhileshrangani4: done! [here](https://github.com/tambo-ai/tambo/pull/1647/commits/61019c56c9c311189acddcffa77449a9c3861b6c)

apps/web/server/api/routers/device-auth.ts
akhileshrangani4: true, we do not have the guarantee of `session` being valid. I'll make the suggested changes.

akhileshrangani4: done [here](https://github.com/tambo-ai/tambo/pull/1647/commits/e3ec1b89feefbd28286f5c12c403190eb82e7614)

cli/src/lib/device-auth.ts
akhileshrangani4: yes, I did make the change above, so removed the fallback, [here](https://github.com/tambo-ai/tambo/pull/1647/commits/da12f8451cec755af8bda57b644cf46c4b62f8f1)

cli/src/lib/device-auth.ts
akhileshrangani4: actually, there was no good reason to save the incomplete token to disk. I've now updated it to use an in-memory token for the getUser request, and we only persist to disk once we have the complete token with user info. Added `setInMemoryToken()` to `token-storage.ts` for this.

akhileshrangani4: commit [here](https://github.com/tambo-ai/tambo/pull/1647/commits/5a30fad1d2d01dffa2eea605173a8291fe64ce39)

cli/src/lib/device-auth.ts
akhileshrangani4: I think this is now outdated since I'm using the in-memory token approach. If you meant clearToken() - that's removed now. If you meant saveToken - that's async but it's properly awaited.

cli/src/lib/token-storage.ts
akhileshrangani4: that comment is also now stale. With the in-memory token approach we only persist complete tokens now, so removed it.

PR #1651, opened 2026-01-06
ci(release-please): skip workflow for dependabot pushes

lachieh:
## Summary
- Skip the release-please workflow when triggered by dependabot pushes
- Dependabot-triggered push events don't have access to Actions secrets (only Dependabot secrets), causing the workflow to fail
- Dependency updates will be included in the next human-triggered release instead

## Test plan
- [ ] Verify workflow skips when a dependabot PR is merged
- [ ] Verify workflow runs normally for human-triggered pushes

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -15,6 +15,9 @@ permissions:
 
 jobs:
   release-please:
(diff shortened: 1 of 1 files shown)


lachieh:
This is actually a larger issue. I thought it was just dependabot breaking the release-please action but it's everything.

PR #1664, opened 2026-01-06
feat(react): replace image attachments with general file attachment system

michaelmagan:
## Summary

- Replace image-only attachments with a general-purpose attachment system supporting images, PDFs, text files, CSV, markdown, JSON, and more
- Add S3-compatible storage service with sensible defaults (Supabase for local dev, Tambo Cloud API for production)
- Align attachment handling with the broader resource model using `attachment://` URI scheme

Fixes https://github.com/tambo-ai/analytics-template/issues/15

## Why

The previous `useMessageImages` implementation only supported images, limiting what users could share with the AI. This PR introduces a flexible attachment system that:

1. **Supports multiple file types** - Images, PDFs, text files, CSV, markdown, JSON, code files
2. **Uses a consistent resource model** - Attachments use `attachment://` URIs that align with how Tambo handles resources generally
3. **Works out of the box** - Defaults to Tambo Cloud API storage, falls back to Supabase for local development

## Design Decisions

**Inlining content**: Attachments are fetched from storage and inlined (base64 for binary, text for text files) before sending to the LLM. This simplifies local development since there's no need for the LLM to have network access to fetch pre-signed URLs.

**Storage defaults**: Uses Tambo Cloud API storage by default. For local development, configurable to use Supabase or any S3-compatible provider via environment variables.

**Backwards compatibility**: `useMessageImages` and `StagedImage` are deprecated but still exported for compatibility until 1.0.0.

## Screenshots
Screenshot 2026-01-06 at 11 45 58 AM

Screenshot 2026-01-06 at 11 45 37 AM

Screenshot 2026-01-06 at 11 45 32 AM

## Test Plan

- [x] Upload image attachment → verify preview and submission
- [x] Upload PDF attachment → verify it's included in message
- [x] Upload text/CSV file → verify preview shows content summary
- [x] Test with Supabase storage locally
- [x] Verify backwards compatibility with `useMessageImages` import

### DIFF
--- a/react-sdk/src/hooks/use-message-attachments.test.ts
+++ b/react-sdk/src/hooks/use-message-attachments.test.ts
@@ -11,10 +11,20 @@ function createMockFile(
   type: string,
   content = "test content",
 ): File {
-  // Create content of the specified size
-  const actualContent = content.padEnd(size, "x").slice(0, size);
+  // For small files, create actual content; for large files (>1MB), use minimal content
+  const isLargeFile = size > 1024 * 1024;
+  const actualContent = isLargeFile
+    ? content
+    : content.padEnd(size, "x").slice(0, size);
   const blob = new Blob([actualContent], { type });
-  return new File([blob], name, { type });
+  const file = new File([blob], name, { type });
+
+  // For large files, mock the size property to avoid memory/timeout issues
+  if (isLargeFile) {
+    Object.defineProperty(file, "size", { value: size, writable: false });
+  }
+
+  return file;
 }
 
 describe("useMessageAttachments", () => {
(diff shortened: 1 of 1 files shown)


apps/api/src/threads/util/content.ts
```diff
@@ -68,9 +68,32 @@ export function convertContentDtoToContentPart(
             resource: part.resource,
           };
         }
-        default:
+        default: {
+          // Handle MCP resource_link type - convert to our Resource type
+          // We check at runtime because the type isn't in our ContentPartType enum
+          const partType = (part as { type: string }).type;
+          if (partType === "resource_link") {
```
alecf, line 75: actually it looks like resource_link is not even a thing, nobody else generates this?

alecf: oooh I see this could be coming from MCP? I don't want to solve this with casting, I want to make sure we have the right `type` in our types, so if something *could* come from MCP, then that should be expicit in this function

alecf (commented):
I think overall this PR is combining two things
1) better support for MCP multi-modal tool responses
2) adding the ability to upload images via an s3 service

I think we should untangle these two things - I just sent claude off to solve the first one

alecf:
@michaelmagan I did the MCP resource_link stuff in https://github.com/tambo-ai/tambo/pull/1666

apps/api/src/threads/util/content.ts
michaelmagan: Yup. This was just dealing with some MCP errors I was getting during testing.

michaelmagan:
@alecf cool. i'l branch off that PR to add the rest and close this PR.

Issue #1653, opened 2026-01-06
feat(cli): Detect and use project's package manager (pnpm/yarn/npm)


michaelmagan:
## Problem

CLI hardcodes `npm install` when adding components. Users with pnpm or yarn get errors.

**Reported error:**
```
npx tambo add message-thread-collapsible
npm error Cannot read properties of null (reading 'matches')
```

## Environment

- Node v23.11.0
- npm 10.9.2 (but user uses pnpm 9.15.4)
- Existing React app (non-Next.js)

## Proposed Solution

Detect lockfiles and use the appropriate package manager:
- `pnpm-lock.yaml` → pnpm
- `yarn.lock` → yarn
- `package-lock.json` → npm

lachieh:
Closed in #1684

PR #1667, opened 2026-01-06
fix(docs): missing `@returns`

nicolasmondain:
Ref: https://github.com/tambo-ai/tambo/issues/1663

PR #1684, opened 2026-01-07
feat(cli): Detect and use project's package manager (pnpm/yarn/npm)

nicolasmondain:
Ref: https://github.com/tambo-ai/tambo/issues/1653

### DIFF
--- a/cli/src/commands/create-app.ts
+++ b/cli/src/commands/create-app.ts
@@ -2,10 +2,15 @@ import chalk from "chalk";
 import fs from "fs";
 import ora from "ora";
 import path from "path";
-import { execSync, interactivePrompt } from "../utils/interactive.js";
+import {
+  execFileSync,
+  execSync,
+  interactivePrompt,
+} from "../utils/interactive.js";
 import {
   detectPackageManager,
   getInstallCommand,
+  validatePackageManager,
 } from "../utils/package-manager.js";
 
 // Define available templates
@@ -273,12 +278,13 @@ export async function handleCreateApp(
     }
 
     // Install dependencies with spinner
-    // Detect package manager from the target directory (which is now cwd)
-    const pm = detectPackageManager();
+    // Detect package manager from the target directory (which is now cwd after chdir above)
+    const pm = detectPackageManager(targetDir);
+    validatePackageManager(pm);
     const installCmd = getInstallCommand(pm);
     // --legacy-peer-deps is npm-specific
     const legacyPeerDepsFlag =
-      options.legacyPeerDeps && pm === "npm" ? " --legacy-peer-deps" : "";
+      options.legacyPeerDeps && pm === "npm" ? ["--legacy-peer-deps"] : [];
 
     const installSpinner = ora({
       text: `Installing dependencies using ${pm}...`,
@@ -286,7 +292,8 @@ export async function handleCreateApp(
     }).start();
 
     try {
-      execSync(`${pm} ${installCmd}${legacyPeerDepsFlag}`, { stdio: "ignore" });
+      const args = [installCmd, ...legacyPeerDepsFlag];
+      execFileSync(pm, args, { stdio: "ignore", allowNonInteractive: true });
       installSpinner.succeed("Dependencies installed successfully");
     } catch (_error) {
       installSpinner.fail("Failed to install dependencies");
--- a/cli/src/utils/package-manager.ts
+++ b/cli/src/utils/package-manager.ts
@@ -1,8 +1,10 @@
+import { execFileSync } from "child_process";
 import fs from "fs";
 import path from "path";
 
 /**
  * Supported package managers
+ * TODO: Add Bun support (uses bun.lockb) - currently falls back to npm
  */
 export type PackageManager = "npm" | "pnpm" | "yarn";
 
@@ -29,13 +31,30 @@ export function detectPackageManager(
 }
 
 /**
- * Gets the install command for a package manager
+ * Validates that the detected package manager is actually installed on the system.
+ * Throws an error with a helpful message if not installed.
+ *
+ * @param pm The package manager to validate
+ * @throws Error if the package manager is not installed
+ */
+export function validatePackageManager(pm: PackageManager): void {
+  try {
+    execFileSync(pm, ["--version"], { stdio: "ignore" });
+  } catch {
+    throw new Error(
+      `Detected ${pm} from lockfile but ${pm} is not installed. Please install ${pm} first.`,
+    );
+  }
+}
+
+/**
+ * Gets the install command for a package manager when installing specific packages
  *
  * @param pm The package manager
- * @returns The install command (e.g., "install" for npm/yarn, "add" for pnpm)
+ * @returns The install command (e.g., "install" for npm, "add" for pnpm/yarn)
  */
 export function getInstallCommand(pm: PackageManager): string {
-  return pm === "pnpm" ? "add" : "install";
+  return pm === "npm" ? "install" : "add";
 }
 
 /**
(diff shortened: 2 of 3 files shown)


akhileshrangani4:
hey @nicolasmondain thanks for attempting this! it seems like main is not merged into this branch. Can you please do that first so that it does not have changes of the last commit from #1667 - it would make it easier for me to review since this because of that this PR has a lot of out of scope changes.

nicolasmondain:
@akhileshrangani4 Sorry about that. Here is the refined PR 🚀

akhileshrangani4 (commented):
(Superseded by comprehensive review below)

akhileshrangani4 (commented):
(Superseded by comprehensive review below)

cli/src/utils/package-manager.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/**
+ * Gets the install command for a package manager
+ *
+ * @param pm The package manager
+ * @returns The install command (e.g., "install" for npm/yarn, "add" for pnpm)
+ */
+export function getInstallCommand(pm: PackageManager): string {
+  return pm === "pnpm" ? "add" : "install";
```
akhileshrangani4: This breaks Yarn completely. Yarn does not support `yarn install <package>` - it only supports `yarn add <package>`.

Current code:
```typescript
return pm === "pnpm" ? "add" : "install";
```

Should be:
```typescript
return pm === "npm" ? "install" : "add";
```

This way:
- npm uses `npm install` 
- pnpm uses `pnpm add`
- yarn uses `yarn add`

Right now any Yarn user will get "error No package name specified" when running the CLI.

See official docs: https://classic.yarnpkg.com/lang/en/docs/cli/add/

cli/src/commands/create-app.ts
```diff
@@ -269,21 +273,25 @@ export async function handleCreateApp(
...
+      text: `Installing dependencies using ${pm}...`,
       spinner: "dots",
     }).start();
 
     try {
-      execSync(
-        `npm install${options.legacyPeerDeps ? " --legacy-peer-deps" : ""}`,
-        { stdio: "ignore" },
-      );
+      execSync(`${pm} ${installCmd}${legacyPeerDepsFlag}`, { stdio: "ignore" });
```
akhileshrangani4: This line uses execSync with string interpolation instead of execFileSync like everywhere else in the PR. This is a shell injection risk and also cannot use the allowNonInteractive flag.

Change this:
```typescript
execSync(`${pm} ${installCmd}${legacyPeerDepsFlag}`, { stdio: "ignore" });
```

To this:
```typescript
const args = [installCmd, ...legacyPeerDepsFlag];
execFileSync(pm, args, { stdio: "ignore", allowNonInteractive });
```

This matches the pattern used in component.ts and npm-packages.ts.

cli/src/commands/create-app.ts
```diff
@@ -269,21 +273,25 @@ export async function handleCreateApp(
     }
 
     // Install dependencies with spinner
+    // Detect package manager from the target directory (which is now cwd)
+    const pm = detectPackageManager();
+    const installCmd = getInstallCommand(pm);
+    // --legacy-peer-deps is npm-specific
+    const legacyPeerDepsFlag =
```
akhileshrangani4, line 286: This should return an array like everywhere else in the PR, not a string.

Current:
```typescript
const legacyPeerDepsFlag = options.legacyPeerDeps && pm === "npm" ? " --legacy-peer-deps" : "";
```

Should be:
```typescript
const legacyPeerDepsFlag = options.legacyPeerDeps && pm === "npm" ? ["--legacy-peer-deps"] : [];
```

The string version only works with execSync (which you should change anyway), but breaks with execFileSync.

cli/src/commands/create-app.ts
```diff
@@ -269,21 +273,25 @@ export async function handleCreateApp(
     }
 
     // Install dependencies with spinner
+    // Detect package manager from the target directory (which is now cwd)
+    const pm = detectPackageManager();
```
akhileshrangani4: Small risk here - detectPackageManager() defaults to process.cwd(), which works because you did process.chdir(targetDir) above. But if someone refactors this later and moves the detection call, it will break.

Consider being explicit:
```typescript
const pm = detectPackageManager(targetDir);
```

Or add a comment:
```typescript
// MUST be called after process.chdir(targetDir)
const pm = detectPackageManager();
```

cli/src/utils/package-manager.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Detects the package manager used in a project by checking for lockfiles
+ * Priority order: pnpm-lock.yaml > yarn.lock > package-lock.json > npm (default)
+ *
+ * @param projectRoot The root directory of the project (defaults to cwd)
+ * @returns The detected package manager
+ */
+export function detectPackageManager(
+  projectRoot: string = process.cwd(),
+): PackageManager {
+  if (fs.existsSync(path.join(projectRoot, "pnpm-lock.yaml"))) {
```
akhileshrangani4, line 21: Would be good to validate the detected package manager is actually installed before trying to use it.

Right now if someone has a pnpm-lock.yaml but does not have pnpm installed globally, they will get a cryptic "command not found" error.

Suggestion - add a helper:
```typescript
function validatePackageManager(pm: PackageManager): void {
  try {
    execFileSync(pm, ['--version'], { stdio: 'ignore' });
  } catch {
    throw new Error(`Detected ${pm} from lockfile but ${pm} is not installed. Install it first.`);
  }
}
```

Then call it after detection in each command.

cli/src/utils/package-manager.ts
```diff
@@ -0,0 +1,83 @@
...
+  return pm === "yarn" ? "--dev" : "-D";
+}
+
+/**
+ * Gets the package runner command (npx equivalent) for a package manager
+ *
+ * @param pm The package manager
+ * @returns The runner command (e.g., "npx", "pnpm dlx", "yarn dlx")
+ */
+export function getPackageRunner(pm: PackageManager): string {
```
akhileshrangani4: getPackageRunner() returns a string like "pnpm dlx" but it is never used anywhere - only getPackageRunnerArgs() is used.

Either remove it or use it if needed somewhere.

cli/src/utils/package-manager.ts
akhileshrangani4, line 21: Not a blocker but FYI - Bun is getting popular and uses bun.lockb. Right now Bun users will fall back to npm which might work but not ideal.

Maybe add a TODO comment or file a follow-up issue?

akhileshrangani4 (commented):
Overall good, still a few things:

akhileshrangani4 (changes requested):
The main issues that need fixing:

1. Yarn command is broken - uses `yarn install` instead of `yarn add`
2. Inconsistent use of execSync vs execFileSync in create-app.ts
3. String flag instead of array for legacyPeerDepsFlag in create-app.ts

These need to be fixed before merge.

michaelmagan:
@nicolasmondain this is on our roadmap so we are hoping to get this closed this week. 

any chance you can fix @akhileshrangani4 revisions?

nicolasmondain:
> @nicolasmondain this is on our roadmap so we are hoping to get this closed this week.
> 
> any chance you can fix @akhileshrangani4 revisions?

@michaelmagan I'll push some updates according to @akhileshrangani4 feedbacks EOD (Paris time) 🫡

nicolasmondain:
@michaelmagan just pushed my pending changes actually: https://github.com/tambo-ai/tambo/pull/1684/changes/ac2a004829e6866302f9e5f412469a11e0351ff7

akhileshrangani4 (approved):
Looks good! Thank you for your contribution!

PR #1728, opened 2026-01-11
fix(cli): add GenerationStage enum to components and Jest mock

### LINKED ISSUE #1704: Replace magic strings with GenerationStage enum
## Summary

Generation stage checks use magic strings instead of a typed enum. This makes refactoring error-prone and reduces type safety.

## Files

- `cli/src/registry/message-suggestions/message-suggestions.tsx`
- `cli/src/registry/thread-content/thread-content.tsx`

## Task

Create a GenerationStage enum and replace magic string comparisons with enum values throughout the codebase.

## Verification

- [ ] Tests added/updated to cover new functionality
- [ ] `npm run check-types` passes
- [ ] `npm run lint` passes
- [ ] `npm test` passes
- [ ] Generation stage logic works identically

See [CONTRIBUTING.md](https://github.com/tambo-ai/tambo/blob/main/CONTRIBUTING.md) for full contribution guidelines.

mizurex:
Closes #1704 

Replaces magic strings with GenerationStage enum values in components and tests.

### DIFF
--- a/cli/__tests__/__mocks__/@tambo-ai-react.ts
+++ b/cli/__tests__/__mocks__/@tambo-ai-react.ts
@@ -6,10 +6,21 @@
 import { jest } from "@jest/globals";
 import type { Mock } from "jest-mock";
 
+export enum GenerationStage {
+  IDLE = "IDLE",
+  CHOOSING_COMPONENT = "CHOOSING_COMPONENT",
+  FETCHING_CONTEXT = "FETCHING_CONTEXT",
+  HYDRATING_COMPONENT = "HYDRATING_COMPONENT",
+  STREAMING_RESPONSE = "STREAMING_RESPONSE",
+  COMPLETE = "COMPLETE",
+  ERROR = "ERROR",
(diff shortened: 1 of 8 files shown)


lachieh:
Thank you, @mizurex!

alecf:
@CharlieHelps this PR created a "mock" enum, where this exact enum is available as an import elsewhere, we shouldn't be creating a mock enum. Can you create a PR to fix this so the mock just brings in the real value?

mizurex:
@alecf I intended to create the mock enum and not export it, but I copy pasted it and didn’t realize I was exporting it. can I create a PR to revert it back to how it was? Sorry for the ping, thanks.

alecf:
@mizurex the problem isn't the exporting - the problem is that we shouldn't create a mock enum ever. There's no value in a "mock" version of static data, especially since the mock enum can then drift from the real enum

(I was hoping our bot could take care of this but for various reasons he's not responding...) - if you wouldn't mind a followup PR that would be great, thanks!

Issue #1705, opened 2026-01-08
Extract paste handling into reusable utility


michaelmagan:
## Summary

Paste handling logic is duplicated between message-input and text-editor components. This should be extracted into a shared utility.

## Files

- `cli/src/registry/message-input/message-input.tsx`
- `cli/src/registry/message-input/text-editor.tsx`

## Task

Extract common paste handling logic into a reusable utility function and update both components to use it.

## Verification

- [ ] Tests added/updated to cover new functionality
- [ ] `npm run check-types` passes
- [ ] `npm run lint` passes
- [ ] `npm test` passes
- [ ] Paste functionality works in both components

See [CONTRIBUTING.md](https://github.com/tambo-ai/tambo/blob/main/CONTRIBUTING.md) for full contribution guidelines.

lachieh:
The underlying implementation of this particular functionality has some pretty bad code smells. So I'm gonna log a new issue that describes a refactor to a hook for this functionality.

PR #1732, opened 2026-01-12
fix(message-input): clean up paste handling

mizurex:
solves - #1705 

Added getImageItems helper used by both text-editor.tsx and message-input.tsx.


michaelmagan:
@mizurex thanks for contributing. Could you add test's please?

lachieh (approved):
Thanks, @mizurex! You're on a roll!

PR #1734, opened 2026-01-12
feat(cli): add framework detection for automatic env var prefix selection

akhileshrangani4:
## Summary

- Added framework detection utility that automatically detects the project framework (currently Next.js) and returns the appropriate environment variable prefix
- Refactored `tambo init` command to use framework detection instead of hardcoding `NEXT_PUBLIC_` prefix
- Updated API key handling to support multiple env var formats and migrate existing keys to detected framework's prefix
- fixes TAM-862

## Why

Previously, the `tambo init` command always used `NEXT_PUBLIC_TAMBO_API_KEY`, which only works for Next.js projects. Users of other frameworks had to manually remove the prefix. This change:

- Automatically detects the framework and uses the correct prefix (e.g., `NEXT_PUBLIC_` for Next.js, `VITE_` for Vite when added)
- Improves UX by removing manual steps
- Sets up infrastructure for supporting additional frameworks (Vite, CRA, etc.) in the future
- Maintains backward compatibility by detecting and migrating existing API keys

## Test Plan

1. Run framework detection tests: `npm test -w cli -- framework-detection.test.ts`
2. Test Next.js detection: Run `tambo init` in a Next.js project and verify it uses `NEXT_PUBLIC_TAMBO_API_KEY`
3. Test non-framework detection: Run `tambo init` in a non-Next.js project and verify it uses `TAMBO_API_KEY` (no prefix)
4. Test existing key migration: Run `tambo init` with an existing `NEXT_PUBLIC_TAMBO_API_KEY` and verify it's detected and can be overwritten
5. Verify CLI linting and type checking: `npm run lint -w cli && npm run check-types -w cli`

### DIFF
--- a/cli/src/commands/init.ts
+++ b/cli/src/commands/init.ts
@@ -18,6 +18,12 @@ import {
   NonInteractiveError,
 } from "../utils/interactive.js";
 import {
+  findAllTamboApiKeys,
+  findTamboApiKey,
+  setTamboApiKey,
+  type TamboApiKeyName,
+} from "../utils/dotenv-utils.js";
+import {
   detectFramework,
   getTamboApiKeyEnvVar,
 } from "../utils/framework-detection.js";
@@ -78,17 +84,18 @@ async function writeApiKeyToEnv(apiKey: string): Promise<boolean> {
 
     // Detect framework and get appropriate env var name
     const framework = detectFramework();
-    const envVarName = getTamboApiKeyEnvVar();
-    const envContent = `\n${envVarName}=${apiKey.trim()}\n`;
+    const envVarName = getTamboApiKeyEnvVar() as TamboApiKeyName;
 
     if (framework) {
       console.log(chalk.gray(`\nDetected ${framework.displayName} project`));
     }
 
     if (!fs.existsSync(targetEnvFile)) {
-      fs.writeFileSync(targetEnvFile, "# Environment Variables\n");
+      fs.writeFileSync(
+        targetEnvFile,
+        `# Environment Variables\n${envVarName}=${apiKey.trim()}\n`,
+      );
       console.log(chalk.green(`\n✔ Created new ${targetEnvFile} file`));
-      fs.appendFileSync(targetEnvFile, envContent);
       console.log(
         chalk.green(`\n✔ API key saved to ${targetEnvFile} as ${envVarName}`),
       );
@@ -96,24 +103,14 @@ async function writeApiKeyToEnv(apiKey: string): Promise<boolean> {
     }
 
     const existingContent = fs.readFileSync(targetEnvFile, "utf8");
-    // Match any Tambo API key variant (with or without framework prefix)
-    const keyRegex = /^(NEXT_PUBLIC_|VITE_|REACT_APP_)?TAMBO_API_KEY=.*/gm;
-    const existingMatches = existingContent.match(keyRegex);
+    const existingKeyNames = findAllTamboApiKeys(existingContent);
 
-    if (existingMatches && existingMatches.length > 0) {
-      // Extract key names from matches
-      const existingKeyNames = existingMatches.map(
-        (match) => match.split("=")[0],
-      );
-
+    if (existingKeyNames.length > 0) {
       // Build appropriate warning message
-      let warningMessage: string;
-      if (existingKeyNames.length === 1) {
-        warningMessage = `⚠️  This will overwrite the existing value of ${existingKeyNames[0]} in ${targetEnvFile}, are you sure?`;
-      } else {
-        // Multiple keys found - warn about all of them
-        warningMessage = `⚠️  Found multiple Tambo API key variants in ${targetEnvFile}:\n   ${existingKeyNames.join(", ")}\n   This will remove all of them and replace with ${envVarName}. Continue?`;
-      }
+      const warningMessage =
+        existingKeyNames.length === 1
+          ? `⚠️  This will overwrite the existing value of ${existingKeyNames[0]} in ${targetEnvFile}, are you sure?`
+          : `⚠️  Found multiple Tambo API key variants in ${targetEnvFile}:\n   ${existingKeyNames.join(", ")}\n   This will remove all of them and replace with ${envVarName}. Continue?`;
 
       const { confirmReplace } = await interactivePrompt<{
         confirmReplace: boolean;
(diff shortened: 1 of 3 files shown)


cli/src/__fixtures__/mock-fs-setup.ts
```diff
@@ -200,53 +201,91 @@ export function createBasicProject(): Record<string, string | null> {
...
+/**
+ * Creates a basic Next.js project structure
+ * Use this when testing Next.js-specific behavior (e.g., env var prefix detection)
+ */
+export function createNextProject(): Record<string, string | null> {
+  return {
+    "/mock-project/package.json": JSON.stringify({
+      name: "test-project",
+      dependencies: {
+        next: "^14.0.0",
```
alecf: we should start fresh wtih next 16

cli/src/__fixtures__/mock-fs-setup.ts
```diff
@@ -200,53 +201,91 @@ export function createBasicProject(): Record<string, string | null> {
...
+    }),
+    "/mock-project/src": null,
+  };
+}
+
+/**
+ * Creates a Next.js project with React dependency and src directory
+ * Use this when testing Next.js-specific behavior (e.g., env var prefix detection)
+ */
+export function createNextProjectWithReact(): Record<string, string | null> {
```
alecf: very minor but these functions "WithReact" should really be with something like "WithTamboSDK"

cli/src/commands/init.ts
```diff
@@ -71,38 +76,56 @@ async function writeApiKeyToEnv(apiKey: string): Promise<boolean> {
...
-        ),
+        chalk.green(`\n✔ API key saved to ${targetEnvFile} as ${envVarName}`),
       );
       return true;
     }
 
     const existingContent = fs.readFileSync(targetEnvFile, "utf8");
-    const keyRegex = /^NEXT_PUBLIC_TAMBO_API_KEY=.*/gm;
+    // Match any Tambo API key variant (with or without framework prefix)
+    const keyRegex = /^(NEXT_PUBLIC_|VITE_|REACT_APP_)?TAMBO_API_KEY=.*/gm;
```
alecf: In general we need to avoid multi-line, global regexes. I'll add this to CLAUDE.md - their behavior is really complicated, and this doesn't even address any of the standard `dotenv` formatting like comments, quotes and multi-line formatting

Please use `dotenv` direct and do not try to get at this with string manipulation.

cli/src/commands/init.ts
```diff
@@ -111,28 +134,41 @@ async function writeApiKeyToEnv(apiKey: string): Promise<boolean> {
...
+        if (!replacedFirst) {
+          replacedFirst = true;
+          return `${envVarName}=${apiKey.trim()}`;
+        }
+        // Remove subsequent matches (return empty string, will leave blank line)
+        return "";
+      });
+
+      // Clean up any double blank lines left from removals
+      const cleanedContent = updatedContent.replace(/\n{3,}/g, "\n\n");
```
alecf: again this can be done without regex, but also if you find yourself cleaning up you might wonder if there was a way to get it right the first time? I hope that `dotenv` will clear some of this up

(though...just now realizing that dotenv probably throws out comments and whatnot... hrm...)

cli/src/commands/init.ts
```diff
@@ -111,28 +134,41 @@ async function writeApiKeyToEnv(apiKey: string): Promise<boolean> {
...
-      fs.writeFileSync(targetEnvFile, updatedContent);
-      console.log(
-        chalk.green(`\n✔ Updated existing API key in ${targetEnvFile}`),
-      );
-      console.log(
-        chalk.gray(
-          "\nNote: If you're not using Next.js, remove 'NEXT_PUBLIC_' from the variable name in your env file",
-        ),
-      );
+      // Remove all existing key variants and add the new one
```
alecf: I'm starting to think this is all just too smart for our own good - like maybe just remove them and add new ones at the end of the file?

cli/src/commands/init.ts
alecf: I'm starting to think maybe it would make sense to break all the dotenv stuff out into a separate file. There is an ancient library called `edit-dotenv` - I wonder if we could mimic some of this... then at least you could have a separate set of explicit tests for just replacing a variable in a string buffer and not have it all tangled up in this file

cli/src/commands/init.ts
```diff
@@ -183,24 +219,28 @@ function displaySelfHostInstructions(): void {
 
 /**
  * Checks for existing API key in .env files
+ * Supports multiple env var formats (with or without framework prefixes)
  * @returns string | null Returns existing API key if found, null otherwise
  */
 async function checkExistingApiKey(): Promise<string | null> {
   const envFiles = [".env.local", ".env"];
+  // Match any Tambo API key variant (with or without framework prefix)
+  const keyRegex = /^(NEXT_PUBLIC_|VITE_|REACT_APP_)?TAMBO_API_KEY=(.+)$/m;
```
alecf: and now we have this same regex used elsewhere...

cli/src/__fixtures__/mock-fs-setup.ts
akhileshrangani4: yup, you are right, will fix that.

cli/src/__fixtures__/mock-fs-setup.ts
akhileshrangani4: yes, done.

cli/src/commands/init.ts
akhileshrangani4: Agreed, replaced with `dotenv.parse()` for reading values. Created a new `src/utils/dotenv-utils.ts` utility that uses the `dotenv` parser for correct handling of quotes, escapes, comments, etc.

cli/src/commands/init.ts
akhileshrangani4: Fixed this - the new approach filters out old key lines and appends the new key at the end, so there's nothing to clean up.

cli/src/commands/init.ts
akhileshrangani4: Yep, that's exactly what `setTamboApiKey()` does now - filters out any existing Tambo key lines and appends the new one at the end.

cli/src/commands/init.ts
akhileshrangani4: Also created `src/utils/dotenv-utils.ts` with `findTamboApiKey(), findAllTamboApiKeys(), setTamboApiKey(),` etc. Also added `dotenv-utils.test.ts`.

cli/src/commands/init.ts
akhileshrangani4: Both usages now go through the shared utility functions. Key names are defined once in `TAMBO_API_KEY_NAMES` constant.

alecf (commented):
nice this is much cleaner

PR #1766, opened 2026-01-14
feat(docs): Implement middleware to serve markdown when `Accept` prefers

### LINKED ISSUE #1727: feat(docs): serve Markdown when Accept prefers it
### Context
The goal from the referenced post is *content negotiation* for LLM agents: keep the normal HTML experience for browsers, but return Markdown/plain text from the *same URL* when the request’s `Accept` header prefers `text/markdown` or `text/plain` over `text/html`.

Reference: https://www.skeptrune.com/posts/use-the-accept-header-to-serve-markdown-instead-of-html-to-llms/

### What we already have
The docs app already has a Markdown representation for every docs page:
- `/:path*.mdx` rewrites to `/llms.mdx/:path*` (see `docs/next.config.mjs`)
- `/llms.mdx/:path*` returns the page’s `getLLMText(page)` output (`docs/src/app/llms.mdx/[[...slug]]/route.ts`)

So the missing piece is: when an agent requests `/some/doc/page` with an `Accept` header that prefers Markdown/plain, we should internally serve the existing Markdown representation.

### Scope
Implement Accept-header based Markdown serving in the docs app (small, self-contained change; no new deps):

1. Add `docs/src/middleware.ts` that:
   - Runs only on `GET` (and maybe `HEAD`).
   - Skips non-doc routes (at minimum: `/_next`, `/api/*`, `/llms.txt`, `/llms-full.txt`, `/llms.mdx/*`, `/robots.txt`, plus obvious static assets like `*.ico`, `*.png`, etc.).
   - Reads `request.headers.get('accept')` and decides whether Markdown is preferred.
   - If Markdown is preferred, rewrites internally to `/llms.mdx${pathname}` (preserving the original path).

2. “Preferred Markdown” definition (match the blog post intent):
   - If `text/markdown` is present and occurs *before* `text/html`, prefer Markdown.
   - Else if `text/plain` is present and occurs *before* `text/html`, prefer Markdown.
   - If `text/html` is absent, treat `text/markdown` or `text/plain` as preferred.

   (A simple index-based check is fine; we don’t need full RFC parsing for this first pass.)

Optional nice-to-have (only if it’s trivial): set `Content-Type: text/markdown; charset=UTF-8` for `/llms.mdx/*` responses.

### Acceptance criteria
- Visiting docs in a browser is unchanged (normal HTML).
- `curl -H 'Accept: text/markdown' http://localhost:8263/getting-started/quickstart` returns Markdown/plain text (not HTML).
- `curl -H 'Accept: text/html, text/markdown' http://localhost:8263/getting-started/quickstart` returns HTML.
- `/api/search` and `/llms.txt` are not affected by the middleware.

### Verification
From repo root:
```bash
npm run lint -- --filter=@tambo-ai/docs
npm run check-types -- --filter=@tambo-ai/docs
npm run build -- --filter=@tambo-ai/docs
```

Manual spot-check:
```bash
npm run dev:docs
curl -i -H 'Accept: text/markdown' http://localhost:8263/getting-started/quickstart
```

BeeBombshell:
Fixes #1727

### Changes
- [x] implemented `middleware.ts` to:
  - only process `GET` or `HEAD` requests
  - filter routes ([check list](https://github.com/tambo-ai/tambo/compare/main...BeeBombshell:1727-serve-markdown?expand=1#diff-eb2f3506883e28e46187f167e68d517e81ae9c279234cdbcbc9b48e9536faa45R40-R45) + static assets)
  - parses `Accept` header to determine if markdown is preferred ([index-based comparison](https://github.com/tambo-ai/tambo/compare/main...BeeBombshell:1727-serve-markdown?expand=1#diff-eb2f3506883e28e46187f167e68d517e81ae9c279234cdbcbc9b48e9536faa45R10) - defaults to `text/html`)
  - if markdown, rewrites internally to `/llms.mdx{pathname}` (instead of redirecting)

- [x] also implemented matcher config to skip middleware execution for static files
- [x] set `Content-Type: text/markdown; charset=UTF-8` for `/llms.mdx/*` responses

### Verification
Accd. to the [acceptance criteria](https://github.com/tambo-ai/tambo/issues/1727#issue-3800503065) mentioned:
1. > Visiting docs in a browser is unchanged (normal HTML)
image

2. > `curl -H 'Accept: text/markdown' http://localhost:8263/getting-started/quickstart` returns Markdown/plain text (not HTML).
3. > `curl -H 'Accept: text/html, text/markdown' http://localhost:8263/getting-started/quickstart` returns HTML.
image

4. > `/api/search` and `/llms.txt` are not affected by the middleware.
image

### DIFF
--- a/docs/src/middleware.ts
+++ b/docs/src/middleware.ts
@@ -1,31 +1,69 @@
 import { NextRequest, NextResponse } from "next/server";
 
 /**
- * Determines if the Accept header prefers Markdown/plain text over HTML.
+ * Parses the Accept header and extracts content types with their q-values.
+ * Returns an Array<{type, q}> sorted by quality (descending).
+ */
+function parseAcceptHeader(
+  acceptHeader: string,
+): Array<{ type: string; q: number }> {
+  return (
+    acceptHeader
+      .split(",")
+      .map((part) => {
+        const [type, ...params] = part.split(";").map((s) => s.trim());
+
+        // Default q-value to 1.0 if not present
+        let q = 1.0;
+        for (const param of params) {
+          // Extract q-values from parameters like "q=0.9"
+          const match = param.match(/^q\s*=\s*([0-9.]+)$/i);
+          if (match) {
+            q = parseFloat(match[1]);
+            break;
+          }
+        }
+
+        return { type: type.toLowerCase(), q };
+      })
+      // Sort by q-value (descending)
+      .sort((a, b) => b.q - a.q)
+  );
+}
+
+/**
+ * Determines if the Accept header prefers Markdown/plain text over HTML (sorted by q-values)
  * Returns true if:
- * - `text/markdown` appears before `text/html`
- * - `text/plain` appears before `text/html`
+ * - `text/markdown` has higher q-value than `text/html`
+ * - `text/plain` has higher q-value than `text/html`
  * - `text/html` is absent and `text/markdown` or `text/plain` is present
  */
 function prefersMarkdown(acceptHeader: string | null): boolean {
   if (!acceptHeader) return false;
 
-  const accept = acceptHeader.toLowerCase();
-  const htmlIndex = accept.indexOf("text/html");
-  const markdownIndex = accept.indexOf("text/markdown");
-  const plainIndex = accept.indexOf("text/plain");
+  const sortedEntries = parseAcceptHeader(acceptHeader);
 
-  // If `text/html` is not present
+  // Find indices in the sorted entries
+  const htmlIndex = sortedEntries.findIndex(
+    (entry) => entry.type === "text/html",
+  );
+  const markdownIndex = sortedEntries.findIndex(
+    (entry) => entry.type === "text/markdown",
+  );
+  const plainIndex = sortedEntries.findIndex(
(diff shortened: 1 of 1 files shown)


BeeBombshell:
Hey @michaelmagan, this is ready for review! Added some screenshots from local testing above. Happy to adjust anything as needed! :)

lachieh:
@BeeBombshell this is awesome! Thanks for getting into this. If you could take a look at what Charlie suggests above, that'd be great! For point 1, it might be as simple as confirming that the redirects work as expected. For point 2, I'd just sort the values via the [q-value](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values) (assume no `q=` is equivalient to `q=0`) before finding their index.

BeeBombshell:
Hi @lachieh, thanks for the guidance on this! :D
- For 1, I've added filtering for the base `/llms.txt` path, as well as the paths ending in `.mdx` (which are already handled by `next.config.mjs` rewrites). Also confirmed the `.mdx` redirects work as expected locally.
- For 2, I've updated the logic to extract content types from the Accept header, sort them by priority, and then use them to determine markdown preference

PR #1767, opened 2026-01-14
feat(api): implement v1 API phase 1 - foundation & schema

alecf:
## Summary

- Add `V1RunStatus` enum to core package with states: `idle`, `waiting`, `streaming`
- Add v1 fields to `threads` table for run lifecycle, outcome tracking, and next-run requirements
- Create v1 module structure with discriminated union DTOs for content blocks
- Add RFC 9457 Problem Details error helper
- Modify `ThreadInProjectGuard` to support both `:id` and `:thread_id` route params

## Details

This is Phase 1 of 4 for implementing the v1 API as defined in `plans/api-v1-proposal.md`.

### Database Schema

New columns added to `threads` table, organized by concern:

**Current Run Lifecycle:**
| Column | Type | Default | Description |
|--------|------|---------|-------------|
| `run_status` | text | `'idle'` | Streaming status (idle/waiting/streaming) |
| `current_run_id` | text | null | ID of active run |
| `status_message` | text | null | Human-readable status |

**Last Run Outcome:**
| Column | Type | Default | Description |
|--------|------|---------|-------------|
| `last_run_cancelled` | boolean | null | Whether last run was cancelled |
| `last_run_error` | jsonb | null | Error info from last run (code, message) |

**Next Run Requirements:**
| Column | Type | Default | Description |
|--------|------|---------|-------------|
| `pending_tool_call_ids` | jsonb | null | Tool calls awaiting client results |
| `last_completed_run_id` | text | null | ID of last completed run |

### New Files

```
apps/api/src/v1/
├── v1.module.ts              # Module registration
├── v1.errors.ts              # RFC 9457 Problem Details helper
└── dto/
    ├── content.dto.ts        # Discriminated unions (text, resource, tool_use, tool_result, component)
    ├── message.dto.ts        # Message input/output DTOs
    ├── thread.dto.ts         # Thread CRUD DTOs
    ├── run.dto.ts            # Run creation DTOs
    └── tool.dto.ts           # Tool and component definitions
```

### Naming Conventions

**TypeScript class names** have a `V1` prefix for code clarity (distinguishes from legacy types):
- `V1TextContentDto`, `V1MessageDto`, `V1ThreadDto`, `V1CreateRunDto`, etc.

**API schema names** (exposed in OpenAPI/Swagger) are clean without `V1` prefix where possible:
- `TextContent`, `Message`, `Tool`, `CreateRunRequest`, etc.

**Exceptions** (collision with existing schemas): These keep the `V1` prefix in the API:
- `V1Thread` (collides with existing `Thread`)
- `V1ThreadWithMessages` (collides with existing `ThreadWithMessages`)
- `V1AvailableComponent` (collides with existing `AvailableComponent`)

### Key Implementation Details

- **V1RunStatus**: Simple 3-state enum aligned with AG-UI's run model. Edge cases (cancellation, errors, pending tool calls) tracked via separate Thread fields, not status values.
- **Discriminated unions**: Use `@Equals()` with definite assignment (`!:`) instead of default values to ensure proper validation
- **Guard flexibility**: `ThreadInProjectGuard` now checks both `thread_id` (v1) and `id` (legacy) params
- **Module registered**: V1Module added to `app.module.ts` imports
- **Reused DTOs**: `ResourceDto` reused from existing threads DTOs to avoid duplication

## Test plan

- [x] Lint passes
- [x] Type-check passes (all 13 packages)
- [x] Tests pass (all packages)
- [ ] Migration applies cleanly in dev environment
- [ ] V1 module loads without errors

## Next Steps (Phase 2-4)

- Phase 2: Thread & Message CRUD endpoints
- Phase 3: Run Streaming & Tools
- Phase 4: Component State & Integration Tests

### DIFF
--- a/packages/db/migrations/0086_magical_cannonball.sql
+++ b/packages/db/migrations/0086_magical_cannonball.sql
@@ -21,4 +21,5 @@ CREATE TABLE "runs" (
 ALTER TABLE "runs" ADD CONSTRAINT "runs_thread_id_threads_id_fk" FOREIGN KEY ("thread_id") REFERENCES "public"."threads"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
 CREATE INDEX "runs_thread_id_idx" ON "runs" USING btree ("thread_id");--> statement-breakpoint
 CREATE INDEX "runs_status_idx" ON "runs" USING btree ("status");--> statement-breakpoint
-CREATE INDEX "runs_created_at_idx" ON "runs" USING btree ("created_at");
+CREATE INDEX "runs_created_at_idx" ON "runs" USING btree ("created_at");--> statement-breakpoint
+ALTER TABLE "threads" ADD CONSTRAINT "threads_current_run_id_fk" FOREIGN KEY ("current_run_id") REFERENCES "public"."runs"("id") ON DELETE set null ON UPDATE no action;
--- a/packages/db/src/schema.ts
+++ b/packages/db/src/schema.ts
@@ -602,6 +602,8 @@ export const threads = pgTable(
       index("threads_project_updated_idx").on(table.projectId, table.updatedAt),
       // Stand-alone index on updated_at to aid generic recency queries.
       index("threads_updated_at_idx").on(table.updatedAt),
+      // Note: threads.current_run_id -> runs.id FK is added via migration SQL
+      // to avoid circular type inference issues between threads and runs tables
     ];
   },
 );


apps/api/src/v1/dto/thread.dto.ts
```diff
@@ -0,0 +1,224 @@
...
+  @IsString()
+  createdAt!: string;
+
+  @ApiProperty({
+    description: "When the thread was last updated (ISO 8601)",
+    example: "2024-01-15T12:05:00Z",
+  })
+  @IsString()
+  updatedAt!: string;
+}
```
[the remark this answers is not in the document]
MichaelMilstead: Confused by this comment, I'm not seeing where `processedToolCallIds` is added to any schema or interface?

apps/api/src/v1/dto/run.dto.ts
```diff
@@ -0,0 +1,158 @@
...
+  V1AvailableComponentDto,
+  V1ToolChoiceDto,
+} from "./tool.dto";
+import { V1CreateThreadDto } from "./thread.dto";
+
+/**
+ * Request DTO for creating a run on an existing thread.
+ */
+@ApiSchema({ name: "CreateRunRequest" })
+export class V1CreateRunDto {
```
MichaelMilstead, line 26: seems like the idea for handling "additional context" (of various types) is: Instead of having a field in this Run DTO to store additional context along with tools and components like we currently do, any additional context will be turned into content blocks and added to the user's message, maybe all as `V1ResourceContentDto`s. Is that right?

MichaelMilstead: Currently the backend system prompt has information about certain pieces of additional context the messages might contain (context attachments, current interactables, etc.) along with examples of the shapes of those objects. I wonder if we still want the DTO to expect certain types additional content with specific shapes, or whether we should be less explicit in the system prompt?

MichaelMilstead (approved):
Nice! Just left a couple comments for clarification

apps/api/src/v1/dto/run.dto.ts
alecf: @MichaelMilstead huh, I did not include anything about additional context, definitely an oversight. I actually do think that's the right approach though (using content blocks) and we leave it as convenience behavior in the SDK

Except... when we retrieve a thread's messages, we might want to filter that out. I wonder if we want to somehow flag those content blocks as hidden / additional / etc?

hmm. I don't think that blocks this PR but worth thinking about before officially releasing the 1.0 API.

packages/core/src/threads.ts
```diff
@@ -292,6 +292,22 @@ export enum GenerationStage {
...
+  IDLE = "idle",
+  /** A run is currently executing */
+  RUNNING = "running",
+  /** Run is paused waiting for client-side tool results */
+  AWAITING_INPUT = "awaiting_input",
+  /** Run was cancelled (by user or connection drop) */
+  CANCELLED = "cancelled",
+  /** Run failed with an error */
+  FAILED = "failed",
+}
```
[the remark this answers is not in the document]
alecf: (cleaned this up, went back to proposal, this was an oversight of the implementation)

PR #1784, opened 2026-01-15
refactor(api): improve storage config type

### LINKED ISSUE #1770: refactor(api): type-narrow StorageConfigService.hasStorageConfig
### Context
In `apps/api`, `ThreadsService` adds an `attachment` resource fetcher when S3 is configured. Right now, `StorageConfigService.hasStorageConfig()` returns a plain boolean, so call sites rely on non-null assertions like `this.storageConfig.s3Client!`.

We should tighten the typing so callers can avoid `!`, and also treat whitespace-only `S3_BUCKET` / `API_KEY_SECRET` values as “unset”. (This work was briefly included in PR #1759, but removed to keep that PR focused on the web meta-tag change.)

Reference (original patch): commit `4b24f360d02c09001bf84644012b25adacf88772`.

### What we already have
- `apps/api/src/common/services/storage-config.service.ts`:
  - `s3Client: S3Client | undefined`
  - `bucket: string`
  - `signingSecret: string`
  - `hasStorageConfig(): boolean`
- `apps/api/src/threads/threads.service.ts` uses `hasStorageConfig()` and then calls `createAttachmentFetcher(this.storageConfig.s3Client!, ...)`.

### Scope
Small, self-contained refactor (no new deps):

1. Update `StorageConfigService` to:
   - `trim()` `S3_BUCKET` and default to `user-files` when empty/whitespace.
   - `trim()` `API_KEY_SECRET` and treat empty/whitespace as unset.
   - Change `hasStorageConfig()` to be a type predicate, so within `if (this.storageConfig.hasStorageConfig())` TypeScript knows `this.storageConfig.s3Client` is a real `S3Client`.
   - Ensure the guard checks bucket + signing secret are non-empty after trimming.

2. Update `ThreadsService` (and any other call sites you find) to remove non-null assertions:
   - After the `hasStorageConfig()` check, destructure `{ s3Client, bucket, signingSecret } = this.storageConfig` and pass those into `createAttachmentFetcher(...)`.

### Acceptance criteria
- No non-null assertions (`!`) are required for `s3Client` at call sites after the change.
- Whitespace-only `S3_BUCKET` and `API_KEY_SECRET` are treated as unset.
- CI-style checks pass locally (from repo root):

```bash
npm run lint -- --filter=@tambo-ai-cloud/api --output-logs=none
npm run check-types -- --filter=@tambo-ai-cloud/api --output-logs=none
npm test -- --filter=@tambo-ai-cloud/api --output-logs=none
```

Optional nice-to-have: add a small unit test for `hasStorageConfig()` behavior (missing / whitespace / fully configured).

rakesh-paulraj1:
fixes #1770 

- Adds a TypeScript type predicate to `hasStorageConfig()` that narrows `s3Client` from `S3Client | undefined` to `S3Client`. 
- Trims bucket and secret values to handle whitespace and validates bucket is non-empty.


lachieh:
Thanks, @rakesh-paulraj1. Could you add some test coverage for these changes?

apps/api/src/common/services/storage-config.service.ts
```diff
@@ -34,8 +37,14 @@ export class StorageConfigService {
   /**
    * Check if storage is fully configured for attachment operations.
    * Requires both S3 client and signing secret to be available.
+   *
+   * @returns true if storage is configured with S3 client,non-empty bucket,and signing secret
    */
-  hasStorageConfig(): boolean {
-    return this.s3Client !== undefined && this.signingSecret.length > 0;
+  hasStorageConfig(): this is StorageConfigService & { s3Client: S3Client } {
```
alecf: this intersection is brittle in case the type changes. Use MarkRequired from ts-essentials:
https://github.com/ts-essentials/ts-essentials/tree/master/lib/mark-required

rakesh-paulraj1: Thats a great feedback but i think `MarkRequired` is  used to make the property required but not narrow type 
The `s3Client` is already required `readonly s3Client: S3Client | undefined;` so i feel the MarkRequired has no effect i guess
I think we can use the `Merge` from ts-essentials 
https://github.com/ts-essentials/ts-essentials/tree/master/lib/merge

```
hasStorageConfig(): this is Merge<StorageConfigService, { s3Client: S3Client }> {
  return this.s3Client !== undefined && 
  this.bucket.length > 0 && 
  this.signingSecret.length > 0;
}
```

PR #1791, opened 2026-01-15
feat(web,cli): redirect auth from tambo.co to app.tambo.co

akhileshrangani4:
## Summary
- Add configurable auth redirects from marketing domain (tambo.co) to app domain (console.tambo.co)
- Update CLI default API URL to console.tambo.co

## Why
Auth entrypoints (/login, /api/auth/*) need to redirect from the marketing site to the app subdomain to support the domain split between tambo.co (marketing) and console.tambo.co (application).

## Changes
- `apps/web/next.config.mjs`: Conditional redirects based on `AUTH_REDIRECT_FROM_HOST` env var
- `apps/web/lib/env.ts`: Added `AUTH_REDIRECT_FROM_HOST` server env var
- `cli/src/lib/api-client.ts`: Default API URL changed to `https://console.tambo.co`
- `turbo.json`: Added `AUTH_REDIRECT_FROM_HOST` and `NEXT_PUBLIC_APP_URL` to globalEnv

## Configuration
Set these env vars for production:
```
AUTH_REDIRECT_FROM_HOST=tambo.co
NEXT_PUBLIC_APP_URL=https://console.tambo.co
```

## Test Plan
- [ ] Verify requests to `tambo.co/login` redirect to `console.tambo.co/login`
- [ ] Verify requests to `console.tambo.co/login` work normally (no redirect loop)
- [ ] Verify CLI auth flow works with new default URL

### DIFF
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -15,24 +15,34 @@ const jiti = createJiti(fileURLToPath(import.meta.url));
 jiti.import("./lib/env").catch(console.error);
 
 // Auth redirect config: when AUTH_REDIRECT_FROM_HOST is set, redirect auth routes to NEXT_PUBLIC_APP_URL
+// Supports both bare host and host-with-port matching since Host headers may include ports
 const authRedirectFromHost = process.env.AUTH_REDIRECT_FROM_HOST;
 const authRedirectToUrl = process.env.NEXT_PUBLIC_APP_URL;
+const authRedirectHosts = authRedirectFromHost
+  ? authRedirectFromHost.includes(":")
+    ? [authRedirectFromHost]
+    : [
+        authRedirectFromHost,
+        `${authRedirectFromHost}:443`,
+        `${authRedirectFromHost}:80`,
+      ]
+  : [];
 const authRedirects =
-  authRedirectFromHost && authRedirectToUrl
-    ? [
+  authRedirectHosts.length > 0 && authRedirectToUrl
+    ? authRedirectHosts.flatMap((host) => [
         {
           source: "/login",
-          has: [{ type: "host", value: authRedirectFromHost }],
+          has: [{ type: "host", value: host }],
           destination: `${authRedirectToUrl}/login`,
           permanent: true,
         },
         {
           source: "/api/auth/:path*",
-          has: [{ type: "host", value: authRedirectFromHost }],
+          has: [{ type: "host", value: host }],
           destination: `${authRedirectToUrl}/api/auth/:path*`,
           permanent: true,
         },
-      ]
+      ])
     : [];
 
 /** @type {import('next').NextConfig} */
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -63,6 +63,7 @@ NEXTAUTH_URL=http://localhost:8260
 # When set, redirects /login and /api/auth/* from this host to NEXT_PUBLIC_APP_URL
 # e.g. AUTH_REDIRECT_FROM_HOST=tambo.co with NEXT_PUBLIC_APP_URL=https://console.tambo.co
 # will redirect tambo.co/login -> console.tambo.co/login
+# If no port is specified, also matches :443 and :80 variants to handle Host headers with ports
 AUTH_REDIRECT_FROM_HOST=
 # Get these from a developer. You must use the "developer" clients here, not
 # "production" clients, so that they support localhost-based redirects


cli/src/commands/auth.test.ts
```diff
@@ -65,7 +65,7 @@ jest.unstable_mockModule("../lib/api-client.js", () => ({
     }
   },
   verifySession: async () => mockVerifySessionResult,
-  getApiBaseUrl: () => "https://api.tambo.co",
+  getApiBaseUrl: () => "https://app.tambo.co",
```
lachieh: I don't think that this is what we want, right? `api.tambo.co` is hosting `apps/api` and that's not moving, right?

akhileshrangani4: the cli is using trpc api server, which is currently at `tambo.co` - `apps/web` and will be at `apps.tambo.co`

akhileshrangani4: I might have accidentally set it to `api.tambo.co` before when it needed to be `tambo.co`

apps/web/next.config.mjs
```diff
@@ -14,10 +14,32 @@ const jiti = createJiti(fileURLToPath(import.meta.url));
...
+const authRedirectFromHost = process.env.AUTH_REDIRECT_FROM_HOST;
+const authRedirectToUrl = process.env.NEXT_PUBLIC_APP_URL;
+const authRedirects =
+  authRedirectFromHost && authRedirectToUrl
+    ? [
+        {
+          source: "/login",
+          has: [{ type: "host", value: authRedirectFromHost }],
+          destination: `${authRedirectToUrl}/login`,
+          permanent: false,
```
lachieh: Let's make these permanent. It doesn't really affect anything too seriously, just changes the redirect from a 302 to a 301 and the only real difference is that with a 302, the browser doesn't remember the redirect past the request, whereas 301 it caches it for the session.

akhileshrangani4: yup, thats right, doing it.

akhileshrangani4: done.

cli/src/commands/auth.test.ts
michaelmagan: Not directly related to your PR but this is confusing because of the naming convention here. Since we have an api.tambo.co route.

I wonder if we should add comments // this is the TRPC api route not the NestJS route or open another follow up PR to fix the naming convention so it's more clearer.

I have a feeling this is going to happen over and over again with developers/agents.

akhileshrangani4: yes, true, I will add comments here.

akhileshrangani4: added

lachieh: Appreciate it. I think changing the name of the variable is actually a good idea too. We're already using `NEXT_PUBLIC_TAMBO_API_URL` in a few apps which is eerily similar. I'll think about the name tonight but I'm leaning towards `TAMBO_CLI_CONSOLE_URL`.

apps/web/next.config.mjs
```diff
@@ -14,10 +14,32 @@ const jiti = createJiti(fileURLToPath(import.meta.url));
...
+        },
+        {
+          source: "/api/auth/:path*",
+          has: [{ type: "host", value: authRedirectFromHost }],
+          destination: `${authRedirectToUrl}/api/auth/:path*`,
+          permanent: false,
+        },
+      ]
+    : [];
```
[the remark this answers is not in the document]
lachieh: This is not wrong, but I don't know that it's worth handling the case of someone setting the `NEXT_PUBLIC_APP_URL` with a trailing slash.

cli/src/lib/api-client.ts
```diff
@@ -87,8 +87,12 @@ interface StubRouter {
...
 // ============================================================================
 
+/**
+ * Returns the base URL for the tRPC API (apps/web at app.tambo.co).
+ * Note: This is NOT the NestJS REST API at api.tambo.co.
+ */
 export function getApiBaseUrl(): string {
-  return process.env.TAMBO_API_URL ?? "https://tambo.co";
+  return process.env.TAMBO_API_URL ?? "https://app.tambo.co";
 }
```
lachieh, line 96: Per slack convo, let's rename to console

```suggestion
/**
 * Returns the base URL for the Tambo Console API (apps/web at app.tambo.co).
 * Note: This is NOT the NestJS REST API at api.tambo.co.
 */
export function getApiBaseUrl(): string {
  return process.env.TAMBO_CLI_CONSOLE_BASE_URL ?? "https://app.tambo.co";
}
```

Optionally, rename the function to `getConsoleBaseUrl` (if not in this PR, then a `good first issue`)

apps/web/next.config.mjs
```diff
@@ -14,10 +14,32 @@ const jiti = createJiti(fileURLToPath(import.meta.url));
 // Import env here to validate the environment variables during build. Using jiti we can import .ts files :)
 jiti.import("./lib/env").catch(console.error);
 
+// Auth redirect config: when AUTH_REDIRECT_FROM_HOST is set, redirect auth routes to NEXT_PUBLIC_APP_URL
+const authRedirectFromHost = process.env.AUTH_REDIRECT_FROM_HOST;
+const authRedirectToUrl = process.env.NEXT_PUBLIC_APP_URL;
```
lachieh, line 20: Is the same app instance going to be hosting the two domains until we split the app out (i.e. both domains pointing to one railway service)? Or are we going to deploy two services for a short time?

lachieh: It's same instance, both domains

akhileshrangani4:
I am only adding `AUTH_REDIRECT_FROM_HOST` env variable with this pr which is temporary so we should be fine.

cli/src/lib/api-client.ts
akhileshrangani4: updated this.

alecf (approved):
mostly deferring to @lachieh here - I assume his 👍 on the charlie comment is that we should do that...

akhileshrangani4:
@alecf yup, did that now

PR #1821, opened 2026-01-19
fix(web): prevent optional env vars from failing validation when empty

### LINKED ISSUE #1819: chore(web): comment out optional env vars in .env.example to prevent validation errors
## Problem

First-time developers copying `.env.example` to `.env` encounter validation errors like:

```
Invalid environment variables: [
  {
    code: 'too_small',
    minimum: 1,
    type: 'string',
    message: 'String must contain at least 1 character(s)',
    path: [ 'INTERNAL_SLACK_USER_ID' ]
  },
  ...
]
```

This happens because `@t3-oss/env-nextjs` with Zod validation treats empty strings differently than undefined:
- Variable not set → `undefined` → passes `.optional()` validation
- Variable set to empty (`VAR=`) → `""` → fails `.min(1)` validation

## Proposed Solution

1. Update validation to accept empty string as a valid input. See `allowEmptyString` usage.

2. Comment out optional env vars that don't have default values in `.env.example`:

   ```bash
   # Before:
   SLACK_OAUTH_TOKEN=
   INTERNAL_SLACK_USER_ID=
   RESEND_API_KEY=
   
   # After:
   # SLACK_OAUTH_TOKEN=
   # INTERNAL_SLACK_USER_ID=
   # RESEND_API_KEY=
   ```

See [CONTRIBUTING.md](https://github.com/tambo-ai/tambo/blob/main/CONTRIBUTING.md) for full contribution guidelines.

MethodManav:
## Proposed Changes

- Fixes #1819 

## Problem

First-time developers copying `.env.example` to `.env.local` encounter Zod validation errors 

This occurs because some environment variables are marked as **optional** while still using `.min(1)`:

- `.optional()` allows the value to be `undefined`
- `.min(1)` explicitly disallows empty strings (`""`)

When a variable exists in `.env` but is left empty (`VAR=`), it is parsed as an empty string rather than `undefined`. As a result, validation fails even though the variable is intended to be optional.

Since optional environment variables in `.env.example` are commonly left empty, this breaks local setup for new contributors.

---

## Solution

Empty strings are transformed to `undefined` for optional environment variables before validation, preserving strict validation for non-empty values.

### Before
```ts
INTERNAL_SLACK_USER_ID: z.string().min(1).optional(),
SLACK_OAUTH_TOKEN: z.string().min(1).optional(),

### After
```ts
INTERNAL_SLACK_USER_ID: z.string().transform(allowEmptyString).optional(),
SLACK_OAUTH_TOKEN: z.string().transform(allowEmptyString).optional(),

### DIFF
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -58,7 +58,7 @@ GITHUB_TOKEN=
 
 ## Nextauth login stuff
 # Run `openssl rand -hex 8` to generate a random secret
-NEXTAUTH_SECRET=YOUR_NEXTAUTH_SECRET_HERE 
+NEXTAUTH_SECRET=YOUR_NEXTAUTH_SECRET_HERE
 NEXTAUTH_URL=http://localhost:8260
 # When set, redirects /login and /api/auth/* from this host to NEXT_PUBLIC_APP_URL
 # e.g. AUTH_REDIRECT_FROM_HOST=tambo.co with NEXT_PUBLIC_APP_URL=https://console.tambo.co
--- a/apps/web/lib/env.ts
+++ b/apps/web/lib/env.ts
@@ -45,11 +45,14 @@ export const env = createEnv({
     EMAIL_FROM_DEFAULT: z.string().transform(allowEmptyString).optional(),
 
     // Whitelabeling (server-side copies; optional so can be omitted)
-    TAMBO_WHITELABEL_ORG_NAME: z.string().min(1).optional().or(z.literal("")),
+    TAMBO_WHITELABEL_ORG_NAME: z
+      .string()
+      .transform(allowEmptyString)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: apps/web/.env.example [hunk 1]
```diff
@@ -58,7 +58,7 @@ GITHUB_TOKEN=
 
 ## Nextauth login stuff
 # Run `openssl rand -hex 8` to generate a random secret
-NEXTAUTH_SECRET=...
+NEXTAUTH_SECRET=YOUR_NEXTAUTH_SECRET_HERE
```
michaelmagan: Trailing space here - can you remove it?

MethodManav: Fixed. Trailing space removed. Thanks for pointing it out.

PR #1826, opened 2026-01-19
feat(cli): agent-first CLI rewrite (experimental)

michaelmagan:
### Background

On Friday I opened some PRs to improve the Claude Code experience with our CLI and realized something: our first CLI was built with humans in mind. Interactive prompts, confirmation dialogs, nice menus—all that makes sense when a person is typing commands.

But in a world where coding agents are running CLIs, that's actually friction. Prompts block execution, confirmations need special handling, and spinners pollute the stdout that agents are trying to parse.

So I had Claude reimagine our CLI with agents as the primary user. Rewrote it with that expectation as the default, added back interactive behavior where it makes sense (TTY detection, missing required flags), and packaged it with a skill that can use this CLI to set up and build Tambo apps.

### What Changed

- Commands work without prompts when you give them the right flags
- `--json` output mode so agents (or you with `jq`) can parse results
- `--yes`/`--no-interactive` flags to skip confirmations
- Stdin support for piped input
- Claude Code skill (`/tambov1`) for building agent components with Tambo

### Status

- **Experimental** - not sure if we ship this or backport the concepts to the current CLI
- **Needs tests** - coverage gaps, especially around non-interactive flows

### Testing Locally

```bash
cd cli-v1
npm install
npm run build

# Run commands
./dist/cli.js init --help
./dist/cli.js components --json | jq '.[]'

# Or link globally
npm link
tambo-v1 --help
```

### DIFF
--- a/cli-v1/skills/tambov1/marketplace.json
+++ b/cli-v1/skills/tambov1/marketplace.json
@@ -23,12 +23,7 @@
     "ai-tools",
     "context-attachments"
   ],
-  "platforms": [
-    "claude",
-    "codex",
-    "gemini",
-    "chatgpt"
-  ],
+  "platforms": ["claude", "codex", "gemini", "chatgpt"],
   "commands": [
     {
       "name": "create-agent-component",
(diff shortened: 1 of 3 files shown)


cli-v1/AGENTS.md
```diff
@@ -0,0 +1,70 @@
+# AGENTS.md - cli-v1 (tambov1)
+
+**Read root @../AGENTS.md` first.** This adds cli-v1-specific rules.
```
alecf: no need to add this - I believe it should be doing this automatically

PR #1841, opened 2026-01-20
fix(web): add guard for SLACK_OAUTH_TOKEN in callSlackAPI

### LINKED ISSUE #1840: fix(web): add guard for SLACK_OAUTH_TOKEN before API calls
## Context

Discovered during review of #1821 by @MethodManav.

## Problem

The Slack integration uses `SLACK_OAUTH_TOKEN` directly without checking if it's configured:

https://github.com/tambo-ai/tambo/blob/main/apps/web/lib/slack.ts#L19

If the env var is unset or empty, API calls fail with an unhelpful "Bearer undefined" error.

## Solution

Add a guard at the top of `callSlackAPI()` that throws a clear error if `SLACK_OAUTH_TOKEN` is not set.

## Acceptance Criteria

- [ ] Add guard to `callSlackAPI()`
- [ ] Add a test for the guard

MethodManav:
# Context

Fixes #1840 

Currently, the Slack integration uses `SLACK_OAUTH_TOKEN` directly without validating whether it is configured. When the environment variable is missing or empty, Slack API calls fail with an unhelpful **"Bearer undefined"** error.

---

## Changes

- **Added Guard Clause**
  - Added a validation check at the start of `callSlackAPI` to ensure `SLACK_OAUTH_TOKEN` exists and is not an empty string.

- **Improved Error Messaging**
  - The function now throws a clear, descriptive error:
    ```
    SLACK_OAUTH_TOKEN environment variable is not set
    ```
- **Unit Tests**
  - Added a unit test in `lib/slack.test.ts` to verify the guard logic when the token is missing or empty.

---

## Result

- Prevents invalid Slack API requests
- Fails fast with an actionable error message
- Improves developer experience during configuration and debugging

### DIFF
--- a/apps/web/lib/slack.ts
+++ b/apps/web/lib/slack.ts
@@ -12,29 +12,25 @@ export async function callSlackAPI<T>(
   endpoint: string,
   body: Record<string, unknown>,
 ): Promise<T> {
-  try {
-    if (!env.SLACK_OAUTH_TOKEN || env.SLACK_OAUTH_TOKEN.trim() === "") {
-      throw new Error("SLACK_OAUTH_TOKEN environment variable is not set");
-    }
-    const response = await fetch(`${SLACK_API_BASE}/${endpoint}`, {
-      method: "POST",
-      headers: {
-        "Content-Type": "application/json",
-        Authorization: `Bearer ${env.SLACK_OAUTH_TOKEN}`,
-      },
-      body: JSON.stringify(body),
-    });
-
-    const data = await response.json();
-
-    if (!data.ok) {
-      throw new Error((data as SlackAPIError).error);
-    }
-
-    return data as T;
-  } catch (error) {
-    throw new Error(`Slack API call failed: ${error}`);
+  if (!env.SLACK_OAUTH_TOKEN) {
+    throw new Error("SLACK_OAUTH_TOKEN environment variable is not set");
   }
+  const response = await fetch(`${SLACK_API_BASE}/${endpoint}`, {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+      Authorization: `Bearer ${env.SLACK_OAUTH_TOKEN}`,
+    },
+    body: JSON.stringify(body),
+  });
+
+  const data = await response.json();
+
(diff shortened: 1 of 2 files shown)


apps/web/lib/slack.ts
```diff
@@ -12,22 +12,29 @@ export async function callSlackAPI<T>(
...
 ): Promise<T> {
-  const response = await fetch(`${SLACK_API_BASE}/${endpoint}`, {
-    method: "POST",
-    headers: {
-      "Content-Type": "application/json",
-      Authorization: `Bearer ${env.SLACK_OAUTH_TOKEN}`,
-    },
-    body: JSON.stringify(body),
-  });
+  try {
```
michaelmagan: The try-catch wrapper causes double-wrapped errors like `"Slack API call failed: Error: SLACK_OAUTH_TOKEN..."`. Just throw directly - see `inviteInternalUser` in this file (line 146) for the pattern.

apps/web/lib/slack.ts
```diff
@@ -12,22 +12,29 @@ export async function callSlackAPI<T>(
...
+    if (!env.SLACK_OAUTH_TOKEN || env.SLACK_OAUTH_TOKEN.trim() === "") {
```
michaelmagan: The `.trim()` check is redundant - your #1821 already transforms empty strings to `undefined`, so `!env.SLACK_OAUTH_TOKEN` handles both cases.

michaelmagan (dismissed):
Hey @MethodManav, nice follow-up to #1821! The guard makes sense now that empty strings become `undefined`. One ask: can you simplify to match existing patterns? See inline comments.

apps/web/lib/slack.ts
MethodManav: Okay, I’ve dropped the try/catch in callSlackAPI and now throw errors directly

apps/web/lib/slack.ts
MethodManav: Agreed — removed the .trim() check

lachieh (approved):

lachieh:
Thanks, @MethodManav!

PR #1793, opened 2026-01-15
feat(ui-registry): create centralized component registry package

lachieh:
## Summary

- Create `packages/ui-registry` as a shared internal package for Tambo components
- This is **Phase 1 only** of centralizing the component registry per `plans/centralize-component-registry.md`
- Copy all components from `cli/src/registry` to the new package
- Update imports to use `@tambo-ai/ui-registry/*` paths
- Add package.json with explicit exports for all components
- Move CSS files to `src/styles/` directory and tailwind config

This enables internal apps (showcase, web, docs) to import directly from `@tambo-ai/ui-registry` instead of maintaining separate copies.

## What's NOT included (future PRs)

**Phase 2: Migrate Consumers**
- Update showcase, web, and docs apps to import from `@tambo-ai/ui-registry`
- Add `transpilePackages: ["@tambo-ai/ui-registry"]` to Next.js configs
- Delete old component directories from consumer apps
- Create CLI prebuild script to copy registry to `cli/dist/registry/`

**Phase 3: Verification & Cleanup**
- Integration testing of `tambo add` with new structure
- HMR verification across all consumer apps
- CI enforcement for exports matching filesystem

## Test plan

- [x] `npm run lint` passes for ui-registry package
- [x] `npm run check-types` passes for ui-registry package

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1872, opened 2026-01-21
refactor(ui-registry): migrate consumers to centralized component registry

lachieh:
## Summary

Phase 2 of the component registry centralization (TAM-976). Migrates all consumer apps (web, docs, showcase) to import shared components from `@tambo-ai/ui-registry` instead of maintaining local copies.

- Update imports from local paths to `@tambo-ai/ui-registry/components/*`
- Add `@tambo-ai/ui-registry` workspace dependency to web, docs, showcase
- Add `transpilePackages` config to all Next.js apps
- Delete ~20,000 lines of duplicated component code
- Remove sync script infrastructure (`scripts/sync-showcase-components.ts`)
- Remove `sync:showcase` turbo task and lint-staged trigger
- Keep web-specific components (marketing, interactables integration)

## Test plan

- [x] Verify `npm run lint` passes
- [x] Verify `npm run check-types` passes
- [x] Verify `npm test` passes
- [x] Start showcase dev server, confirm components render
- [x] Start web dev server, confirm components render
- [x] Start docs dev server, confirm components render
- [x] Test HMR: edit component in `packages/ui-registry/`, verify change appears in running apps

## Related

- Fixes TAM-976
- Follows Phase 1: #1793

### DIFF
--- a/showcase/src/app/components.css
+++ b/showcase/src/app/components.css
@@ -3,3 +3,4 @@
 
 /* Styles for ui components are imported directly from the ui-registry package */
 @import "@tambo-ai/ui-registry/styles.css";
+@import "@tambo-ai/ui-registry/styles/globals-v4.css";


apps/web/app/(authed)/internal/smoketest/page.tsx
```diff
@@ -5,7 +5,7 @@ import { Card } from "@/components/ui/card";
 import {
   ThreadContent,
   ThreadContentMessages,
-} from "@/components/ui/tambo/thread-content";
+} from "@tambo-ai/ui-registry/components/thread-content";
```
michaelmagan, line 8: Do we want every update to our UI components to automatically update our web app implementation?

I think this is obviously true for showcase (where we demo latest components) but not clear we want this in our app?

docs/src/app/(docs)/[[...slug]]/page.tsx
```diff
@@ -1,5 +1,5 @@
 import { Suspense } from "react";
-import { MessageThreadCollapsible } from "@/components/tambo/message-thread-collapsible";
+import { MessageThreadCollapsible } from "@tambo-ai/ui-registry/components/message-thread-collapsible";
```
Comment (michaelmagan) [inline docs/src/app/(docs)/[[...slug]]/page.tsx]: same thoughts in our docs. do we want this to always be latest version from component library?

apps/web/app/(authed)/internal/smoketest/page.tsx
lachieh: Yeah it's a fair question. I think since we already had the sync happening at commit with lint-staged we go with it, and then if we need to skip or modify certain components, those become the exception and they are copied back into the individual apps.

apps/web/providers/message-thread-panel-provider.tsx
```diff
@@ -1,6 +1,6 @@
 "use client";
 
-import type { TamboEditor } from "@/components/ui/tambo/text-editor";
```
alecf, line 3: this kinda looks like a file that would be in the components library? Or maybe it has custom code in it?

lachieh: Yeah this one is custom to `web`. The remaining components in `apps/web/components/ui/tambo` are diverged enough that they are staying in place.

akhileshrangani4 (approved):

Issue #1706, opened 2026-01-08
Add useMemo optimization to ScrollableMessageContainer


michaelmagan:
## Summary

The ScrollableMessageContainer recalculates values on every render that could be memoized for better performance.

## Files

- `packages/ui-registry/src/components/scrollable-message-container/scrollable-message-container.tsx`

## Task

Identify expensive calculations and wrap them with useMemo to prevent unnecessary recalculations.

## Verification

- [ ] Tests added/updated to cover new functionality
- [ ] `npm run check-types` passes
- [ ] `npm run lint` passes
- [ ] `npm test` passes
- [ ] Component renders efficiently (no unnecessary recalculations)

See [CONTRIBUTING.md](https://github.com/tambo-ai/tambo/blob/main/CONTRIBUTING.md) for full contribution guidelines.

bashward:
Can I work on this @michaelmagan ?

vishalkumargeed:
```cli/src/registry/scrollable-message-container/scrollable-message-container.tsx```
This doesn't exit in the codebase ! 
I guess this is the path of the file after building and used by other applications ! 

Correct file path would be : ``` packages/ui-registry/src/components/scrollable-message-container/scrollable-message-container.tsx```

lachieh:
Hey, thanks for checking, @bashward. We don't assign issues until someone has at least opened a PR. In this case, @vishalkumargeed beat you to it but feel free to look at any other issues with `good first issue` or the `status: help wanted` label.

michaelmagan:
Closing for now as we focus on 1.0 launch.

PR #1887, opened 2026-01-22
fix(cli): optimization to ScrollableMessageContainer #1706

vishalkumargeed:
### What This Improves
- handleScroll is now stable across renders, preventing unnecessary re-renders if it were passed as a prop (though in this case it's just attached to onScroll).
- generationStage only recalculates when the actual generation stage changes, not on every render.

### 2 Commits 
- Optimized the SchrollableMessageContainer 
- Fixed the test script as earlier it was executing the jest as a javascript file ; now it executes as a shell script. Same thing applied in the ```  packages/ui-registry/package.json``` : 
``` "test": "NODE_OPTIONS=--experimental-vm-modules jest" ```

### Verification Done

- Tests added/updated to cover new functionality
- npm run check-types passes
- npm run lint passes
- npm test passes
- Component renders efficiently (no unnecessary recalculations)

### DIFF
--- a/packages/ui-registry/src/components/scrollable-message-container/scrollable-message-container.tsx
+++ b/packages/ui-registry/src/components/scrollable-message-container/scrollable-message-container.tsx
@@ -3,7 +3,7 @@
 import { GenerationStage, useTambo } from "@tambo-ai/react";
 import { cn } from "@tambo-ai/ui-registry/utils";
 import * as React from "react";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState, useMemo } from "react";
 
 /**
  * Props for the ScrollableMessageContainer component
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/ui-registry/src/components/scrollable-message-container/scrollable-message-container.tsx [hunk 1]
```diff
@@ -50,10 +50,13 @@ export const ScrollableMessageContainer = React.forwardRef<
     }));
   }, [thread.messages]);
 
-  const generationStage = thread?.generationStage ?? GenerationStage.IDLE;
+  const generationStage = React.useMemo(
+    () => thread?.generationStage ?? GenerationStage.IDLE,
+    [thread?.generationStage],
+  );
```
lachieh, line 56: Since we're already importing other hooks from the `react` package, this should follow the same pattern.

```ts
import { useMemo } from "react"
// ...
const generationStage = useMemo(...)
```

lachieh (approved):

PR #1905, opened 2026-01-24
fix(create-tambo-app): use cross-spawn for Windows compatibility

michaelmagan:
## Summary

- Replace native Node.js `spawn` with `cross-spawn` to fix Windows compatibility
- Fixes "npx was not found on your PATH" error on Windows even when npx is installed

## Problem

On Windows, `npx` is actually a batch script (`npx.cmd`). Node's native `spawn()` without `shell: true` cannot resolve `.cmd` extensions, causing the error:

```
create-tambo-app: `npx` was not found on your PATH. Install Node.js (which provides npx) or ensure `npx` is available, then retry.
```

This happens even when `npx -V` works fine in the same terminal.

## Solution

Use `cross-spawn` - the industry-standard package for cross-platform process spawning. It's a drop-in replacement that handles Windows quirks automatically without needing `shell: true` (which has security implications).

**Why cross-spawn over alternatives:**
- `shell: true` - shell injection risk
- Platform detection (`npx.cmd` on Windows) - brittle, misses edge cases
- `cross-spawn` - used by ESLint, Jest, npm itself. Already in our dep tree.

## Changes

- `create-tambo-app/src/index.ts` - one import change
- `create-tambo-app/package.json` - add `cross-spawn` dependency

## Test plan

- [ ] Verify build passes: `npm run build -w create-tambo-app`
- [ ] Test on Windows: `npx create-tambo-app@latest my-app`
- [ ] Test on macOS/Linux: `npx create-tambo-app my-app`

### DIFF
--- a/create-tambo-app/src/index.ts
+++ b/create-tambo-app/src/index.ts
@@ -27,6 +27,6 @@ child.on("error", (err: NodeJS.ErrnoException) => {
   process.exit(1);
 });
 
-child.on("exit", (code) => {
+child.on("exit", (code: number | null) => {
   process.exit(code ?? 0);
 });


lachieh:
@CharlieHelps fix the merge conflicts against main

Issue #1817, opened 2026-01-17
CLI doesn't configure PostCSS/Vite plugin for Tailwind v4 in non-Next.js projects


michaelmagan:
## Problem

When using `npx tambo init` or `npx tambo full-send` in a Vite project with Tailwind CSS v4, the CLI correctly sets up `globals.css` with the v4 syntax (`@import "tailwindcss"`) but doesn't configure the PostCSS toolchain needed to process it.

This results in **completely broken styles** - the Tailwind classes aren't compiled at all.

## Root Cause

1. **Vite detection is commented out** in `cli/src/utils/framework-detection.ts` (lines 59-62)
2. **No `postcss.config.mjs` is created** - the CLI assumes the framework (Next.js) already handles PostCSS
3. **`@tailwindcss/postcss` or `@tailwindcss/vite` aren't installed** during setup

Next.js has built-in PostCSS support so this works fine there. But Vite (and likely other frameworks like Remix, Astro, etc.) need explicit configuration.

## Reproduction

1. Create a new Vite React project: `npm create vite@latest my-app -- --template react-ts`
2. Install Tailwind v4: `npm install tailwindcss@^4`
3. Run `npx tambo full-send`
4. Start the dev server - styles are completely broken

Example repo: https://github.com/alessandro308/tambo-vite-experiment

## Expected Behavior

When Tailwind v4 + Vite is detected, the CLI should either:

**Option A (Vite plugin - recommended):**
- Install `@tailwindcss/vite`
- Update `vite.config.ts` to include the plugin

**Option B (PostCSS):**
- Install `@tailwindcss/postcss`
- Create `postcss.config.mjs`

## Workaround

Users can manually fix this:

```bash
npm install @tailwindcss/vite
```

```ts
// vite.config.ts
import tailwindcss from '@tailwindcss/vite'

export default defineConfig({
  plugins: [react(), tailwindcss()],
})
```

## Affected Frameworks

- Vite (confirmed)
- Likely also: Remix, Astro, SvelteKit, and other non-Next.js frameworks using Tailwind v4

michaelmagan:
Thanks for reporting this @alessandro308!

sriramsowmithri9807:
hey @michaelmagan @lachieh check out the PR of #1915

themavik:
This issue appears to be resolved on the current `main` branch. The CLI now:

[...]

PR #1913, opened 2026-01-25
CLI doesn't configure PostCSS/Vite plugin for Tailwind v4 in non-Next.js projects 

vishalkumargeed:
This fixes issue #1817

# My approach 
I have gone with OPTION A   that is : 
Install ```@tailwindcss/vite```
Update ```vite.config.ts``` to include the plugin

# Changes I have made
1.  added the framework detection : Vite 
2.   ``` tailwind-setup.ts``` giving the user flexibity that :
-  by using ``` npx tambo full-send```  this command automatically install tailwind into the project >4.0.0 
-  No need to install external command when setup react using vite . As while setup vite there is a step in order to bring tailwindcss into the project . I thought this could be a good feature ! As the common flow is :  https://tailwindcss.com/docs/installation/using-vite

4.   created this file ``` cli\src\commands\add\setupViteTailwindPlugin.ts``` . This has the logic for creating : 
   -  Ensures that the @tailwindcss/vite plugin is installed (using the correct package manager for the project).
    -  Locates the Vite configuration file (vite.config.ts, .js, .mjs, or .cjs).
    -  If a Vite config file exists:
         - For TypeScript configs, parses and modifies the file using ts-morph, ensuring `tailwindcss` is imported and added to the `plugins` array.
         - For JavaScript configs, modifies the file contents via string manipulation to import `tailwindcss` and add it to the `plugins` array.
    4. If no config exists, generates a new vite.config.ts with React and Tailwind plugins preconfigured.
    
    
# Here is the Video demo for the same that it's working 
## 1.  Vite : React + Javascript : 
 -  import tailwindcss from '@tailwindcss/vite'
   - plugins: [react(), tailwindcss()]
  - Should show: "@import "tailwindcss" in App.css
  - In package.,json :
   "tailwindcss": "^4.x.x"
    "@tailwindcss/vite": "^4.x.x"
https://drive.google.com/file/d/18uMzqh8UTC-oShABBbEEjqTX_IlDsffW/view?usp=sharing

## 2. Vite : React + Typescript 
https://drive.google.com/file/d/1jB8C508BhLkFhNsbwbeiVYZn8HgTsm1c/view?usp=sharing
And here is the screenshot that the dependencies got isntalled after executing "npx tambo full-send"
image

# As you can see in the above video : 
1. When configuring the project with Vite : React + Typescript , it shows this error : 
image

2. When google about this error it shows : 
When selecting React + TypeScript, if vite.config.ts exists but lacks a plugins property, the code threw an error instead of adding it. This caused setup to fail even though Tailwind packages were installed.

But if that's the case then my for react+javascript logic should not work! 
If you can come up with this solution I'm more than happy ! 
Also I asked this issue in the discord channel also  :https://discord.com/channels/1251581895414911016/1464967150773993645

# In short 
- User creates a new Vite application using "npm create vite@latest <name of the   folder >" 
- Executes "npx tambo full-send" with this no need to perform the step to install tailwind as this command installs tailwind css 
- selects javascript (everything works fine ) 
- selects typescript : dependenices gets installed but not overwritting the vite.config.ts file !

### DIFF
--- a/cli/src/commands/add/component.ts
+++ b/cli/src/commands/add/component.ts
@@ -98,6 +98,7 @@ export function cn(...inputs: ClassValue[]) {
     "postcss",
     "autoprefixer",
     "tailwind-merge",
+    "@tailwindcss/vite",
     "clsx",
   ]);
 
--- a/cli/src/commands/add/setupViteTailwindPlugin.ts
+++ b/cli/src/commands/add/setupViteTailwindPlugin.ts
@@ -1,0 +1,370 @@
+/**
+ * @param projectRoot The root directory of the project
+ */
+
+import chalk from "chalk";
+import fs from "fs";
+import path from "path";
+import {
+  Project,
+  ScriptKind,
+  SyntaxKind,
+  type ObjectLiteralExpression,
+} from "ts-morph";
+import { execFileSync } from "../../utils/interactive.js";
+import {
+  detectPackageManager,
+  formatPackageArgs,
+  getDevFlag,
+  getInstallCommand,
+} from "../../utils/package-manager.js";
+
+export async function setupViteTailwindPlugin(projectRoot: string) {
+  // Configuring how to install
+  const pm = detectPackageManager(projectRoot);
+  const installCmd = getInstallCommand(pm);
+  const devFlag = getDevFlag(pm);
+
+  const packageJsonPath = path.join(projectRoot, "package.json");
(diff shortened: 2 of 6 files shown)


michaelmagan:
Thanks for the PR\! We're laser focused on 1.0 launch and need to close this. If you'd still like to contribute, we have a side quest going for starter templates: https://www.wemakedevs.org/hackathons/tambo/starter-template

PR #1914, opened 2026-01-25
fix(scrollable-message-container): memoize ScrollableMessageContainer JSX for better performance

EhDaYaGhaly:
Wrap the entire JSX container in useMemo to prevent unnecessary re-renders when dependencies haven't changed.

This improves performance by avoiding recalculating the JSX tree on every render when props like className, children, or event handlers remain the same.

Related to #1706


michaelmagan:
Hey thanks for this\! We're in launch mode and don't have bandwidth to review right now — closing for now. Still want to contribute? We're looking for starter templates: https://www.wemakedevs.org/hackathons/tambo/starter-template

PR #1923, opened 2026-01-26
fix(ui-registry): pair tool calls with results by tool_call_id

### LINKED ISSUE #1701: Pair tool-call content blocks with tool-result by ID
## Summary

Tool calls and results should be paired by their IDs for accurate rendering and correlation. Currently the pairing logic may not handle all edge cases.

## Files

- `cli/src/registry/message/message.tsx`

## Task

Improve tool-call/tool-result pairing logic to match by ID reliably, handling cases where results arrive out of order or IDs are missing.

## Verification

- [ ] `npm run check-types` passes
- [ ] `npm run lint` passes
- [ ] `npm test` passes
- [ ] Tool calls correctly pair with their results in the UI

rakesh-paulraj1:
## Summary
Fixes tool-call/tool-result pairing to use ID-based matching instead of positional matching

## Changes
- Replace positional matching with `Array.find()` using `tool_call_id`
- Search entire thread messages array regardless of result order
- Hide ToolcallInfo component when `tool_call_id` is missing (broken state)
- Added 5 comprehensive tests for ID matching scenarios

## Test Coverage
-  Tool call pairs with matching result by `tool_call_id`
-  Component hidden when `tool_call_id` is missing
-  Correct pairing with out-of-order results
-  Multi-tool scenarios with parallel calls
-  No result shown when `tool_call_id` doesn't match

Fixes #1701

### DIFF
--- a/apps/api/jest.config.ts
+++ b/apps/api/jest.config.ts
@@ -53,7 +53,7 @@ const config: Config = {
   coverageThreshold: {
     global: {
       branches: 64,
-      lines: 49,
+      lines: 52,
     },
   },
 
--- a/apps/web/jest.config.ts
+++ b/apps/web/jest.config.ts
@@ -51,8 +51,8 @@ const config: Config = {
   testPathIgnorePatterns: ["/node_modules/", "/.next/", "/dist/"],
   coverageThreshold: {
     global: {
-      branches: 8,
-      lines: 13,
+      branches: 9,
+      lines: 16,
     },
   },
 };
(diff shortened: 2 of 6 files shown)


michaelmagan:
Thanks for this\! We're heads down on 1.0 launch and can't review PRs right now — going to close this for now. If you still want to contribute, we're running a side quest for starter templates: https://www.wemakedevs.org/hackathons/tambo/starter-template

PR #1968, opened 2026-01-28
fix(cli): install components to tambo/ subdirectory in full-send mode

akhileshrangani4:
## Summary

- Fix component installation in `tambo full-send` to correctly install to `tambo/` subdirectory
- Fix cross-component file references by converting legacy `src/registry/` paths to correct `components/` or `lib/` prefix
- Fix spinner blocking interactive prompts during full-send by deferring tailwind setup
- Fix `globals.css` detection to find existing files instead of creating duplicates in wrong directories
- Export additional components (`Tooltip`, `TooltipProvider`, `createMarkdownComponents`, `markdownComponents`)

## Why

When running `npx tambo full-send`, components were being installed with incorrect file content (just the path string instead of actual code) for cross-component dependencies like `suggestions-tooltip.tsx`. Additionally, the installation would hang when the user chose "No" for src/ directory because the ora spinner was blocking the tailwind CSS prompt.

## Test Plan

- [x] `npm run lint` passes
- [x] `npm test` for CLI (add, init, update tests) - all 57+ tests pass
- [ ] Test `npx tambo full-send` with "Yes" for src/
- [ ] Test `npx tambo full-send` with "No" for src/ - should no longer hang
- [ ] Test `npx tambo add message-input` - suggestions-tooltip.tsx installs correctly
- [ ] Test `npx tambo add message-thread-full` - lib/thread-hooks.ts installs correctly

### DIFF
--- a/packages/ui-registry/src/components/message-suggestions/message-suggestions.tsx
+++ b/packages/ui-registry/src/components/message-suggestions/message-suggestions.tsx
@@ -409,4 +409,10 @@ function getSuggestionButtonClassName({
   return "bg-background hover:bg-accent hover:text-accent-foreground";
 }
 
-export { MessageSuggestions, MessageSuggestionsList, MessageSuggestionsStatus };
+export {
+  MessageSuggestions,
+  MessageSuggestionsList,
+  MessageSuggestionsStatus,
+  Tooltip,
+  TooltipProvider,
+};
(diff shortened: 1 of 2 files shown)


akhileshrangani4:
@lachieh I also added fix for some other things, could you take a look? thanks! updated PR description too!

lachieh (approved):
Also good additions! As a follow up, tests to cover these would be great but doesn't have to hold up this PR.

akhileshrangani4:
@lachieh yes will do next

PR #2002, opened 2026-01-28
fix(web): store Google auth name and image in user metadata

akhileshrangani4:
## Summary
- Add explicit profile callback to Google provider to ensure name and image are properly mapped from the OAuth response
- Fix `createUser` in the adapter to store both name and image (as `avatar_url`) in `raw_user_meta_data`
- Fix `updateUser` to preserve existing metadata using atomic JSONB merge instead of overwriting
- Skip unnecessary DB writes when no metadata or email updates are provided
- Make `User.idToken` optional since profile callback doesn't have access to it

## Why
We need access to Google user names for downstream workflows and integrations. Unlike GitHub (which has a public API to fetch user data from provider ID), Google doesn't expose a public endpoint for this. The `access_token` expires in 1 hour, so it can't be used for historical records. The name data exists in the `id_token` JWT stored in `identities.identity_data`, but many external tools don't support JWT decoding.

Rather than requiring external tooling to parse JWTs, this should be handled at the source: store the name fields from Google's profile response directly in `auth.users.raw_user_meta_data` during login.

## What this means for users
- **New users**: Name and avatar are stored in `auth.users.raw_user_meta_data` on first login
- **Existing users who log in again**: Name and avatar get populated on next login
- **Existing users who never log in again**: Would need a one-time backfill migration (data exists in `identities.identity_data.id_token`)

## Problem
The name from Google OAuth wasn't being reliably stored because:
1. No explicit profile callback meant relying on NextAuth defaults
2. `createUser` only stored `name`, not `image`
3. `updateUser` would overwrite all metadata with an empty object if name wasn't provided in the update

## Test plan
- [ ] Sign in with Google and verify `name` is stored in `auth.users.raw_user_meta_data`
- [ ] Verify `avatar_url` is also stored in the same metadata
- [ ] Verify subsequent logins don't wipe the existing metadata
- [ ] Verify no unnecessary DB writes occur when updateUser is called without changes

### DIFF
--- a/apps/web/lib/auth.ts
+++ b/apps/web/lib/auth.ts
@@ -60,6 +60,21 @@ async function withTimeout<T>(
   });
 }
 
+/**
+ * Google OIDC ID token claims. Profile claims (name, picture, etc.) are included
+ * when the `profile` scope is requested.
+ * @see https://developers.google.com/identity/openid-connect/openid-connect#obtainuserinfo
+ */
+interface GoogleProfile {
+  sub: string;
+  email: string;
+  email_verified?: boolean;
+  name?: string;
+  given_name?: string;
+  family_name?: string;
+  picture?: string;
+}
+
 const ProviderConfig = {
   google: {
     clientId: env.GOOGLE_CLIENT_ID!,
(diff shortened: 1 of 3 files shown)


apps/web/lib/auth.ts
```diff
@@ -60,6 +60,16 @@ async function withTimeout<T>(
   });
 }
 
+interface GoogleProfile {
```
alecf, line 68: would be nice just to know where this came from (is there a URL reference?)

alecf (approved):

akhileshrangani4:
‎

apps/web/lib/auth.ts
akhileshrangani4: added it.

akhileshrangani4: comes from here: https://developers.google.com/identity/openid-connect/openid-connect#obtainuserinfo

alecf: I meant put it in the code :)

alecf: Oh I see it

akhileshrangani4: haha, I know, I put it in as well

PR #2031, opened 2026-01-29
Feat(template): add nextjs-tambo-prisma-sqlite starter template

Joystonm:
This PR adds a minimal, CLI-first starter template demonstrating Next.js + Tambo + Prisma (SQLite) with zero manual setup.

The goal of this template is to provide a simple, reproducible foundation that works immediately after install, without dashboards, credentials, or interactive configuration.
nds.

https://github.com/user-attachments/assets/09a25ca2-5c0e-4c26-9778-12e970ccb681


michaelmagan:
Thanks for this. The Prisma + SQLite angle is interesting, but the design doesn't meet our standards. Compare yours to the [official template](https://github.com/tambo-ai/tambo-template) and you'll see the gap.

We also already have a Next.js starter, so the main value here would be the database integration. If you wanted to revisit this, it'd be cool to see Prisma configured to use SQLite locally but with clear instructions for swapping to Postgres/MySQL for production. That'd be genuinely useful. But the design needs to match our templates first.

PR #2082, opened 2026-01-30
feat(templates): add generative-bento starter

tosif121:
https://youtu.be/1QJjMeEqjpA


michaelmagan:
Hey, thanks for submitting this. I see there's a YouTube video, but we also need screenshots in the PR.

Beyond that, looking at the video, the design doesn't quite meet the bar we're looking for in Tambo templates. We want templates that developers would be excited to start from.

Our hackathons are a great place to experiment and build with Tambo if you're interested.

PR #2048, opened 2026-01-29
[REVISION] Tambo AI Template : Nextjs , Google Oauth (Authjs) , Prisma ORM , Postgresql

vishalkumargeed:
# This is a revision of #1995 
# Template for Tambo AI (Nextjs )
I kept in mind the idea that : 
- For a general developer they freshly need to create a next application, add authentication , add DB 
- I somwhat make their path easy by creating this template 

# Important Links 
- DEMO VIDEO : https://youtu.be/K--3X_1cXqM

-  Screenshot : 
image

# Recently you gave the feedback 
image

I guess I implemented this ! 
- Looks somwhat similar to the exisitng tambo template 
- reduced the documentation and unecassry ui changes 

# love to hear another feedback

PR #1995, opened 2026-01-28
Tambo AI Template : Nextjs , Google Oauth (Authjs) , Prisma ORM , Postgresql

vishalkumargeed:
# Template for Tambo AI (Nextjs )
I kept in mind the idea that : 
- For a general developer they freshly need to create a next application, add authentication , add DB 
- I somwhat make their path easy by creating this template 

# Important Links 
- DEMO VIDEO : https://youtu.be/DEHzz5L-Gnc
- Live link : https://tambo-nextjs-starter-template.vercel.app/
This is a deployed template , I thought if this template is not even deployable then what's the point of even creating a template . I need to target even the rookie developer. 
-  Screenshot : 
Screenshot 2026-01-28 224757

Checkout the readme.md file of this repo. 

# Things I kept in mind 
- A template is something that's not need to be complex by proving many features or confuse the suer/ developer . 
- Implemented just the core : GOOGLE OAUTH, PRISMA+ POSTGRES 
- Tambo 2 components : generative and interactive , even pass the authentication state so that tambo can deal with that ! 
- Commented out the code wherever possible .
- Took the reference from the apps/web application for the frontend (Landing Page). I though If it's tambo's template then it should looks alike TAMBO ! 
I hope this is the correct appraoch , if there need to be any change please let me know !

PR #2101, opened 2026-01-31
[REVISION] Tambo AI Template : Nextjs , Google Oauth (Authjs) , Prisma ORM , Postgresql

vishalkumargeed:
# This is a revision of #1995  and #2048 
# Template for Tambo AI (Nextjs )
I kept in mind the idea that : 
- For a general developer they freshly need to create a next application, add authentication , add DB 
- I somwhat make their path easy by creating this template 

# Important Links 
- DEMO VIDEO : https://youtu.be/K--3X_1cXqM

-  Screenshot : 
image

# Previous Feedback #2048 
image

- Changes the API naming convention and erased the prefix ```NEXT_PUBLIC``` . (I learned about this btw, thanks I can now have a better picture to a project )
- Added API auth check : checking authenticated user using the session data. Corrected the flow so that no unauthorized user can see or addUser into the database. 
- File name changed : ```proxy.ts``` to ```middlware.ts``` . Although from next.js 16 version they are following this naming convention (after that client side issue came in 3rd of dec I guess ) for better routing, redirects. Functionality yet remains the same. 
- Authentication Layer : I'm using ```Auth.js``` . I guess there might be some kind of confusion. As in the official documentation of auth.js https://authjs.dev/getting-started/installation, it installs their package and dependencies with a name "next-auth@beta" which is based on v5 . But I'm using earlier version which is stable "v4" and because of this there is no term "beta" in  my package.json file ! 
Plus, there is no dedicated package named "authjs". they integrated with nextjs and thus named "next-auth" and  you can check at signIn ,authjs or any other file where authentication is taking part I'm importing from "next-auth" and "@auth/core" (which is again a dependency package of next-auth) .

### DIFF
--- a/community/templates/tambo-nextjs-fullstack-starter/app/api/env-check/route.ts
+++ b/community/templates/tambo-nextjs-fullstack-starter/app/api/env-check/route.ts
@@ -1,7 +1,7 @@
 import { NextResponse } from "next/server";
 
 const ENV_KEYS = [
-  "TAMBO_API_KEY",
+  "NEXT_PUBLIC_TAMBO_API_KEY",
   "GOOGLE_CLIENT_ID",
   "GOOGLE_CLIENT_SECRET",
   "AUTH_SECRET",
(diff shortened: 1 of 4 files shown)


michaelmagan:
Hey, appreciate the effort here - this is your third revision and you've clearly put work into it. The blocker is that it modifies the root package-lock.json file, which we can't accept for community templates. Templates should only touch files in their own directory. The fullstack pattern with Google OAuth and Prisma is ambitious though. Seems like a great base for a hackathon project - looking forward to seeing what you build with it.

PR #2168, opened 2026-02-03
fix(react-sdk): add .js extensions to ESM imports for strict ESM compatibility

akhileshrangani4:
## Summary
- Add explicit `.js` extensions to all 508 relative imports across 135 source files
- Update `tsconfig.json` and `tsconfig.esm.json` to use `NodeNext` module resolution
- Update `jest.config.ts` with `moduleNameMapper` to handle `.js` imports in tests

This fixes compatibility with strict ESM environments like Astro SSR and Node.js ESM mode, which require explicit file extensions for relative imports.

## Test plan
- [x] All 982 tests pass
- [ ] Verify ESM build output has `.js` extensions
- [ ] Test in Astro SSR environment

Fixes TAM-1025


akhileshrangani4:
Replaced with a cleaner approach using tsc-esm-fix

PR #2208, opened 2026-02-05
fix(web): persist Google refresh token in JWT for automatic token refresh

alecf:
## Summary

- Fix automatic token refresh for Google OAuth by persisting the refresh token in the JWT
- Previously, refresh tokens were only available during initial sign-in and couldn't be accessed on subsequent page loads
- Now tokens are stored in the JWT, enabling automatic refresh when users return after the ID token has expired (~1 hour)

## Problem

Google ID tokens expire after 1 hour. The refresh logic in `refreshTokenIfNecessary()` existed but couldn't access the refresh token on subsequent page loads because:
1. The `account` object (containing `refresh_token`) is only populated during initial sign-in
2. The refresh token wasn't being persisted in the JWT
3. On subsequent visits, `account` was `null`, so the refresh code returned early

## Solution

1. Store `account.refresh_token` in `token.refreshToken` during initial sign-in
2. Update `refreshTokenIfNecessary()` to read from `token.refreshToken` instead of `account?.refresh_token`
3. Add `refreshToken` to the JWT TypeScript interface

## Test plan

- [ ] Log out and log back in with Google
- [ ] Wait for ID token to expire (~1 hour) or manually test with shorter expiry
- [ ] Refresh the page and verify the smoketest still works (no 400 errors on token exchange)

**Note**: Existing logged-in users will need to log out and back in to get a JWT with the refresh token stored.

### DIFF
--- a/apps/web/lib/auth.test.ts
+++ b/apps/web/lib/auth.test.ts
@@ -1,0 +1,211 @@
+// Mock all external dependencies BEFORE any imports that might use them
+// This is necessary because auth.ts has transitive dependencies on modules
+// that use ESM features not supported by Jest
+
+jest.mock("@/lib/env", () => ({
+  env: {
+    GOOGLE_CLIENT_ID: "test-google-client-id",
+    GOOGLE_CLIENT_SECRET: "test-google-client-secret",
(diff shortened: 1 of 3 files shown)


michaelmagan:

PR #2288, opened 2026-02-07
 fix: persist previousRunId across page reloads for v1 threads 

MichaelMilstead:
- Store lastCompletedRunId in ThreadState when a run finishes (RUN_FINISHED event) and when thread metadata is fetched from the API
  - Fetch thread metadata (client.threads.retrieve()) in ThreadSyncManager alongside messages, storing lastCompletedRunId for reloaded threads
  - Fall back to lastCompletedRunId in useTamboV1SendMessage when streaming.runId is absent (post-reload scenario)

  Problem: After a page reload, switching to an existing thread and sending a message failed with BadRequestError: 400 previousRunId is required. The streaming.runId was only set during active streaming and lost on reload. The server requires
  previousRunId when a thread already has messages/runs.
  
  fixes: [TAM-1152](https://linear.app/tambo/issue/TAM-1152/v1-react-sdk-cannot-send-messages-after-reloading-page-and-switching)

### DIFF
--- a/react-sdk/src/v1/providers/tambo-v1-stream-context.tsx
+++ b/react-sdk/src/v1/providers/tambo-v1-stream-context.tsx
@@ -277,10 +277,19 @@ function ThreadSyncManager(): null {
     refetchOnWindowFocus: false,
   });
 
-  const { data: threadData, isSuccess: threadSuccess } = useTamboQuery({
+  useTamboQuery({
     queryKey: ["v1-thread-metadata", currentThreadId],
-    queryFn: async () =>
-      await client.threads.retrieve(currentThreadId, { userKey }),
+    queryFn: async () => {
+      const data = await client.threads.retrieve(currentThreadId, { userKey });
+      if (data.lastCompletedRunId && dispatch) {
+        dispatch({
+          type: "SET_LAST_COMPLETED_RUN_ID",
+          threadId: currentThreadId,
+          lastCompletedRunId: data.lastCompletedRunId,
+        });
+      }
+      return data;
+    },
     enabled: shouldFetch,
     staleTime: 1000,
     refetchOnWindowFocus: false,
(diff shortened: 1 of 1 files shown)


react-sdk/src/v1/providers/tambo-v1-stream-context.tsx
```diff
@@ -288,7 +296,22 @@ function ThreadSyncManager(): null {
...
+    if (!threadData.lastCompletedRunId) return;
+
+    dispatch({
+      type: "LOAD_THREAD_MESSAGES",
+      threadId: currentThreadId,
+      messages: [],
+      skipIfStreaming: true,
+      lastCompletedRunId: threadData.lastCompletedRunId,
+    });
+  }, [threadSuccess, threadData, currentThreadId, dispatch]);
```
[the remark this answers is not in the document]
MichaelMilstead: updated to dedicated action

react-sdk/src/v1/providers/tambo-v1-stream-context.tsx
```diff
@@ -288,7 +299,20 @@ function ThreadSyncManager(): null {
     });
 
     lastSyncedThreadRef.current = currentThreadId;
-  }, [isSuccess, messagesData, currentThreadId, dispatch]);
+  }, [messagesSuccess, messagesData, currentThreadId, dispatch]);
+
+  // Sync thread metadata (lastCompletedRunId) separately so messages
+  // aren't blocked if the thread retrieve is slower or fails
+  useEffect(() => {
+    if (!threadSuccess || !threadData || !dispatch) return;
```
alecf: I'm thinking this should be done in the onSuccess for the useTamboQuery for the thread request

MichaelMilstead: moved into the queryfn since apparently `onSuccess` has been removed from react-query? I'm good with going back to useeffect if this feels awkward though

PR #2388, opened 2026-02-12
fix(api): fix potential SQL injection in transaction middleware

RinZ27:
Switching to parameterized queries for the `set_config` call in the transaction middleware. The previous implementation used `sql.raw` to interpolate the `projectId`, which is derived from API keys. 

While these IDs are currently system-generated, I noticed using `sql.raw` here is an anti-pattern that could lead to SQL injection if input constraints change or custom identifiers are introduced. This update uses standard Drizzle parameter binding to ensure the query is executed safely. I also kept the `sql.raw` for the role name since it's a schema constant and typically requires an identifier rather than a parameter in Postgres.


MichaelMilstead:
Closing since this is unused code and should be removed, but thank you for the bump :)

PR #2417, opened 2026-02-15
docs: fix spacing in README a tags

coder-irfan:
Fixed space in README
Corrected formatting: removed extra space in README so text aligns properly for "a" tags.


MichaelMilstead:
Looks like you just need to fix the PR title to use something like "fix: ..."

PR #2435, opened 2026-02-18
fix(react-ui-base): render content blocks in natural order from model response

akhileshrangani4:
## Summary
- Fixes text after a tool call rendering above the tool call UI when using thinking models
- With thinking models, the content array is `[tool_use, text]` but thread-content hardcoded render order as `MessageContent → ToolcallInfo`, pulling all text above all tool calls
- Now iterates `message.content` in order, grouping consecutive text/resource blocks and rendering each `tool_use` block in its actual position
- Adds `toolUse` prop to `ToolcallInfo` / `ToolcallInfoBase.Root` so it can render a specific tool_use block instead of always using the first one

## Changes
- **`react-ui-base/toolcall-info-root.tsx`**: Added `toolUse` prop, uses it to override `getToolCallRequest(message)`, improved `hasToolError` matching by `toolCallRequest.id`
- **`react-ui-base/get-tool-status-message.ts`**: Added optional `toolUseOverride` parameter
- **`ui-registry/message.tsx`**: Added `toolUse` prop to `ToolcallInfoProps`, passes through to base
- **`ui-registry/thread-content.tsx`**: Added `getOrderedContentGroups()` helper, replaced hardcoded `<MessageContent /><ToolcallInfo />` with dynamic ordered rendering

## Test plan
- [x] All 112 react-ui-base tests pass (including toolcall-info and get-tool-status-message tests)
- [x] Full monorepo type checking passes (19/19 packages)
- [x] Lint clean on all changed files
- [x] Manual test with thinking model (e.g. Claude with extended thinking) — verify text after tool call renders below tool call UI
- [x] Manual test with non-thinking model — verify no regression (text before tool call still renders correctly)

Fixes TAM-1180

### DIFF
--- a/packages/ui-registry/src/components/thread-content/thread-content.tsx
+++ b/packages/ui-registry/src/components/thread-content/thread-content.tsx
@@ -10,7 +10,11 @@ import {
   type messageVariants,
 } from "@tambo-ai/ui-registry/components/message";
 import { cn } from "@tambo-ai/ui-registry/utils";
-import { type TamboThreadMessage, useTambo } from "@tambo-ai/react";
+import {
+  type Content,
+  type TamboThreadMessage,
+  useTambo,
+} from "@tambo-ai/react";
 import { type VariantProps } from "class-variance-authority";
 import * as React from "react";
 
@@ -179,24 +183,37 @@ const ThreadContentMessages = React.forwardRef<
                 <ReasoningInfo />
                 <MessageImages />
                 {message.content.map((block, blockIndex) => {
-                  if (block.type === "text" || block.type === "resource") {
-                    return (
-                      <MessageContent
-                        key={`content-${blockIndex}`}
-                        content={[block]}
-                        className={messageContentClassName}
-                      />
-                    );
+                  switch (block.type) {
+                    case "text":
+                    case "resource":
+                      return (
+                        <MessageContent
+                          key={`content-${blockIndex}`}
+                          content={[block]}
+                          className={messageContentClassName}
+                        />
+                      );
+                    case "tool_use":
+                      return (
+                        <ToolcallInfo
+                          key={`tool-${block.id ?? blockIndex}`}
+                          toolUse={block}
+                        />
+                      );
+                    case "tool_result":
+                    case "component":
+                      // tool_result is rendered by ToolcallInfo on the preceding assistant message.
+                      // component is rendered by MessageRenderedComponentArea below.
+                      return null;
+                    default: {
+                      const _exhaustive: never = block;
+                      console.error(
+                        "Unknown content block type:",
+                        (_exhaustive as Content).type,
+                      );
+                      return null;
+                    }
                   }
-                  if (block.type === "tool_use") {
-                    return (
-                      <ToolcallInfo
-                        key={`tool-${block.id ?? blockIndex}`}
-                        toolUse={block}
-                      />
-                    );
-                  }
-                  return null;
                 })}
                 <MessageRenderedComponentArea className="w-full" />
               </div>


packages/ui-registry/src/components/thread-content/thread-content.tsx
```diff
@@ -173,14 +221,30 @@ const ThreadContentMessages = React.forwardRef<
...
+                  groups.map((group, idx) => {
+                    if (group.type === "text") {
+                      return (
+                        <MessageContent
+                          key={`text-${idx}`}
+                          content={group.content}
+                          className={messageContentClassName}
+                        />
+                      );
+                    }
```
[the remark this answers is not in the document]
akhileshrangani4: Fixed in cba5e44 — `ContentGroup` now carries `startIndex` (the position of the first block in the original `message.content` array) and keys use `text-${group.startIndex}` / `tool-${group.toolUse.id ?? group.startIndex}` instead of the group iteration index.

packages/ui-registry/src/components/thread-content/thread-content.tsx
```diff
@@ -102,6 +107,43 @@ const ThreadContent = React.forwardRef<HTMLDivElement, ThreadContentProps>(
...
+    } else {
+      if (currentTextGroup.length > 0) {
+        groups.push({ type: "text", content: [...currentTextGroup] });
+        currentTextGroup = [];
+      }
+      if (block.type === "tool_use") {
+        groups.push({ type: "tool_use", toolUse: block });
+      }
+      // tool_result, component, image_url handled by other components
+    }
```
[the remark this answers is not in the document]
akhileshrangani4: This is intentional and not a correctness risk. The skipped block types are all position-independent:

- **`tool_result`**: Consumed by `ToolcallInfo` on the preceding assistant message via `associatedToolResponse` lookup in `toolcall-info-root.tsx`. These are never rendered as standalone elements — we even filter out messages that only contain `tool_result` blocks (line 175-180).
- **`component`**: Rendered by `MessageRenderedComponentArea` which reads all component blocks from the message context. Position within the content array does not matter.
- **`image_url`**: Rendered by `MessageImages` which similarly reads from message context regardless of position.

Expanding `ContentGroup` to cover these types would duplicate rendering logic that already works correctly via dedicated components, adding complexity without benefit. Added a clarifying comment in the latest commit explaining this.

packages/react-ui-base/src/toolcall-info/root/get-tool-status-message.ts
```diff
@@ -1,18 +1,20 @@
...
 
 /**
  * Gets the status message for a tool call.
  * Returns the custom status message if available, otherwise generates a default one.
  * @param message - The thread message containing the tool call
  * @param isLoading - Whether the tool call is still in progress
+ * @param toolUseOverride - Optional specific tool_use block. If not provided, uses the first from message.
  * @returns The status message string, or null if not a tool call message
  */
 export function getToolStatusMessage(
```
alecf, line 11: @CharlieHelps this function got a new optional parameter, but it is actually only used once outside of tests and it uses this parameter which is actually derived from getToolCallRequest. Can you take a step back and just look at the intended usecase and make this function as simple as possible instead of trying to make the existing behavior more flexible? Simplify ruthlessly

packages/ui-registry/src/components/thread-content/thread-content.tsx
```diff
@@ -102,6 +107,58 @@ const ThreadContent = React.forwardRef<HTMLDivElement, ThreadContentProps>(
 );
 ThreadContent.displayName = "ThreadContent";
 
+/**
+ * Groups content blocks by type in their natural order.
+ * Consecutive text/resource blocks are merged into a single group.
+ * Each tool_use block gets its own group.
+ * This preserves the actual order from the model response, which matters
+ * when thinking models produce [tool_use, text] instead of [text, tool_use].
+ */
```
alecf: I don't actually think we should do things this way - I think we should be rendering the content as it appears in the message without any regrouping. so if the message had text-tool-text then we would render those in that order. But this is in the shared component, so not sure what others thing - @lachieh have any thoughts?

akhileshrangani4: Yeah it was grouping consecutive text blocks together into a single `<MessageContent>`, but you're right — we should be rendering the content as it comes in. Removed all the grouping, tested it and it works that way too. Pushed in 44a8f58.

packages/ui-registry/src/components/thread-content/thread-content.tsx
```diff
@@ -147,6 +204,12 @@ const ThreadContentMessages = React.forwardRef<
       {...props}
     >
       {filteredMessages.map((message, index) => {
+        const groups = getOrderedContentGroups(message.content);
+        const messageContentClassName =
+          message.role === "assistant"
+            ? "text-foreground font-sans"
+            : "text-foreground bg-container hover:bg-backdrop font-sans";
+
```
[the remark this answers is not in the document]
akhileshrangani4: Moot after 44a8f58 — the grouping function was removed entirely. Content blocks are now rendered directly from `message.content.map()`.

packages/ui-registry/src/components/thread-content/thread-content.tsx
```diff
@@ -173,14 +236,30 @@ const ThreadContentMessages = React.forwardRef<
...
+                      return (
+                        <ToolcallInfo
+                          key={`tool-${group.toolUse.id ?? group.startIndex}`}
+                          toolUse={group.toolUse}
+                        />
+                      );
+                    }
+                    return null;
+                  })
+                )}
```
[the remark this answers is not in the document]
akhileshrangani4: Moot after 44a8f58 — removed the grouping and fallback entirely. Now just iterates `message.content` directly.

packages/ui-registry/src/components/thread-content/thread-content.tsx
```diff
@@ -173,14 +178,26 @@ const ThreadContentMessages = React.forwardRef<
...
+                    return (
+                      <ToolcallInfo
+                        key={`tool-${block.id ?? blockIndex}`}
+                        toolUse={block}
+                      />
+                    );
                   }
-                />
-                <ToolcallInfo />
+                  return null;
```
alecf: images are covered by resource, right?

If possible lets use a switch statement so we can use an exhaustive switch on `block.type` and log an error if we hit an unknown block type

akhileshrangani4: Yep, images are covered by resource. Switched to an exhaustive switch in ccff821 — handles `text`/`resource` → `MessageContent`, `tool_use` → `ToolcallInfo`, explicitly skips `tool_result`/`component` with a comment, and the `default` branch does `const _exhaustive: never = block` for compile-time safety + `console.error` at runtime.

PR #2451, opened 2026-02-20
deps: remove gh CLI from mise.toml

akhileshrangani4:
## Summary
- Removes `gh` from `mise.toml` to unblock Railway API deployments
- The mise aqua backend fails attestation verification for all recent gh versions, causing build failures
- `gh` is only used for dev workflows (PRs, issues) and is not needed for app builds — developers still have it via system install

Fixes TAM-1222

## Test plan
- [ ] Verify Railway API build succeeds

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,6 +1,5 @@
 [tools]
 cspell = "9.4.0"
-gh = "2.87.2"
 jq = "1.8.1"
 shellcheck = "0.11.0"
 


lachieh:
I think this is actually an issue with railway using an old version of mise:
The attestation issue was resolved here: https://github.com/jdx/mise/pull/7853

akhileshrangani4:
@lachieh ohh, okay, then I think this is not needed. closing.

PR #2476, opened 2026-02-24
chore(docs): Add Chinese README and a language selection link to the main README.

SXP-Simon:


lachieh:
We don't currently have any Chinese speakers on the team. While I'd love to accept this, there's no way for us to verify. If we add translation capabilities, we'll re-open and review this.

Thanks for the contribution!

PR #2495, opened 2026-02-25
fix(ci): Try pinning mise back a few versions

alecf:
hoping this fixes the build - this version was passing a few days ago


alecf:
that wasn't it!

PR #2553, opened 2026-03-04
refactor(plugins): consolidate tambo skills into two entry points

akhileshrangani4:
## Summary
- keep only two user-facing plugin skills: `generative-ui` and `build-with-tambo`
- convert former standalone skills into reference guides under `plugins/tambo/skills/build-with-tambo/references/`
- update both remaining skills to link those reference guides
- update plugin README to document only the two public skills
- remove obsolete standalone skill invocation examples from references and clarify wording

## Notes
- this simplifies skill selection while preserving all guidance as load-on-demand resources

### DIFF
--- a/plugins/tambo/skills/build-with-tambo/SKILL.md
+++ b/plugins/tambo/skills/build-with-tambo/SKILL.md
@@ -11,12 +11,14 @@ Detect tech stack and integrate Tambo while preserving existing patterns.
 
 Use these guides when you need deeper implementation details for a specific area:
 
-- [Components](references/components.md)
-- [Component Rendering](references/component-rendering.md)
-- [Threads and Input](references/threads.md)
-- [Tools and Context](references/tools-and-context.md)
-- [CLI Reference](references/cli.md)
-- [Add Components to Registry](references/add-components-to-registry.md)
+- [Components](references/components.md) - Create and register Tambo components (generative and interactable).
+- [Component Rendering](references/component-rendering.md) - Handle streaming props, loading states, and persistent component state.
+- [Threads and Input](references/threads.md) - Manage conversations, suggestions, voice input, image attachments, and thread switching.
+- [Tools and Context](references/tools-and-context.md) - Add custom tools, MCP servers, context helpers, and resources.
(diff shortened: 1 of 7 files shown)


plugins/tambo/skills/generative-ui/SKILL.md
```diff
@@ -1,12 +1,22 @@
...
 
 # Generative UI
 
 Build generative UI apps with Tambo — create rich, interactive React components from natural language.
 
+## Reference Guides
+
+For deeper implementation details beyond bootstrap flow, load:
+
+- [components](../build-with-tambo/references/components.md)
```
alecf: if you're linking relatively, it means this skill won't work without the other skill. I don't know if there are best practices for sharing, or if we need to just copy content between the two?

plugins/tambo/skills/build-with-tambo/SKILL.md
```diff
@@ -1,12 +1,23 @@
...
-# Add Tambo to Existing Project
+# Build with Tambo
 
 Detect tech stack and integrate Tambo while preserving existing patterns.
 
+## Reference Guides
+
+Use these guides when you need deeper implementation details for a specific area:
+
+- [Components](references/components.md)
```
alecf: I think add the old frontmatter descriptions to each item in this bulleted list

alecf (changes requested):

plugins/tambo/skills/build-with-tambo/SKILL.md
akhileshrangani4: yup, added that now.

plugins/tambo/skills/generative-ui/SKILL.md
akhileshrangani4: good call. i switched generative-ui to local references and copied the shared docs into generative-ui/references. from what i found, best practice is each skill should be self-contained and not rely on another skill’s folder, so this keeps it portable.

akhileshrangani4: also, if the duplicated docs start drifting, i can add a tiny sync script to keep both reference folders in sync.

PR #2499, opened 2026-02-25
refactor(react-ui-base): harden useRender state/props split across primitives

lachieh:

PR #2582, opened 2026-03-06
fix(react-ui-base): align MessageInput.StopButton with useRender state/props pattern

Logesh-waran2003:
## Summary

- `MessageInput.StopButton` was the only component missed in the #2499 hardening refactor that moved `slot` from props into state across all message-input primitives
- Adds `slot: "message-input-stop"` to state (was missing entirely, so `data-slot` was never rendered on the DOM)
- Extends state interface with `Record<string, unknown>` to match every other primitive
- Adds `aria-hidden` when hidden, consistent with `SubmitButton` behaviour
- Adds `type: "button"` explicitly to props
- Adds `keepMounted` test to match `SubmitButton` test coverage

## Test plan

- [x] All existing stop button tests pass
- [x] New `keepMounted` test verifies `data-state="hidden"` and `aria-hidden="true"` when idle
- [x] `npm run lint` — clean
- [x] `npm run check-types` — clean
- [x] `npm test -w packages/react-ui-base` — 10/10 pass

### DIFF
--- a/packages/react-ui-base/src/message-input/message-input-controls.test.tsx
+++ b/packages/react-ui-base/src/message-input/message-input-controls.test.tsx
@@ -159,6 +159,7 @@ describe("MessageInput controls", () => {
     }
     expect(button.getAttribute("data-state")).toBe("hidden");
     expect(button.getAttribute("aria-hidden")).toBe("true");
+    expect(button.getAttribute("tabindex")).toBe("-1");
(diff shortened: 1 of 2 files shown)


lachieh:
Thanks, @Logesh-waran2003! Making a small addition with tabIndex while were in there and we'll get this in the next release!

PR #2610, opened 2026-03-11
fix(react-sdk): clean up interactable tools on component unmount

akhileshrangani4:
## Summary
- Fix race condition where interactable component tools persisted in the tool registry after navigating away from a page, causing the model to call stale tools
- Add `unregisterTools` (internal) to the registry provider so tools can be removed
- Add project context helper to the web dashboard so the model knows which project the user is viewing
- Add vite temp files to .gitignore

## Why
When a user navigated from a project settings page (with interactable components) to the home page, the per-component tools (`update_component_props_*`, `update_component_state_*`) and global interactable tools remained registered. These stale tools were sent to the model with every message, confusing it into trying to interact with components that no longer existed on the page.

## Changes
- **`withTamboInteractable` HOC** — calls `removeInteractableComponent` on unmount
- **`removeInteractableComponent`** — now also unregisters per-component tools from the registry
- **`clearAllInteractableComponents`** — same cleanup for bulk removal
- **Global interactable tools** — unregistered when the interactable component list becomes empty
- **`unregisterTools`** — new internal method on registry provider (not exposed in public API)
- **Project context helper** — new context helper for the web dashboard that provides project list and current project ID
- **`.gitignore`** — add `vite.config.ts.timestamp-*` pattern

## Test Plan
- [x] Existing tests pass (75 tests across affected files)
- [x] New tests for `unregisterTools` (single, multiple, non-existent names)
- [x] New tests for tool cleanup on `removeInteractableComponent` and `clearAllInteractableComponents`
- [x] New test for HOC unmount calling `removeInteractableComponent`
- [x] Type checks pass for react-sdk and apps/web

### DIFF
--- a/apps/web/components/ui/tambo/message-thread-panel.tsx
+++ b/apps/web/components/ui/tambo/message-thread-panel.tsx
@@ -1,14 +1,13 @@
 "use client";
 
 import { MessageInputTextareaWithInteractables } from "@/components/ui/tambo/message-input-with-interactables";
-import { createProjectContextHelper } from "@/lib/tambo/context-helpers/project-context-helper";
+import { useProjectContextHelper } from "@/lib/tambo/context-helpers/project-context-helper";
 import { registerAllTools } from "@/lib/tambo/tools/tool-registry";
 import { cn } from "@/lib/utils";
 import { useMessageThreadPanel } from "@/providers/message-thread-panel-provider";
 import { api, useTRPCClient } from "@/trpc/react";
 import {
   useTambo,
-  useTamboContextHelpers,
   type Suggestion,
   type TamboThreadMessage,
 } from "@tambo-ai/react";
@@ -174,8 +173,6 @@ export const MessageThreadPanel = forwardRef<
   const { registerTool } = useTambo();
   const trpcClient = useTRPCClient();
   const utils = api.useUtils();
-  const { addContextHelper } = useTamboContextHelpers();
-
   /**
    * Registers all tambo tools with the thread.
    * This effect runs once when the component mounts and registers tools for tambo
@@ -185,13 +182,7 @@ export const MessageThreadPanel = forwardRef<
     registerAllTools(registerTool, { trpcClient, utils });
   }, [registerTool, trpcClient, utils]);
 
-  /**
-   * Registers a context helper that provides the user's projects
-   * and identifies which project they're currently viewing.
-   */
-  useEffect(() => {
-    addContextHelper("userProjects", createProjectContextHelper(trpcClient));
-  }, [addContextHelper, trpcClient]);
+  useProjectContextHelper();
 
   const { data: session } = useSession();
   const isUserLoggedIn = !!session;
--- a/apps/web/lib/tambo/context-helpers/project-context-helper.ts
+++ b/apps/web/lib/tambo/context-helpers/project-context-helper.ts
@@ -1,4 +1,8 @@
-import type { useTRPCClient } from "@/trpc/react";
+"use client";
+
+import { api } from "@/trpc/react";
+import { useTamboContextHelpers } from "@tambo-ai/react";
+import { useCallback, useEffect } from "react";
 
 /**
  * Extracts the project ID from the current URL pathname.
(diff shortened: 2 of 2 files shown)


react-sdk/src/hoc/with-tambo-interactable.tsx
```diff
@@ -151,10 +152,17 @@ export function withTamboInteractable<ComponentProps extends object>(
...
       registerComponent();
-    }, [registerComponent]);
+
+      return () => {
+        if (isInitialized.current && interactableId) {
+          removeInteractableComponent(interactableId);
+          isInitialized.current = false;
+        }
+      };
+    }, [registerComponent, interactableId, removeInteractableComponent]);
```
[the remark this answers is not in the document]
akhileshrangani4: good catch, split it into two effects now so cleanup only fires on unmount. registration still re-runs on prop changes as before, but cleanup is keyed on the stable interactableId only.

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -418,8 +435,15 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
...
-  }, []);
+    setInteractableComponents((prev) => {
+      const toolNames = prev.flatMap((c) => [
+        `update_component_props_${c.id}`,
+        `update_component_state_${c.id}`,
+      ]);
+      unregisterTools(toolNames);
+      return [];
+    });
+  }, [unregisterTools]);
```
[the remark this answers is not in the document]
akhileshrangani4: yep, moved unregisterTools out of the updater, now reads interactableComponents directly, unregisters tools, then sets state to []. no more side effects inside the updater.

react-sdk/src/hoc/with-tambo-interactable.tsx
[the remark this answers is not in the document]
akhileshrangani4: fixed, same approach you suggested, split into a mount-only registration effect and a separate cleanup effect keyed on stable values. no more churn on prop changes.

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -397,9 +407,16 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
     ],
   );
 
-  const removeInteractableComponent = useCallback((id: string) => {
-    setInteractableComponents((prev) => prev.filter((c) => c.id !== id));
-  }, []);
+  const removeInteractableComponent = useCallback(
```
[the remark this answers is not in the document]
akhileshrangani4: nice catch, added unregisterTools in the tool handler too. didn't use the helper directly since it's defined after the useEffect and adding it to deps would cause a re-registration loop, but the tool handler now cleans up the same way.

apps/web/components/ui/tambo/message-thread-panel.tsx
```diff
@@ -182,6 +185,14 @@ export const MessageThreadPanel = forwardRef<
     registerAllTools(registerTool, { trpcClient, utils });
   }, [registerTool, trpcClient, utils]);
 
+  /**
+   * Registers a context helper that provides the user's projects
+   * and identifies which project they're currently viewing.
+   */
+  useEffect(() => {
+    addContextHelper("userProjects", createProjectContextHelper(trpcClient));
+  }, [addContextHelper, trpcClient]);
```
alecf: isn't this where we'd want to unregister (in the useEffect cleanup)

apps/web/lib/tambo/context-helpers/project-context-helper.ts
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @param trpcClient - The tRPC client for fetching project data
+ * @returns An async context helper function
+ */
+export function createProjectContextHelper(
+  trpcClient: ReturnType<typeof useTRPCClient>,
+) {
+  return async (): Promise<ProjectContextData | null> => {
+    try {
+      const projects = await trpcClient.project.getUserProjects.query();
```
alecf: minor bug I do feel like this should somehow be done with a hook. like at a high level someting like:

```ts
function useProjectContextHelper() {
    const { data: projects } = trpc.project.getUserProjects.useQuery();
    return useCallback(() => {
       ...
       return {
          currentProjectId,
          ...
       }
    }, [...]);
}
```

and then in the register call you can say

```
const createProjectCotnext = useProjectContextHelper();
useEffect(...)
...
```

that will keep react lifecycles sane

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -152,6 +159,10 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
             } as const;
           }
 
+          unregisterTools([
```
alecf: I'm concerned that this is being hardcoded - is there any place you can get the list of tools that were previously registered?

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -418,8 +439,13 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
   );
 
   const clearAllInteractableComponents = useCallback(() => {
+    const toolNames = interactableComponents.flatMap((c) => [
+      `update_component_props_${c.id}`,
```
alecf: so this pattern has now repeated like 3 times - I just feel like this should be accessible from the registry or something

apps/web/components/ui/tambo/message-thread-panel.tsx
akhileshrangani4: I am intentionally not cleaning up here, the assistant should always have project context available so it does not hallucinate when the user asks "create api key for x project" while being in the y project, so we don't want to unregister it on unmount

react-sdk/src/providers/tambo-interactable-provider.tsx
akhileshrangani4: extracted a `perComponentToolNames()` helper for now so it's at least in one place. toolRegistry is already exposed on the registry context so we could filter `Object.keys(toolRegistry)` by component ID. but that couples to naming conventions at runtime instead of at definition time. for a proper follow-up we could add something like `getToolsByPrefix` or `getToolsForComponent` to the registry so the interactable provider doesn't need to know the naming convention at all

react-sdk/src/providers/tambo-interactable-provider.tsx
akhileshrangani4: yep, consolidated into `perComponentToolNames()` so at least it's not repeated. agree a registry-level API for this would be cleaner, will do that as a follow-up

apps/web/lib/tambo/context-helpers/project-context-helper.ts
akhileshrangani4: refactored to a hook. `useProjectContextHelper` now uses useQuery internally and self-registers with addContextHelper, so the consumer just calls the hook.

PR #2620, opened 2026-03-11
refactor(react-sdk): add registry-level tool ownership for interactable components

akhileshrangani4:
## Summary
- Removes hardcoded tool naming conventions (`perComponentToolNames`, `GLOBAL_INTERACTABLE_TOOL_NAMES`) from the interactable provider
- Adds local `registerToolForComponent` and `unregisterToolsForComponent` helpers inside the interactable provider for ownership-based cleanup
- Registry provider stays fully generic with no ownership concepts
- Follow-up to #2610 — see Alec's PR comment about decoupling tool naming from the interactable provider

## Why
The interactable provider previously had to know tool naming conventions to clean up tools on unmount. This couples the provider to implementation details of tool naming. Now it tracks ownership locally via a componentId -> toolNames map, and cleanup is just `unregisterToolsForComponent(id)` without knowing any tool names. The registry provider remains generic, only exposing `registerTool` / `unregisterTools`.

## Test Plan
- [x] All 1039 react-sdk tests pass
- [x] New tests for dedup behavior and global tool function coverage
- [x] `npm run lint` passes (only pre-existing warnings)

### DIFF
--- a/react-sdk/src/providers/tambo-interactable-provider.test.tsx
+++ b/react-sdk/src/providers/tambo-interactable-provider.test.tsx
@@ -141,6 +141,42 @@ describe("TamboInteractableProvider - State Update Tool Registration", () => {
     expect(stateToolCall[0].description).toContain("MyComponent");
   });
 
+  it("should allow props update tool to update component props", () => {
+    const { result } = renderHook(() => useTamboInteractable(), { wrapper });
+
+    const component: Omit<TamboInteractableComponent, "id" | "createdAt"> = {
+      name: "TestComponent",
+      description: "A test component",
+      component: () => <div>Test</div>,
+      props: { title: "original" },
+      propsSchema: z.object({ title: z.string() }),
+    };
+
+    let componentId = "";
+    act(() => {
+      componentId = result.current.addInteractableComponent(component);
+    });
+
+    // Find the props update tool and call it
+    const propsToolCall = mockRegisterTool.mock.calls.find((call) =>
+      call[0].name.startsWith("update_component_props_"),
+    );
+
+    const toolFn = propsToolCall[0].tool;
+
+    let updateResult = "";
+    act(() => {
+      updateResult = toolFn({
+        componentId,
+        newProps: { title: "updated" },
+      });
+    });
+
+    expect(updateResult).toBe("Updated successfully");
+    const comp = result.current.getInteractableComponent(componentId);
+    expect(comp?.props.title).toBe("updated");
+  });
+
   it("should allow state update tool to update multiple state values", () => {
     const { result } = renderHook(() => useTamboInteractable(), { wrapper });
 


react-sdk/src/providers/tambo-registry-provider.tsx
```diff
@@ -235,6 +257,45 @@ export const TamboRegistryProvider: React.FC<
...
+
+      unregisterTools(toolNames);
+      setToolComponentOwnership((prev) => {
+        const next = { ...prev };
+        delete next[componentId];
+        return next;
+      });
+    },
+    [toolComponentOwnership, unregisterTools],
+  );
```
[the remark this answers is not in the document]
akhileshrangani4: good catch - fixed in d5768d23a. `unregisterToolsForComponent` now reads ownership inside the `setToolComponentOwnership` updater so the dep array is just `[unregisterTools]`. no more identity churn, no more loop.

react-sdk/src/providers/tambo-registry-provider.tsx
```diff
@@ -235,6 +257,45 @@ export const TamboRegistryProvider: React.FC<
...
+      setToolRegistry((registry) =>
+        registryWithTool(warnOnOverwrite)(registry, tool),
+      );
+      setToolComponentOwnership((prev) => ({
+        ...prev,
+        [componentId]: [
+          ...(prev[componentId] ?? []),
+          (tool as { name: string }).name,
+        ],
+      }));
```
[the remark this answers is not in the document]
akhileshrangani4: yep, also fixed in the same commit. `registerToolForComponent` now checks `existing.includes(tool.name)` before appending and returns `prev` unchanged if the name is already tracked.

react-sdk/src/providers/tambo-registry-provider.tsx
```diff
@@ -47,6 +47,13 @@ export interface TamboRegistryContext {
   registerTool: RegisterToolFn;
   registerTools: RegisterToolsFn;
   unregisterTools: (names: string[]) => void;
+  registerToolForComponent: (
+    componentId: string,
+    tool: unknown,
```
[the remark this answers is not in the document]
akhileshrangani4: agreed, no reason for this to be `unknown` since its a new function with no overload baggage. changed to `TamboTool` on both the interface and the implementation, cast is gone.

react-sdk/src/providers/tambo-interactable-provider.tsx
```diff
@@ -204,9 +192,13 @@ export const TamboInteractableProvider: React.FC<PropsWithChildren> = ({
...
       // No interactable components — remove the global tools
-      unregisterTools(GLOBAL_INTERACTABLE_TOOL_NAMES);
+      unregisterToolsForComponent(GLOBAL_INTERACTABLE_OWNER);
     }
-  }, [interactableComponents, registerTool, unregisterTools]);
+  }, [
+    interactableComponents,
+    registerToolForComponent,
+    unregisterToolsForComponent,
+  ]);
```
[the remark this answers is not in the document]
akhileshrangani4: yep this is the other side of the same loop. with `unregisterToolsForComponent` now stable (no `toolComponentOwnership` in deps), this effect only re-runs when `interactableComponents` actually changes. all good now.

react-sdk/src/providers/tambo-registry-provider.tsx
```diff
@@ -235,6 +257,42 @@ export const TamboRegistryProvider: React.FC<
     });
   }, []);
 
+  const registerToolForComponent = useCallback(
```
MichaelMilstead: I feel that this interactable-specific logic about a tool "belonging" to a component should not be in the registry provider

akhileshrangani4: Yeah, good question. Right now, interactable components are really the main place where we use this kind of tool ownership pattern, so there may be a cleaner place for it later outside the registry provider. But since the registry already handles the tool lifecycle, putting ownership tracking there felt like the most natural thing for now.

The main thing we wanted to avoid was going back to the old setup where the interactable provider had to manually keep track of tool names through `perComponentToolNames` and `GLOBAL_INTERACTABLE_TOOL_NAMES`, and then build the right names just to call `unregisterTools`. That felt pretty fragile and also tied the interactable provider to tool naming details it really should not need to know about.

With ownership tracking in the registry, cleanup becomes just `unregisterToolsForComponent(id)`, and you do not have to think about names at all. If we find a better place for it later, we can definitely move it, but for now this felt like the simplest way to keep things clean.

MichaelMilstead: Agreed that it's better to store a map of componentId -> toolIds rather than searching for the "expected tool" based on the component's id or name, but I think we can keep that in the interactable provider.

Maybe we just move this new state val and these new functions into the interactable provider, so we get rid of the hardcoded toolnames, and the registryprovider stays generic

akhileshrangani4: yeah that's fair, we can move the ownership map and the three functions into the interactable provider and just have it call registerTool / unregisterTools on the registry directly. keeps the registry generic and the bookkeeping stays close to the only consumer that needs it. will refactor.

akhileshrangani4: @MichaelMilstead done in [c6c5e02](https://github.com/tambo-ai/tambo/pull/2620/commits/c6c5e02008a1d431671d5bfdfeafa7f782c08511). moved the ownership state and the three functions into the interactable provider. registry provider is back to being fully generic, just register/unregister by name.

alecf (approved):

PR #2692, opened 2026-03-27
fix(web, db, core): improve MCP OAuth registration flow

rehmatsg:
## Summary
- improve outbound MCP OAuth registration metadata so remote servers like Mapbox accept Tambo's client registration
- switch MCP OAuth to a stable callback URL with session recovery via OAuth `state`
- add manual client registration fallback in the dashboard and support a hosted client metadata document URL
- stop relying on arbitrary session rows for steady-state MCP OAuth reuse and clear legacy redirect-bound registrations on reauth

## Why this was needed

Tambo’s outbound MCP OAuth registration payload was too minimal for some remote MCP servers.

In practice:
- Notion MCP worked because its authorization server accepted the minimal client metadata Tambo sent.
- Mapbox MCP failed during dynamic client registration because it required `response_types`, which Tambo was not sending.

This change makes Tambo’s MCP OAuth client registration more complete and more compatible with stricter servers by:
- sending a fuller client metadata payload (`response_types`, `grant_types`, `client_uri`, `token_endpoint_auth_method`, `application_type`)
- switching to a stable callback URL with OAuth `state` for session recovery
- adding a manual client registration fallback for servers where automatic registration is unavailable or fails

## Context
This changes Tambo Cloud's outbound MCP OAuth flow to be compatible with stricter remote MCP servers while keeping the existing project-scoped auth behavior.

## Testing
- [x] `corepack npm run lint`
- [x] `corepack npm run check-types -w packages/core`
- [x] `corepack npm run check-types -w packages/db`
- [x] `corepack npm run check-types -w apps/web`
- [x] `corepack npm test -w packages/core -- --runTestsByPath src/mcp-oauth-client.test.ts`
- [x] `corepack npm test -w apps/web -- --runTestsByPath components/dashboard-components/project-details/mcp-server-editor.test.tsx`
- [ ] `corepack npm run check-types`
- [ ] `corepack npm test`

## Notes
Root `check-types` and `test` are currently blocked by unrelated existing failures in `packages/backend/src/services/llm/ai-sdk-client.ts` after the AI SDK v6 upgrade:
- `LanguageModelV3` missing from `@ai-sdk/provider`
- stream event variant mismatches for `tool-approval-request` and `tool-output-denied`

## PR Checklist
- [x] Conventional Commit title
- [x] Linked context
- [x] Tests added/updated
- [ ] Docs updated (not user-facing)
- [ ] Screen recording attached (not applicable)
- [ ] All checks pass locally (blocked by unrelated backend build/type errors)


apps/web/app/oauth/callback/route.ts
```diff
@@ -25,13 +25,25 @@ const callbackParamsSchema = z
...
+    if (validatedParams.error) {
+      return redirectToOAuthError(request, validatedParams.error);
+    }
+
+    const sessionId = validatedParams.state ?? validatedParams.sessionId;
+    if (!sessionId) {
+      throw new Error("OAuth session ID missing from callback state");
+    }
+
+    if (!validatedParams.code) {
```
alecf, line 52: I think this check is unreachable because code now has a `.min(1..)`

apps/web/components/dashboard-components/project-details/mcp-server-editor.tsx
```diff
@@ -235,9 +239,17 @@ export function McpServerEditor({
   const showAuthButton =
     !!mcpRequiresAuth &&
     !mcpIsAuthed &&
-    !authResult?.redirectUrl &&
     projectId &&
-    redirectToAuth;
+    redirectToAuth &&
+    authResult?.status !== "manual_client_registration_required";
```
alecf: move `"manual_client_registration_required"` into a constant in `@tambo-ai-cloud/core` and use that everywhere

apps/web/components/dashboard-components/project-details/mcp-server-editor.tsx
```diff
@@ -235,9 +239,17 @@ export function McpServerEditor({
...
+  const manualRegistrationResult =
+    authResult?.status === "manual_client_registration_required"
+      ? authResult
+      : null;
+  const manualRegistrationHelpText =
+    manualRegistrationResult?.reason === "registration_not_available"
```
alecf: same for "registration_not_available" etc

packages/db/src/oauth/OAuthLocalProvider.ts
```diff
@@ -34,20 +43,32 @@ export class OAuthLocalProvider implements OAuthClientProvider {
...
     this._serverUrl = serverUrl;
   }
 
-  get redirectUrl(): string {
-    return this._saveAuthUrl?.toString() ?? "";
+  get redirectUrl(): string | undefined {
+    return this._saveAuthUrl?.toString();
+  }
+
+  get clientMetadataUrl(): string | undefined {
```
alecf: not sure why this (and redirectUrl above) needs to be a getter - seems like this is fully derivable from props passed to the constructor (I realize that redirectUrl was already a getter, but lets get rid of both, getters can be confusing to debug)

packages/core/src/mcp-oauth-client.ts
```diff
@@ -0,0 +1,131 @@
...
+      buildMcpOAuthClientMetadataUrl(
+        getBaseUrlFromRedirectUrl(expectedRedirectUrl),
+      )
+    );
+  }
+
+  return clientInformation.redirect_uris.includes(expectedRedirectUrl);
+}
+
+function isTamboMcpOAuthClientId(clientId: string): boolean {
```
alecf, line 117: I don't really understand whats going on here - why would a client id be a url?

alecf (commented):
here's my initial (manual) pass, but I'm running this through some deeper analysis due to the security risks.

apps/web/app/oauth/callback/route.ts
```diff
@@ -89,30 +90,17 @@ export async function GET(request: NextRequest) {
 
     return NextResponse.redirect(redirectUrl);
   } catch (error) {
```
alecf, line 102: **🔴 P1 (Blocks Merge): Server-side error logging was removed**

The previous version had `console.error("Error handling OAuth callback:", error)` here. The new code only redirects to the error page — DB failures, token exchange errors, network timeouts, and malformed OAuth responses are no longer logged server-side.

When users report "OAuth failed," operators will have nothing in logs or Sentry to investigate.

**Fix:** Restore server-side error logging with session context before the redirect:
```ts
} catch (error) {
    console.error("OAuth callback failed", {
      sessionId: queryParams.state ?? queryParams.sessionId,
      error,
    });
```

apps/web/server/api/routers/tools.ts
```diff
@@ -204,36 +222,137 @@ export const toolsRouter = createTRPCRouter({
...
+            throw new Error("OAuth redirect URL was not set");
+          }
+
+          return {
+            status: "redirect",
+            redirectUrl,
+          };
+        }
+      } catch (error) {
+        if (!shouldReusePersistedState) {
```
alecf: **🔴 P1 (Blocks Merge): Stale persisted OAuth state creates an infinite failure loop**

When `shouldReusePersistedState` is `true` and `auth()` throws (expired/revoked tokens, failed refresh), this guard skips cleanup because `!shouldReusePersistedState` is `false`. The stale tokens remain in `toolProviderUserContexts`. Every subsequent attempt reuses the same broken tokens and fails identically — users get stuck with no self-healing escape hatch.

**Fix:** When persisted state is reused and auth fails, clear the stale tokens:
```ts
if (!shouldReusePersistedState) {
  await clearStoredMcpOAuthSession(db, localProvider.state());
} else {
  // Clear stale tokens so the next attempt re-authenticates from scratch
  await db.update(schema.toolProviderUserContexts)
    .set({ mcpOauthTokens: null, mcpOauthClientInfo: null })
    .where(eq(schema.toolProviderUserContexts.id, toolProviderUserContextId));
}
```

apps/web/app/oauth/callback/route.ts
```diff
@@ -25,13 +25,25 @@ const callbackParamsSchema = z
...
+      throw new Error("OAuth session ID missing from callback state");
+    }
+
+    if (!validatedParams.code) {
+      throw new Error("Authorization code is required");
+    }
+
+    const { code } = validatedParams;
     const db = getDb(env.DATABASE_URL);
     const oauthClient = await db.query.mcpOauthClients.findFirst({
```
alecf, line 58: **🔴 P1 (Blocks Merge): Missing unique index on `mcp_oauth_clients.session_id`**

This `findFirst` (and 5+ similar calls in `OAuthLocalProvider`) queries by `sessionId` on a column with no index. Every call does a sequential scan. More critically, there's no uniqueness constraint, so a bug or race condition could create duplicate sessions and `findFirst` would return an arbitrary row (session confusion in an OAuth flow).

**Fix:** Add a unique index on `mcp_oauth_clients.session_id` in the schema. This also enables replacing the read-then-write patterns in `saveClientInformation`/`saveCodeVerifier` with upserts.

apps/web/server/api/routers/tools.ts
```diff
@@ -549,3 +663,108 @@ async function upsertToolProviderUserContext(
...
+      undefined,
+      fetchFn,
+    );
+    const discoveredAuthorizationServer =
+      resourceMetadata.authorization_servers?.[0];
+
+    if (discoveredAuthorizationServer) {
+      authorizationServer = new URL(discoveredAuthorizationServer);
+    }
+  } catch {
```
alecf: **🟡 P2: Empty catch blocks in discovery swallow all error types**

Both catch blocks in this function (here and line 710) silently swallow every error — DNS failures, TLS errors, HTTP 500s, and timeouts all get the same treatment as a 404.

The second catch is particularly dangerous: a transient network error sets `metadataDiscovered: false`, which causes the caller to skip the manual registration prompt and attempt DCR instead — a flaky network silently changes the flow path.

**Fix:** Add `console.warn` logging in both catch blocks:
```ts
} catch (error) {
    console.warn("Failed to discover OAuth resource metadata", {
      serverUrl,
      error: error instanceof Error ? error.message : String(error),
    });
}
```

apps/web/server/api/routers/mcp-oauth-registration.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  if (
+    error instanceof ServerError ||
+    error instanceof TemporarilyUnavailableError
+  ) {
+    return false;
+  }
+
+  const message = getErrorMessage(error).toLowerCase();
```
alecf: **🟡 P2: Fragile string matching can misclassify errors**

After the typed error checks (good), this falls through to substring matching on error messages. A network error like `"Failed to connect to registration endpoint"` matches `"registration"`. An unrelated `"Invalid response_types in token response"` matches `"response_types"`. False positives show users a manual registration form when the real problem is something else.

**Suggestion:** Log when this heuristic fires so it can be audited, and consider being more conservative — treat unrecognized errors as generic failures rather than guessing based on message content.

apps/web/server/api/routers/tools.ts
```diff
@@ -204,36 +222,137 @@ export const toolsRouter = createTRPCRouter({
...
+          }
+
+          return {
+            status: "redirect",
+            redirectUrl,
+          };
+        }
+      } catch (error) {
+        if (!shouldReusePersistedState) {
+          await clearStoredMcpOAuthSession(db, localProvider.state());
```
alecf: **🟡 P2: `clearStoredMcpOAuthSession` failure replaces the original error**

If this cleanup query throws (DB connection lost, constraint violation), it replaces the `error` from the `auth()` call. The caller sees a confusing database error instead of the actual OAuth failure.

**Fix:** Wrap in its own try-catch:
```ts
try {
  await clearStoredMcpOAuthSession(db, localProvider.state());
} catch (cleanupError) {
  console.error("Failed to clean up OAuth session", {
    sessionId: localProvider.state(),
    cleanupError,
    originalError: error,
  });
}
```

apps/web/app/oauth/callback/route.ts
```diff
@@ -63,17 +75,6 @@ export async function GET(request: NextRequest) {
...
-        new URL(
-          `/auth/error?error=${encodeURIComponent(validatedParams.error)}`,
-          request.url,
-        ),
-      );
-    }
-
     await auth(oauthProvider, {
       serverUrl: oauthClient.sessionInfo.serverUrl,
       authorizationCode: code,
```
alecf, line 90: **🟡 P2: `mcpOauthClients` session records are never cleaned up on success**

`clearStoredMcpOAuthSession` is only called on error paths (in `tools.ts`). After `auth()` succeeds (just below this line) and tokens are persisted to `toolProviderUserContexts`, the `mcpOauthClients` row — containing the code verifier and client info — is left permanently. These records accumulate without bound.

**Fix:** Delete the session record after successful token exchange:
```ts
await auth(oauthProvider, { ... });

// Clean up ephemeral OAuth session now that tokens are persisted
await db.delete(schema.mcpOauthClients)
  .where(eq(schema.mcpOauthClients.sessionId, sessionId));
```

apps/web/server/api/routers/tools.ts
```diff
@@ -549,3 +663,108 @@ async function upsertToolProviderUserContext(
     return newToolProviderUserContext.id;
```
alecf: **🟡 P2: `tools.ts` has grown to 770 lines — extract auth orchestration**

The file now contains ~400 lines of auth helper functions that are separate concerns from the tRPC route definitions. Additionally, `upsertToolProviderUserContext` (line 638) is a pure DB operation that belongs in `packages/db/src/operations/` per project conventions.

**Suggestion:** Extract auth orchestration to `apps/web/server/api/services/mcp-authorization.ts` and move the DB operation to `packages/db/src/operations/`.

packages/db/src/oauth/OAuthLocalProvider.ts
```diff
@@ -58,6 +79,21 @@ export class OAuthLocalProvider implements OAuthClientProvider {
...
+            ),
+          });
+
+        if (toolProviderUserContext?.mcpOauthClientInfo) {
+          this._clientInformation = toolProviderUserContext.mcpOauthClientInfo;
+          return this._clientInformation;
+        }
+      }
+
       const session = await this.db.query.mcpOauthClients.findFirst({
```
alecf: **🟡 P2: Redundant DB queries — `clientInformation()` and `codeVerifier()` fetch the same row independently**

Both methods query `mcpOauthClients` by `sessionId` in separate round trips for the same row. `clientInformation()` may also first hit `toolProviderUserContexts` (line 82), so a single auth flow can trigger 3+ queries when 1 suffices.

Combined with the read-before-write pattern in `saveClientInformation`/`saveCodeVerifier`, a single authorization can generate ~10 DB queries — against an unindexed column (see P1 re: missing index).

**Suggestion:** Cache the session row on first fetch so both methods share a single query. Once the unique index is in place, replace read-then-write with `onConflictDoUpdate`.

apps/web/app/oauth/callback/route.ts
```diff
@@ -10,11 +10,11 @@ import { z } from "zod/v3";
...
-    state: z.string().optional(),
-    error: z.string().optional(),
+    code: z.string().min(1, "Authorization code is required").optional(),
+    state: z.string().min(1).optional(),
+    error: z.string().min(1).optional(),
     redirect_uri: z.string().url().optional(),
-    sessionId: z.string(),
+    sessionId: z.string().min(1).optional(),
   })
   .passthrough();
```
alecf: **🟡 P2: `.passthrough()` weakens input validation on a security-critical endpoint**

Allows arbitrary additional query parameters through without validation. While extra params are currently unused, this defeats strict input validation on an OAuth callback. If future code accesses additional properties from `validatedParams`, they bypass schema checks.

**Fix:** Remove `.passthrough()` — Zod defaults to `.strip()` which silently drops unknown keys.

apps/web/app/oauth/error/page.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <Card className="w-full">
+        <CardHeader>
+          <CardTitle>OAuth Authorization Failed</CardTitle>
+        </CardHeader>
+        <CardContent className="flex flex-col gap-4">
+          <p className="text-sm text-muted-foreground">
+            Tambo Cloud could not finish the MCP authorization flow.
+          </p>
+          {error && (
+            <pre className="overflow-x-auto rounded-md bg-muted p-3 font-mono text-xs whitespace-pre-wrap break-words">
```
alecf: **🟡 P2: Error message reflection enables phishing via crafted URLs**

No XSS risk (React escapes content), but the `error` query param is attacker-controlled. Anyone can craft `/oauth/error?error=Your+account+has+been+compromised.+Visit+evil.com+to+reset.` and it renders as a legitimate-looking Tambo error page.

**Suggestion:** Use a fixed set of error codes mapped to predefined messages, rather than reflecting arbitrary strings. Or truncate the message and strip URLs.

apps/web/server/api/routers/tools.ts
```diff
@@ -204,36 +222,137 @@ export const toolsRouter = createTRPCRouter({
...
+            "Tambo Cloud OAuth requires NEXT_PUBLIC_APP_URL or VERCEL_URL to resolve to an HTTPS URL or localhost",
+        });
+      }
+
       const toolProviderUserContextId = await upsertToolProviderUserContext(
         db,
         toolProviderId,
         contextKey,
       );
+      const toolProviderUserContext =
```
alecf: **🟡 P2: Redundant re-query after `upsertToolProviderUserContext`**

The upsert function (line 638) fetches the full row via `findFirst` or gets it from `.returning()` but only returns the `id`, forcing this second query to get the full context.

**Fix:** Return the full `toolProviderUserContext` row from `upsertToolProviderUserContext` instead of just the ID.

apps/web/server/api/routers/tools.ts
```diff
@@ -549,3 +663,108 @@ async function upsertToolProviderUserContext(
...
+    }
+
+    return {
+      authorizationServer: authorizationServer.toString(),
+      hasRegistrationEndpoint: !!metadata.registration_endpoint,
+      metadataDiscovered: true,
+      supportsClientMetadataDocument:
+        metadata.client_id_metadata_document_supported === true,
+    };
+  } catch {
```
alecf: **🟡 P2: Second empty catch — same issue as line 689**

This catch converts any failure (network errors, timeouts, 500s) into `metadataDiscovered: false`. The caller uses this to decide NOT to prompt for manual registration, so a transient failure silently changes flow behavior.

**Fix:** Add `console.warn` logging, and consider distinguishing 404 ("endpoint doesn't exist") from 5xx/network errors ("endpoint failed").

alecf (changes requested):
## Security-focused review of the MCP OAuth registration flow

Thoroughly reviewed with 5 specialized analysis passes (security, architecture, performance, silent-failure, simplicity). The OAuth flow changes are architecturally sound and solve a real problem (Mapbox MCP rejecting minimal client metadata). However there are several issues — particularly around observability, error handling, and a potential infinite-retry loop — that should be addressed before merge.

### Additional findings not attached inline

**SSRF DNS rebinding (P2, pre-existing):** `apps/web/lib/urlSecurity.ts` resolves DNS at validation time but the HTTP request happens later, leaving a DNS rebinding window. `ALLOW_LOCAL_MCP_SERVERS` also completely disables all URL safety checks. Not blocking for this PR but worth tracking.

**`saveTokens` doesn't verify the DB update affected rows (P2):** `OAuthLocalProvider.saveTokens` (line 240) updates `toolProviderUserContexts` but doesn't check if any row was matched. If the row was deleted (race condition), tokens are set in memory but never persisted — user sees success but next MCP call fails.

14 inline comments below (3 P1 blocking, 11 P2 should-fix).

packages/core/src/mcp-oauth-client.ts
rehmatsg: This is from MCP docs:

> MCP clients and authorization servers SHOULD support OAuth Client ID Metadata Documents as specified in [OAuth Client ID Metadata Document](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00). This approach enables clients to use HTTPS URLs as client identifiers, where the URL points to a JSON document containing client metadata. This addresses the common MCP scenario where servers and clients have no pre-existing relationship.

https://modelcontextprotocol.io/specification/draft/basic/authorization#client-id-metadata-documents

apps/web/components/dashboard-components/project-details/mcp-server-editor.tsx
rehmatsg: refactored this in latest commit

apps/web/components/dashboard-components/project-details/mcp-server-editor.tsx
rehmatsg: refactored this in latest commit

apps/web/app/oauth/callback/route.ts
rehmatsg: Fixed in `9b4d98ef`.

apps/web/server/api/routers/tools.ts
rehmatsg: Fixed in `9c86de2c`.

When `authorizeMcpServer` reuses persisted MCP OAuth state and `auth()` fails, we now clear the persisted `mcpOauthTokens` and `mcpOauthClientInfo` on `toolProviderUserContexts` before surfacing the error. Fresh auth failures still only clean up the ephemeral `mcpOauthClients` session row.

apps/web/app/oauth/callback/route.ts
rehmatsg: Fixed in `5203d66b`.

This adds a unique constraint on `mcp_oauth_clients.session_id`, so the database now enforces the one-row-per-session invariant instead of relying on `findFirst()` behavior.

I also updated `OAuthLocalProvider` to use single-statement upserts keyed by `sessionId` for `saveClientInformation()` and `saveCodeVerifier()`, which removes the current read-then-write race window.

packages/core/src/mcp-oauth-client.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  if (!("redirect_uris" in clientInformation)) {
+    if (!isTamboMcpOAuthClientId(clientInformation.client_id)) {
+      return true;
+    }
+
+    return (
+      clientInformation.client_id ===
+      buildMcpOAuthClientMetadataUrl(redirectUrl)
```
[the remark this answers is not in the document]
rehmatsg: Fixed in `a7402054`. `hasCompatibleMcpOAuthClientRedirect` now derives the base URL from `expectedRedirectUrl` via `getBaseUrlFromRedirectUrl()` before comparing metadata-document client IDs, and the parameter name was updated so the callback URL vs base URL distinction is explicit.

apps/web/server/api/routers/mcp-oauth-registration.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    message.includes("registration") ||
+    message.includes("client details") ||
+    message.includes("client metadata") ||
+    message.includes("client_id") ||
+    message.includes("client id") ||
+    message.includes("response_types")
+  );
```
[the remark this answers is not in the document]
rehmatsg: Fixed in `f31c9c22`. I tightened the fallback classification so it no longer treats broad substrings like `registration` or bare `response_types` as manual-registration errors. The heuristic now only matches explicit client-metadata validation-style failures and logs when that path fires.

apps/web/server/api/routers/tools.ts
```diff
@@ -549,3 +663,87 @@ async function upsertToolProviderUserContext(
...
+    authorizationServer,
+    reason,
+    suggestedClientMetadata: {
+      clientName: suggestedClientMetadata.client_name ?? "Tambo Cloud",
+      clientUri: suggestedClientMetadata.client_uri ?? baseUrl,
+      redirectUris: suggestedClientMetadata.redirect_uris,
+      grantTypes: suggestedClientMetadata.grant_types ?? [],
+      responseTypes: suggestedClientMetadata.response_types ?? [],
+      tokenEndpointAuthMethod:
+        suggestedClientMetadata.token_endpoint_auth_method ?? "none",
```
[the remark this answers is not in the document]
rehmatsg: Fixed in `68e1249a`. I removed the dead `??` fallbacks from the suggested client metadata response and replaced them with a single assertion of the required fields, so we no longer hide impossible optionality behind runtime defaults.

apps/web/app/oauth/callback/route.ts
rehmatsg: Left this guard intentionally. The schema keeps `code` optional so provider callbacks that only carry `error` still validate, and the explicit `!validatedParams.code` branch still handles malformed success callbacks that omit `code`. That callback hardening landed in `078af8af`.

packages/db/src/oauth/OAuthLocalProvider.ts
rehmatsg: Fixed in `d3f62a07`. `redirectUrl` and the client metadata URL are now constructor-initialized readonly properties instead of derived getters. I left `redirectStartAuthUrl` as the only getter because it is not derivable from constructor input; it is populated later by `redirectToAuthorization()`.

apps/web/server/api/routers/tools.ts
rehmatsg: Fixed across `99a31d0c` and `34dbfe55`. `99a31d0c` added `console.warn` logging to both discovery catch blocks. `34dbfe55` then tightened the authorization-server metadata path so only an actual "no metadata document" result yields `metadataDiscovered: false`; thrown network/5xx failures now stay on the normal auth error path instead of silently forcing DCR.

apps/web/server/api/routers/mcp-oauth-registration.ts
rehmatsg: Fixed in `f31c9c22`. I narrowed the string-matching fallback so generic errors like `Failed to connect to registration endpoint` no longer trigger the manual-registration path. Only explicit client-metadata-style validation failures use that heuristic now, and it logs when it fires.

apps/web/server/api/routers/tools.ts
rehmatsg: Fixed in `7d5c3117`. The session cleanup now runs inside its own guarded helper, so a cleanup failure is logged with the session id and original auth error instead of replacing the real OAuth failure seen by the caller.

apps/web/app/oauth/callback/route.ts
rehmatsg: Fixed in `c8074c69`. After `auth()` returns `AUTHORIZED`, the flow now deletes the ephemeral `mcpOauthClients` session row so the code-verifier/client-info record does not accumulate indefinitely. Cleanup is also guarded so a delete failure does not turn a successful auth into a false failure.

apps/web/server/api/routers/tools.ts
rehmatsg: Fixed in `a8bf1720`. I extracted the MCP auth orchestration into `apps/web/server/api/services/mcp-authorization.ts` and moved `upsertToolProviderUserContext` into the DB operations layer, which brings the router back down to route-focused code.

packages/db/src/oauth/OAuthLocalProvider.ts
rehmatsg: Fixed in `1b7cbd50`. `OAuthLocalProvider` now caches the `mcpOauthClients` session row on first read so `clientInformation()` and `codeVerifier()` share a single lookup instead of issuing separate round trips. The write side was already switched to `onConflictDoUpdate` in `5203d66b` once `sessionId` became unique.

apps/web/app/oauth/callback/route.ts
rehmatsg: Fixed in `1316a2a0`. I removed `.passthrough()` from the OAuth callback query schema, so unknown query parameters are stripped instead of being retained on this security-sensitive endpoint.

apps/web/app/oauth/error/page.tsx
rehmatsg: Fixed in `078af8af`. The error page now uses a fixed set of internal `code` values mapped to predefined messages instead of reflecting arbitrary `error` query text, so crafted URLs can no longer render attacker-controlled copy on a Tambo-branded error page.

apps/web/server/api/routers/tools.ts
rehmatsg: Fixed in `f63e75a5`. `upsertToolProviderUserContext` now returns the full `toolProviderUserContexts` row and the auth flow consumes that row directly, so the immediate follow-up lookup by id is gone.

apps/web/server/api/routers/tools.ts
rehmatsg: Fixed in `34dbfe55`. The second discovery catch still logs, but it no longer turns every failure into `metadataDiscovered: false`. Only a genuine metadata-missing result falls back that way now; thrown timeout/network/5xx failures are surfaced as normal auth errors.

rehmatsg:
@alecf I’ve addressed the remaining review findings, including the `saveTokens()` row-match check in `123752b0`. The branch is synced with the PR head now. Please take another look when you have a minute.

alecf (commented):
## Performance Review

### Performance Summary

The OAuth flow involves sequential external network calls (discovery + auth) plus multiple DB queries. Most of the individual operations are fine for a user-initiated OAuth registration flow (low QPS), but there are several concrete improvements worth making -- particularly around redundant DB queries in `OAuthLocalProvider`, a duplicate validation call in `inspectMcpServer`, and timeout cascading risk.

### Scalability Assessment

This is a user-initiated OAuth registration flow, so it will never see high QPS. The primary concern is not throughput but **individual request latency** -- a single `authorizeMcpServer` call can stack up to ~15 seconds of external fetch timeouts plus multiple sequential DB round-trips, which directly impacts user-perceived responsiveness. The `inspectMcpServer` endpoint is more concerning since it could be called repeatedly from the UI and does redundant work.

### Recommended Actions (prioritized)

1. **[High impact, low effort]** Eliminate duplicate MCP server validation in `inspectMcpServer` -- the MCPClient connection already proves validity
2. **[Medium impact, low effort]** Combine the `clientInformation()` and `tokens()` DB lookups in `OAuthLocalProvider` into a single query
3. **[Medium impact, low effort]** Parallelize independent DB + network operations in `authorizeMcpServer` where data dependencies allow
4. **[Low impact, low effort]** Add an index on `tool_provider_user_contexts.tool_provider_id` if one does not already exist -- this column is queried in `upsertToolProviderUserContext` and in `OAuthLocalProvider` methods

See line comments below for specifics.

packages/db/src/oauth/OAuthLocalProvider.ts
alecf, line 1: **Performance: Redundant DB queries in `clientInformation()` and `tokens()`**

Both `clientInformation()` (line 86) and `tokens()` (line 218) independently query `toolProviderUserContexts` by the same ID when `_usePersistedContextState` is true. During a typical auth flow, the MCP SDK calls both methods, resulting in two identical `SELECT` queries against the same row.

These should be consolidated into a single lazy-loaded fetch that populates both `_clientInformation` and `_tokens` in one round-trip:

```typescript
private async getPersistedContext() {
  if (this._persistedContext !== undefined) return this._persistedContext;
  
  this._persistedContext = await this.db.query.toolProviderUserContexts.findFirst({
    where: eq(schema.toolProviderUserContexts.id, this.toolProviderUserContextId),
  });
  return this._persistedContext;
}
```

Then `clientInformation()` and `tokens()` both call `getPersistedContext()` and read their respective field. This cuts one DB round-trip per auth attempt.

apps/web/server/api/routers/tools.ts
alecf, line 1: **Performance: `inspectMcpServer` does duplicate work -- creates MCPClient AND validates the server**

Lines 345-359: `inspectMcpServer` first creates an `MCPClient` (which connects to the remote MCP server, performs capability negotiation, etc.), then separately calls `validateMcpServer` which creates *another* connection to the same server to check validity.

This is two full MCP connection round-trips to the same remote server per inspect call. The `MCPClient.create()` already proves the server is reachable and negotiates capabilities. The `validateMcpServer` call is redundant here -- you already have a live client.

Consider extracting the version/instructions/capabilities from the already-connected `mcpClient` instance instead of making a second connection via `validateMcpServer`. If `MCPClient` exposes server info after connection (it typically does via the initialization response), use that directly.

apps/web/server/api/services/mcp-authorization.ts
alecf, line 1: **Performance: Sequential DB query + network call that could be parallelized in `authorizeMcpServer`**

Lines 71-123: The function executes these steps sequentially:
1. `db.query.toolProviders.findFirst(...)` -- DB lookup
2. `operations.ensureProjectAccess(db, projectId, userId)` -- another DB query
3. `operations.upsertToolProviderUserContext(db, toolProviderId, contextKey)` -- DB transaction (read + possible write)
4. `getMcpAuthorizationServerDetails(url)` -- two external HTTP requests with 5s timeouts each

Steps 2 and 3 are independent of each other (both depend only on `toolProviderId`/`projectId` from step 1). They could run in parallel:

```typescript
const [, toolProviderUserContext] = await Promise.all([
  operations.ensureProjectAccess(db, projectId, userId),
  operations.upsertToolProviderUserContext(db, toolProviderId, contextKey),
]);
```

This saves one DB round-trip (~1-5ms locally, more with remote DB). Not a huge win in isolation, but in a flow that already stacks up to 15+ seconds of external timeouts, every sequential millisecond matters for perceived responsiveness.

Also worth noting: `getMcpAuthorizationServerDetails` makes two sequential HTTP calls (resource metadata discovery, then authorization server metadata discovery) each with a 5-second timeout. In the worst case where the first call times out, this function alone takes 10 seconds before the `auth()` call even starts (which has its own 5-second timeout). The total worst-case for a single `authorizeMcpServer` call is therefore ~20 seconds. Consider whether the resource metadata discovery timeout should be shorter (e.g., 2 seconds) since it is optional and falls back gracefully.

apps/web/server/api/routers/tools.ts
alecf, line 1: **Performance: `addMcpServer` runs URL safety check and server key uniqueness check sequentially**

Lines 111-141: `getExistingServerKeys` (DB query) and `validateSafeURL` (DNS resolution) are independent of each other. They could run in parallel:

```typescript
const [existingKeys, safetyCheck] = await Promise.all([
  getExistingServerKeys(ctx.db, projectId),
  validateSafeURL(parsedUrl.hostname),
]);
```

Then `validateMcpServer` (which actually connects to the remote server) must remain sequential since it depends on the safety check passing. This parallelization saves the DNS resolution time (~10-100ms) on the happy path.

apps/web/app/oauth/callback/route.ts
alecf, line 1: **Performance: Nested `with` relations in callback query are well-structured, but note the 10-second timeout risk**

The DB query at line 58-71 uses nested `with` relations (`mcpOauthClients -> toolProviderUserContext -> toolProvider`). This is fine -- Drizzle resolves these as JOINs, not N+1 queries, and you only select the `projectId` column from `toolProvider`. The unique constraint on `session_id` (added in this PR's migration) ensures the lookup is indexed. No issue here.

However, the `auth()` call at line 88 uses a 10-second timeout. If the upstream OAuth provider is slow, this route handler will hold a DB connection open for the full 10 seconds (the `db` reference from `getDb` stays alive). If many users hit OAuth callbacks simultaneously with a slow provider, this could exhaust the connection pool. Consider whether the DB connection can be released before the `auth()` call, or ensure the connection pool is sized to handle concurrent slow callbacks.

alecf (commented):
Code Pattern and Quality Review - see full analysis below in thread

alecf (commented):
## Code Review: PR #2692 — fix(web, db, core): improve MCP OAuth registration flow

**Reviewed by**: 8 parallel analysis agents (security, performance, architecture, patterns, data integrity, silent failures, test coverage, simplicity)

---

### Findings Summary

- **Total Findings:** 22
- 🔴 **CRITICAL (P1):** 3 — BLOCKS MERGE
- 🟡 **IMPORTANT (P2):** 11 — Should Fix
- 🔵 **NICE-TO-HAVE (P3):** 8 — Enhancements

See inline comments for details on each finding. The P1 items are:

1. **Missing `contextKey` filter in `upsertToolProviderUserContext`** — Could return wrong user's context, causing OAuth token cross-contamination between users
2. **SSRF via unvalidated authorization server URL** — Malicious MCP servers can make Tambo request internal/cloud-metadata URLs
3. **`clearPersistedMcpOAuthState` can throw and replace original error** — Cleanup failure in catch block overwrites the real auth error

### Positive Observations

- Discriminated union for auth output schema is excellent type design
- `satisfies` for exhaustive error message record is idiomatic
- Test coverage for new modules is solid (7 new test files)
- Service extraction from tools router was the right architectural call
- State-based session recovery is cleaner than session-bound redirects

packages/db/src/operations/project.ts
```diff
@@ -596,6 +596,35 @@ export async function getMcpServer(
...
+  contextKey: string | null,
+): Promise<typeof schema.toolProviderUserContexts.$inferSelect> {
+  return await db.transaction(async (tx) => {
+    const toolProviderUserContext =
+      await tx.query.toolProviderUserContexts.findFirst({
+        where: eq(
+          schema.toolProviderUserContexts.toolProviderId,
+          toolProviderId,
+        ),
+      });
```
alecf, line 611: 🔴 **P1 — CRITICAL: Missing `contextKey` filter causes token cross-contamination**

This `findFirst` query only filters on `toolProviderId` but ignores `contextKey`. If a tool provider has multiple user contexts (different end-users), this returns an arbitrary one — User A's OAuth session could be attached to User B's context row.

```typescript
// Current (broken):
where: eq(schema.toolProviderUserContexts.toolProviderId, toolProviderId),

// Fix:
where: and(
  eq(schema.toolProviderUserContexts.toolProviderId, toolProviderId),
  contextKey === null
    ? isNull(schema.toolProviderUserContexts.contextKey)
    : eq(schema.toolProviderUserContexts.contextKey, contextKey),
),
```

Additionally, this check-then-insert pattern has a race condition under concurrent requests (both transactions read "no row", both try to insert, one fails with a constraint violation). Consider using `INSERT ... ON CONFLICT DO NOTHING` + `SELECT` instead.

*Found by: data-integrity-guardian*

apps/web/server/api/services/mcp-authorization.ts
```diff
@@ -0,0 +1,451 @@
...
+      serverUrl,
+      undefined,
+      fetchFn,
+    );
+    const discoveredAuthorizationServer =
+      resourceMetadata.authorization_servers?.[0];
+
+    if (discoveredAuthorizationServer) {
+      authorizationServer = new URL(discoveredAuthorizationServer);
+    }
```
alecf, line 351: 🔴 **P1 — CRITICAL: SSRF via unvalidated authorization server URL**

The `authorization_servers` URL from the remote MCP server's resource metadata is used directly without SSRF validation. A malicious MCP server could return `authorization_servers: ["http://169.254.169.254/"]` (AWS IMDS) or any internal URL, and Tambo's server would request it via `discoverAuthorizationServerMetadata`.

The initial MCP server URL is validated via `validateServerUrl` in the tRPC input, but this discovered URL bypasses that check entirely.

Fix: validate the discovered URL before using it:

```typescript
if (discoveredAuthorizationServer) {
  const parsed = new URL(discoveredAuthorizationServer);
  const safetyCheck = await validateSafeURL(parsed.hostname);
  if (!safetyCheck.safe) {
    throw new Error(`Unsafe authorization server URL: ${safetyCheck.reason}`);
  }
  authorizationServer = parsed;
}
```

*Found by: security-sentinel*

apps/web/app/oauth/callback/route.ts
alecf: ooops sorry this got posted before I had looked this over, ignore this one

apps/web/server/api/services/mcp-authorization.ts
```diff
@@ -0,0 +1,451 @@
...
+      }
+
+      return {
+        status: MCP_OAUTH_AUTHORIZATION_STATUS_REDIRECT,
+        redirectUrl,
+      };
+    }
+  } catch (error) {
+    if (shouldReusePersistedState) {
+      await clearPersistedMcpOAuthState(db, toolProviderUserContext.id);
```
alecf, line 185: 🔴 **P1 — CRITICAL: Cleanup throw replaces original auth error**

`clearPersistedMcpOAuthState` on this line is NOT wrapped in try-catch. If this DB call fails (connection error, etc.), the original auth error from line 183's catch is lost and replaced by a confusing cleanup error.

Compare with the `else` branch on line 187 which correctly uses `clearStoredMcpOAuthSessionSafely` — that function handles its own errors. This branch should do the same:

```typescript
if (shouldReusePersistedState) {
  try {
    await clearPersistedMcpOAuthState(db, toolProviderUserContext.id);
  } catch (cleanupError) {
    console.error("Failed to clear persisted MCP OAuth state", {
      toolProviderUserContextId: toolProviderUserContext.id,
      cleanupError,
      originalError: error,
    });
  }
}
```

*Found by: silent-failure-hunter*

alecf:
@rehmatsg I'm a little concerned that this PR is getting kind of out of control in terms of the scope of what it is doing - you've effectively refactored our whole MCP Oauth stack, which seems to have surfaced a ton of issues that may have been in existing code, but if we're going to refactor we need to fix them - either first or as a part of a separate refactor.

Can we break this up into smaller, more understandable PRs?

Here's the full list of what we found, most issues pretty legit:

P1 — CRITICAL (blocks merge)

  1. Missing contextKey filter causes token cross-contamination — packages/db/src/operations/project.ts:606
  The findFirst query only filters on toolProviderId, ignoring contextKey. If a tool provider has multiple user contexts, it returns
  an arbitrary one — User A's OAuth tokens could be attached to User B's context.
  2. SSRF via unvalidated authorization server URL — apps/web/server/api/services/mcp-authorization.ts:351
  The authorization_servers URL from remote MCP server responses is used directly without SSRF validation. A malicious MCP server
  could point it at internal networks or cloud metadata endpoints.
  3. clearPersistedMcpOAuthState can throw and replace original error — apps/web/server/api/services/mcp-authorization.ts:185
  Cleanup call in catch block isn't wrapped in try-catch. If the DB call fails, the original auth error is lost. The adjacent else
  branch correctly uses clearStoredMcpOAuthSessionSafely — this branch should do the same.

  P2 — IMPORTANT (should fix)

  4. No session TTL on OAuth sessions — apps/web/app/oauth/callback/route.ts:58
  Sessions in mcp_oauth_clients persist forever. Add a TTL check (e.g. 15 min) when looking up the session in the callback route.
  5. Inverted dependency: service imports from router — apps/web/server/api/services/mcp-authorization.ts:7
  mcp-authorization.ts imports shouldReusePersistedMcpOAuthState, requiresManualMcpClientRegistration, and
  isManualMcpClientRegistrationRequiredError from ../routers/mcp-oauth-registration. Move those 3 pure functions to the service or a
  shared utility.
  6. Resource metadata discovery silent fallback — apps/web/server/api/services/mcp-authorization.ts:352
  When discoverOAuthProtectedResourceMetadata fails, the code catches the error and silently falls back to new URL("/", serverUrl).
  If the real auth server is at a different domain, the code proceeds with the wrong URL. Violates project's "No Silent Fallbacks"
  principle.
  7. No tests for mcp-authorization.ts — apps/web/server/api/services/mcp-authorization.ts
  The 451-line orchestration service has no dedicated test file. getOAuthProvider (4 return paths) and getServerValidity are
  completely untested. The tools.test.ts exercises authorizeMcpServer indirectly but leaves these functions uncovered.
  8. Heuristic string matching fragility — apps/web/server/api/routers/mcp-oauth-registration.ts:73
  isManualMcpClientRegistrationHeuristicMatch matches on free-text error messages. Errors like "Failed to fetch client metadata:
  invalid certificate" would be misclassified as needing manual registration, hiding the real failure.
  9. console.error/console.warn invisible to Sentry — apps/web/server/api/services/mcp-authorization.ts:315, :353, :241
  Multiple log calls in the service use console.error/console.warn which won't reach Sentry in cloud deployments.
  10. Duplicated getErrorMessage utility — apps/web/server/api/services/mcp-authorization.ts:449 and
  apps/web/server/api/routers/mcp-oauth-registration.ts:102
  Identical function defined in two files. Extract to a shared location.
  11. Extract manual registration form from 628-line component —
  apps/web/components/dashboard-components/project-details/mcp-server-editor.tsx:386
  135 lines of inline JSX with its own state (manualClientId, manualClientSecret). Should be a separate McpManualRegistrationForm
  component. Component went from 470 to 628 lines, well over the 200-300 line guideline.
  12. Race condition in upsertToolProviderUserContext — packages/db/src/operations/project.ts:604
  Check-then-insert inside a transaction under READ COMMITTED isolation. Two concurrent requests can both read "no row" and both try
  to insert, causing a constraint violation. Use INSERT ... ON CONFLICT DO NOTHING + SELECT instead.

  P3 — NICE-TO-HAVE (enhancements)

  13. Redundant null check on url — apps/web/server/api/services/mcp-authorization.ts:88
  url is guaranteed non-null by the isNotNull(schema.toolProviders.url) filter on line 75. This guard can never trigger.
  14. Dead assertSuggestedClientMetadataFields — apps/web/server/api/services/mcp-authorization.ts:429
  buildMcpOAuthClientMetadata always returns all the fields this assertion checks. The function will never throw. Delete it.
  15. Verbose 60-character constant names — packages/core/src/mcp-oauth-status.ts:3
  MCP_OAUTH_AUTHORIZATION_STATUS_MANUAL_CLIENT_REGISTRATION_REQUIRED is 62 characters. Consider a namespace object like
  McpOAuthStatus.ManualRegistrationRequired.
  16. Empty catch in isTamboMcpOAuthClientId — packages/core/src/mcp-oauth-client.ts:118
  Use URL.canParse(clientId) (available in Node 22+) instead of try { new URL(...) } catch { return false }.
  17. HTTP 404 wrong status for HTTPS precondition — apps/web/app/oauth/client-metadata/route.ts:12
  Returning 404 when HTTPS is not available is semantically incorrect — the resource exists, it just can't be served. 403 or 503
  would be more appropriate unless the MCP spec mandates 404.
  18. Duplicate DB queries in OAuthLocalProvider — packages/db/src/oauth/OAuthLocalProvider.ts:86 and :219
  clientInformation() and tokens() each independently query toolProviderUserContexts by the same primary key. During a single auth
  flow the MCP SDK calls both, producing two identical SELECT queries. Consolidate into a single lazy-loaded method.
  19. Implicit metadata-URL-as-client-id fallback — packages/db/src/oauth/OAuthLocalProvider.ts:165
  saveCodeVerifier synthesizes { client_id: this.clientMetadataUrl } when no client info exists — surprising behavior with no
  explanatory comment. The error message when clientMetadataUrl is also undefined gives no hint about the HTTPS requirement.

rehmatsg:
@alecf You’re right that this PR ended up mixing a few different things. The original goal was just Mapbox MCP support, which only requires the richer DCR payload. The manual-registration fallback work is a second coherent slice, but it has hard prerequisites: stable callback URL, state-based recovery, durable client storage, API response changes, and the dashboard fallback UI. The rest of the OAuth cleanup and compatibility work can be split out.

Would it be better to split this into small replacement PRs?

- Supporting Mapbox (and other MCPs) with richer DCR metadata payload
- Manual registration fallback plus its required callback/session/storage pieces
- Follow-up OAuth hardening / compatibility cleanup

I’ll also make sure the blocking correctness/security issues called out in review are fixed in the relevant replacement PRs rather than carried forward.

michaelmagan:

PR #2755, opened 2026-04-06
fix(api): preserve content block ordering for inline tool_use/tool_result

seth-tambo:
## Summary

- **Fix**: `contentToV1Blocks` silently dropped inline `tool_use`/`tool_result` blocks (logged as "unknown content type"), then re-appended `tool_use` at the end via the legacy `toolCallRequest` fallback. This broke interleaved ordering, causing post-tool-call text to render before the tool call in the console app.
- **Change**: Handle `tool_use` and `tool_result` inline during content iteration so blocks stay in their natural position. A `foundToolUseInContent` flag prevents duplication with the legacy fallback, which still fires for older messages that only have `toolCallRequest`.
- **Tests**: 6 new tests covering interleaved ordering, no duplication, legacy fallback, inline `tool_result`, and `_tambo_*` display property enrichment.

Fixes TAM-1417

## Test plan

- [x] All 56 tests in `v1-conversions.test.ts` pass
- [x] TypeScript type-check passes
- [x] Lint passes
- [ ] Manual: open console app, trigger a tool call with follow-up text, verify blocks render in order

### DIFF
--- a/apps/api/src/v1/__tests__/v1-conversions.test.ts
+++ b/apps/api/src/v1/__tests__/v1-conversions.test.ts
@@ -293,107 +293,6 @@ describe("v1-conversions", () => {
       });
     });
 
-    it("should add component block when componentDecision exists", () => {
-      const message = {
-        ...baseMessage,
-        content: [{ type: "text", text: "Here is a card" }],
-        componentDecision: {
-          componentName: "WeatherCard",
-          props: { temperature: 72 },
-          message: "",
-          componentState: null,
-        },
-        componentState: { expanded: true },
-      } as unknown as DbMessage;
-
-      const result = contentToV1Blocks(message);
-
-      expect(result).toHaveLength(2);
-      expect(result[0]).toEqual({ type: "text", text: "Here is a card" });
-      expect(result[1]).toEqual({
-        type: "component",
-        id: "comp_msg_123",
-        name: "WeatherCard",
-        props: { temperature: 72 },
-        state: { expanded: true },
-      });
-    });
-
-    it("should warn when componentDecision has no componentName and no toolCallRequest", () => {
-      const warnSpy = jest.spyOn(console, "warn").mockImplementation(() => {});
-
-      const message = {
-        ...baseMessage,
-        componentDecision: {
-          componentName: null,
-          props: { foo: "bar" },
-          message: "",
-          componentState: null,
-        },
-        toolCallRequest: null, // No tool call - this is a data integrity issue
-      } as unknown as DbMessage;
-
-      // Should not throw, but should warn
-      const result = contentToV1Blocks(message);
-      expect(result).toEqual([]); // No blocks generated
-      expect(warnSpy).toHaveBeenCalledWith(
-        expect.stringMatching(/has no componentName.*data integrity issue/),
-      );
-
-      warnSpy.mockRestore();
-    });
-
-    it("should NOT throw when componentDecision has no componentName but has toolCallRequest", () => {
-      // componentDecision without componentName is valid for tool call messages
-      // because it stores _tambo_* status messages
-      const message = {
-        ...baseMessage,
-        role: "assistant",
-        componentDecision: {
-          componentName: null,
-          props: {},
-          message: "Fetching data...",
-          statusMessage: "Working...",
-          componentState: null,
-        },
-        toolCallRequest: {
-          toolName: "getData",
-          parameters: [{ parameterName: "id", parameterValue: "123" }],
-        },
-        toolCallId: "call_123",
-      } as unknown as DbMessage;
-
-      // Should not throw
-      const result = contentToV1Blocks(message);
-
-      // Should have a tool_use block but no component block
-      expect(result).toHaveLength(1);
-      expect(result[0].type).toBe("tool_use");
-    });
-
-    it("should handle component with null props", () => {
-      const message = {
-        ...baseMessage,
-        componentDecision: {
-          componentName: "SimpleCard",
-          props: null,
-          message: "",
-          componentState: null,
-        },
-      } as unknown as DbMessage;
-
-      const result = contentToV1Blocks(message);
-
-      expect(result).toHaveLength(1);
-      expect(result[0]).toEqual({
-        type: "component",
-        id: "comp_msg_123",
-        name: "SimpleCard",
-        props: {},
-        state: undefined,
-      });
-    });
-
     it("should preserve inline tool_use blocks in their natural position", () => {
       const message = {
         ...baseMessage,
@@ -431,67 +330,6 @@ describe("v1-conversions", () => {
       expect(result[2]).toEqual({
         type: "text",
         text: "Here are the results.",
-      });
-    });
-
-    it("should not duplicate tool_use when already present in content", () => {
-      const message = {
-        ...baseMessage,
-        role: "assistant",
-        content: [
-          { type: "text", text: "Searching..." },
-          {
-            type: "tool_use",
-            id: "call_xyz",
-            name: "get_data",
-            input: { id: "123" },
-          },
-        ],
-        toolCallRequest: {
-          toolName: "get_data",
-          parameters: [{ parameterName: "id", parameterValue: "123" }],
-        },
-        toolCallId: "call_xyz",
-      } as unknown as DbMessage;
-
-      const result = contentToV1Blocks(message);
-
-      // Should have 2 blocks (text + tool_use), NOT 3 (text + tool_use + duplicate tool_use)
-      expect(result).toHaveLength(2);
-      expect(result[0]).toEqual({ type: "text", text: "Searching..." });
-      expect(result[1].type).toBe("tool_use");
-    });
-
-    it("should suppress legacy fallback even when inline tool_use name differs from toolCallRequest", () => {
-      const message = {
-        ...baseMessage,
-        role: "assistant",
-        content: [
-          { type: "text", text: "Searching..." },
-          {
-            type: "tool_use",
-            id: "call_inline",
-            name: "search_items",
-            input: { query: "test" },
-          },
-        ],
-        toolCallRequest: {
-          toolName: "different_tool",
-          parameters: [{ parameterName: "id", parameterValue: "456" }],
-        },
-        toolCallId: "call_legacy",
-      } as unknown as DbMessage;
-
-      const result = contentToV1Blocks(message);
-
-      // Should only have the inline tool_use, not the legacy fallback
-      expect(result).toHaveLength(2);
-      expect(result[0]).toEqual({ type: "text", text: "Searching..." });
-      expect(result[1]).toEqual({
-        type: "tool_use",
-        id: "call_inline",
-        name: "search_items",
-        input: { query: "test" },
       });
     });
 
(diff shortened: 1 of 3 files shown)


apps/api/src/v1/v1-conversions.ts
```diff
@@ -216,10 +217,67 @@ export function contentToV1Blocks(
...
+  // from V1 streaming, and must be preserved in their natural position to
+  // maintain interleaved ordering (e.g., text -> tool_use -> text).
   if (Array.isArray(message.content)) {
     for (const part of message.content) {
+      const partType = (part as { type?: unknown }).type;
+
+      // Handle V1-specific content types stored inline in the DB content array.
+      // These are not part of ChatCompletionContentPart, so we handle them
+      // before calling contentPartToV1Block.
+      if (partType === "tool_use") {
```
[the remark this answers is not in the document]
seth-tambo: Added `isUiToolName` filter on the inline path to match the legacy fallback. abc2a120

apps/api/src/v1/v1-conversions.ts
```diff
@@ -216,10 +217,67 @@ export function contentToV1Blocks(
...
+            string,
+            unknown
+          >;
+          if (typeof decision.statusMessage === "string") {
+            input._tambo_statusMessage = decision.statusMessage;
+          }
+          if (typeof decision.completionStatusMessage === "string") {
+            input._tambo_completionStatusMessage =
+              decision.completionStatusMessage;
+          }
```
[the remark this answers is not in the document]
seth-tambo: Extracted `enrichInputWithTamboProperties` helper used by both paths. abc2a120

apps/api/src/v1/v1-conversions.ts
```diff
@@ -216,10 +217,67 @@ export function contentToV1Blocks(
...
+          isError?: boolean;
+        };
+        const toolResultBlock: V1ToolResultContentDto = {
+          type: "tool_result",
+          toolUseId: toolResultPart.toolUseId,
+          content: toolResultPart.content ?? [],
+          isError: toolResultPart.isError,
+        };
+        blocks.push(toolResultBlock);
+        continue;
```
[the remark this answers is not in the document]
seth-tambo: Added clarifying comment. abc2a120

akhileshrangani4 (dismissed):
Review posted in error - please disregard.

akhileshrangani4 (commented):
Review posted in error - please disregard.

apps/api/src/v1/v1-conversions.ts
```diff
@@ -100,6 +100,24 @@ const defaultContentConversionOptions: Required<ContentConversionOptions> = {
...
+function enrichInputWithTamboProperties(
+  input: Record<string, unknown>,
+  componentDecision: DbMessage["componentDecision"],
+): void {
+  if (!componentDecision) return;
+  const decision = componentDecision as unknown as Record<string, unknown>;
+  if (typeof decision.statusMessage === "string") {
+    input._tambo_statusMessage = decision.statusMessage;
+  }
+  if (typeof decision.completionStatusMessage === "string") {
```
akhileshrangani4, line 117: This function mutates `input` in place and returns `void`. AGENTS.md says to prefer immutability. Both call sites create fresh objects so there's no runtime bug, but the signature invites future misuse.

Consider returning the display properties as a separate object and letting the caller spread:

```ts
function getTamboDisplayProperties(
  componentDecision: DbMessage["componentDecision"],
): Record<string, string> {
  if (!componentDecision) return {};
  const decision = componentDecision as unknown as Record<string, unknown>;
  const props: Record<string, string> = {};
  if (typeof decision.statusMessage === "string") {
    props._tambo_statusMessage = decision.statusMessage;
  }
  if (typeof decision.completionStatusMessage === "string") {
    props._tambo_completionStatusMessage = decision.completionStatusMessage;
  }
  return props;
}
```

Then: `const input = { ...toolUsePart.input, ...getTamboDisplayProperties(message.componentDecision) }`

apps/api/src/v1/v1-conversions.ts
```diff
@@ -216,10 +235,59 @@ export function contentToV1Blocks(
     return blocks; // Tool messages don't have other content types
   }
 
-  // Convert standard content parts (non-tool messages)
-  // Including component blocks stored in the content array
+  // Convert standard content parts (non-tool messages).
+  // Including component blocks and V1-specific types stored in the content array.
+  // V1 types (tool_use, tool_result) may be present when the content was saved
```
akhileshrangani4, line 253: Nit: this `(part as { type?: unknown }).type` pattern duplicates what `getContentPartType()` already does right above in this same file. Could reuse it here for consistency.

apps/api/src/v1/v1-conversions.ts
```diff
@@ -216,10 +235,59 @@ export function contentToV1Blocks(
...
+  // from V1 streaming, and must be preserved in their natural position to
+  // maintain interleaved ordering (e.g., text -> tool_use -> text).
   if (Array.isArray(message.content)) {
     for (const part of message.content) {
+      const partType = (part as { type?: unknown }).type;
+
+      // Handle V1-specific content types stored inline in the DB content array.
+      // These are not part of ChatCompletionContentPart, so we handle them
+      // before calling contentPartToV1Block.
```
akhileshrangani4, line 262: The `as unknown as` cast trusts DB content fully. If a required field is missing (no `id`, `name` isn't a string, etc.), this silently produces a malformed DTO. Low priority since data comes from your own DB, but minimal runtime guards like `typeof toolUsePart.id === 'string'` would line up with the fail-fast philosophy.

Also, the tool_use block construction here is basically the same as the legacy fallback below (build `V1ToolUseContentDto`, call enrich, check `isUiToolName`). A shared `buildToolUseBlock(id, name, input, componentDecision)` helper would cut ~10 lines and give one source of truth.

apps/api/src/v1/v1-conversions.ts
```diff
@@ -100,6 +100,24 @@ const defaultContentConversionOptions: Required<ContentConversionOptions> = {
...
+/**
+ * Enrich a tool_use input object with _tambo_* display properties from componentDecision.
+ * Used by both the inline V1 path and the legacy fallback path to keep them in sync.
+ */
+function enrichInputWithTamboProperties(
+  input: Record<string, unknown>,
+  componentDecision: DbMessage["componentDecision"],
+): void {
+  if (!componentDecision) return;
+  const decision = componentDecision as unknown as Record<string, unknown>;
```
akhileshrangani4, line 112: The `as unknown as Record<string, unknown>` cast here is because `ComponentDecisionV2` doesn't include `statusMessage`/`completionStatusMessage` (they come from `LegacyComponentDecision` being spread at write time). Worth a follow-up issue to add those optional fields to the type so you don't need the double cast. Not blocking for this PR.

apps/api/src/v1/__tests__/v1-conversions.test.ts
```diff
@@ -393,6 +393,170 @@ describe("v1-conversions", () => {
...
+      });
+    });
+
+    it("should not duplicate tool_use when already present in content", () => {
+      const message = {
+        ...baseMessage,
+        role: "assistant",
+        content: [
+          { type: "text", text: "Searching..." },
+          {
```
akhileshrangani4: Tests look solid. One gap: there's no test where the inline `tool_use` has a *different* tool name than `toolCallRequest.toolName`. That would confirm the `foundToolUseInContent` flag suppresses the legacy fallback regardless of whether the tools match. Would be a nice addition.

akhileshrangani4 (commented):
Bug fix looks correct and well-tested. The interleaved ordering problem is properly solved, the dedup flag mirrors the existing `foundComponentInContent` pattern nicely, and 6 new tests cover all the key paths. No security or performance concerns.

Left a few suggestions inline, nothing blocking.

_Reviewed with the workflow review skill._

apps/api/src/v1/v1-conversions.ts
seth-tambo: Good call. Refactored to `getTamboDisplayProperties` which returns `Record<string, string>` instead of mutating. Both call sites now spread the result into their input objects.

apps/api/src/v1/v1-conversions.ts
seth-tambo: Fixed. Widened `getContentPartType` to accept `{ type?: unknown }` and replaced the inline cast with a call to it.

apps/api/src/v1/v1-conversions.ts
seth-tambo: Both addressed. Added runtime guards on the inline path (validates `id`/`name` are strings, `input` is an object). Also extracted `buildToolUseBlock` helper shared by both the inline and legacy fallback paths.

apps/api/src/v1/v1-conversions.ts
seth-tambo: Agreed, not blocking for this PR. Created a follow-up: https://linear.app/tambo/issue/TAM-1481/add-statusmessage-and-completionstatusmessage-to-componentdecisionv2

apps/api/src/v1/__tests__/v1-conversions.test.ts
seth-tambo: Added. New test: "should suppress legacy fallback even when inline tool_use name differs from toolCallRequest" confirms `foundToolUseInContent` suppresses the fallback regardless of tool name match.

alecf (dismissed):
I don't understand the "legacy" stuff - I think that was just the AI copying patterns from pre-v1 - I say nuke anything here having to do with "legacy" messages, we shouldn't need to support it (i.e. legacy messages probably come from the old API and the new API has been out long enough that we shouldn't have to support that case)

akhileshrangani4 (commented):
Review posted in error by automated agent - please disregard.

akhileshrangani4 (dismissed):
Review posted in error by automated agent - please disregard.

apps/api/src/v1/v1-conversions.ts
```diff
@@ -216,16 +248,67 @@ export function contentToV1Blocks(
...
+            : {};
+        blocks.push(
+          buildToolUseBlock(id, name, input, message.componentDecision),
+        );
+        continue;
+      }
+
+      // Content is already in V1 format since it was stored by V1 streaming;
+      // no need to run sub-parts through contentPartToV1Block.
+      if (partType === "tool_result") {
```
akhileshrangani4, line 291: The `tool_use` handler above validates `id` and `name` are strings and logs a warning for malformed blocks. This `tool_result` handler does none of that. If `toolUseId` is missing or not a string, you get a silently invalid API response.

Worth adding the same pattern here:

```typescript
if (partType === "tool_result") {
  const toolResultPart = part as unknown as Record<string, unknown>;
  const toolUseId = toolResultPart.toolUseId;
  if (typeof toolUseId !== "string") {
    logger.warn(
      \`Skipping malformed tool_result block (toolUseId: \${typeof toolUseId})\`,
    );
    continue;
  }
  // ... rest of construction
}
```

Also no tests for malformed `tool_result` or `isError: true` on the inline path.

apps/api/src/v1/v1-conversions.ts
```diff
@@ -100,7 +102,42 @@ const defaultContentConversionOptions: Required<ContentConversionOptions> = {
...
+  id: string,
+  name: string,
+  input: Record<string, unknown>,
+  componentDecision: DbMessage["componentDecision"],
+): V1ToolUseContentDto {
+  return {
+    type: "tool_use",
+    id,
+    name,
+    input: { ...input, ...getTamboDisplayProperties(componentDecision) },
```
akhileshrangani4, line 136: Nit: the display props spread comes after `input`, so if a user's tool happens to have a parameter named `_tambo_statusMessage`, it gets silently overwritten. Unlikely but easy to avoid by flipping the spread order so user input wins:

```typescript
input: { ...getTamboDisplayProperties(componentDecision), ...input },
```

akhileshrangani4 (commented):
Nice cleanup. The inline content handling is the right call and the ordering fix is solid. A couple things worth addressing.

michaelmagan:

PR #2781, opened 2026-04-07
docs(skills): add Tambo skills reference to plugin skills

akhileshrangani4:
## Summary
- Add a `references/skills.md` doc to both `building-with-tambo` and `generative-ui` Claude Code plugin skills, covering the Tambo skills feature (project-scoped agent instructions managed via CLI and dashboard)
- Update both SKILL.md files to reference the new doc as a "next step after setup"

## Why
Neither plugin skill mentioned Tambo's skills feature, so agents helping users set up Tambo apps had no way to surface skills as a next step. This closes that gap using the same duplicated-reference pattern as the existing shared docs.

Fixes TAM-1473

## Test Plan
- [x] Both `references/skills.md` files are identical (verified via `diff`)
- [x] SKILL.md reference entries follow the existing format (bold load trigger + description)
- [x] CLI commands in the doc match the actual `tambo skills` subcommands
- [x] Field limits match `packages/core/src/skills.ts` constants
- [x] Quick Start, Contents annotations, `npx` prefix, dashboard terminology, and H1 title all follow patterns from existing reference docs (`cli.md`, `components.md`, `tools-and-context.md`)


alecf (approved):
looks good - I feel like it would be good to mention/include the skill-creator skill from skills.md, something like "if the skill-creator skill is available, use it to suggest improvements to the user's skill" or "use it to creat the user's skill" or something

The challenge with having the skill-creator actually create the skill is that it tends to create a dummy skill directory from a template, and then delete the extra files, so it might be better as a reference for reviewing skills instead.

akhileshrangani4:
@alecf that is a great idea, will add that.

PR #2783, opened 2026-04-08
fix(web,react-sdk): fix setState-during-render error and agent settings UI

akhileshrangani4:
## Summary
- Fix React warning "Cannot update a component (`TamboRegistryProvider`) while rendering a different component (`TamboInteractableProvider`)" by converting `toolComponentOwnership` from `useState` to `useRef` -- the ownership map is pure bookkeeping that never drives rendering
- Remove phantom sidebar from `SettingsPageSkeleton` and `AgentPageSkeleton` loading states (the actual pages have no sidebar)
- Reorder agent settings sections: LLM Providers > Custom Instructions > MCP > Skills > Memory > Tool Call
- Add missing `MemorySettingsSkeleton`

## Why
LLM provider configuration is the first thing most users need to set up when configuring their agent, but it was buried at the bottom of the page. The new order puts the most essential setup (provider keys, instructions) at the top and less frequently changed settings (memory, tool call limits) at the bottom, which feels more intuitive.

## Test plan
- [x] Open project agent tab while data is loading -- skeleton should match page layout (no sidebar flash)
- [x] Open project settings tab while loading -- same check
- [x] Verify agent tab section order matches: LLM Providers, Custom Instructions, MCP, Skills, Memory, Tool Call
- [x] Verify no "Cannot update a component while rendering" console error when navigating to agent tab
- [x] `npm test` passes

### DIFF
--- a/apps/web/components/dashboard-components/agent-settings.tsx
+++ b/apps/web/components/dashboard-components/agent-settings.tsx
@@ -61,29 +61,15 @@ export function AgentSettings({ projectId }: AgentSettingsProps) {
           Configure the behavior of your Tambo agent.
         </p>
         <div className="space-y-6">
+          <InteractableProviderKeySection
+            projectId={project.id}
+            onEdited={handleRefreshProject}
+          />
+
           <InteractableCustomInstructionsEditor
             projectId={project.id}
             customInstructions={project.customInstructions}
             allowSystemPromptOverride={project.allowSystemPromptOverride}
-            onEdited={handleRefreshProject}
-          />
-
-          <InteractableSkillsSection
-            projectId={project.id}
-            defaultLlmProviderName={project.defaultLlmProviderName ?? undefined}
-            defaultLlmModelName={project.defaultLlmModelName ?? undefined}
-          />
-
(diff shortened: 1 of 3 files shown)


apps/web/components/dashboard-components/agent-settings.tsx
```diff
@@ -61,23 +61,28 @@ export function AgentSettings({ projectId }: AgentSettingsProps) {
           Configure the behavior of your Tambo agent.
         </p>
         <div className="space-y-6">
+          <InteractableProviderKeySection
```
MichaelMilstead: I personally think it makes more sense to immediately show the "behavioral" settings first, and LLM config after, to avoid new users being overwhelmed and thinking they need to setup some LLM provider before trying things out. Maybe in a later PR the LLM config deserves its own separate tab

MichaelMilstead (approved):
Nice, just one comment about settings order

apps/web/components/dashboard-components/agent-settings.tsx
akhileshrangani4: makes sense, moved it back down

PR #2919, opened 2026-06-06
fix(client): mcp server connection failures are no longer silent

### LINKED ISSUE #2917: MCP Server Connection Failures Are No Longer Silent
## What was the issue?

MCP servers in TamboClient would fail silently. When a connection failed, it just logged to console and moved on — no way for users to know their servers aren't connected, no way to retry, no way to check status programmatically.

## What changed?

**Connection status is now tracked.** Every MCPClient has a status — `connecting`, `connected`, or `failed` — and captures the error message if something goes wrong. Instead of throwing on failure, the client is still returned so you can inspect what happened.

**Status is exposed.** `getMcpClients()` now returns connection info (status, error, server details) instead of raw client objects. You can see exactly which servers are connected and which aren't.

**Failed servers can be retried.** Calling `connectMcpServer()` again on a failed connection clears the old failure and tries again.

**Status changes are observable.** There's a new `onMcpConnectionChange` callback on TamboClientOptions that fires whenever an MCP server connects or fails, so you can react to changes (show a warning, trigger a retry, etc.).

**No more silent failures.** The fire-and-forget `.catch()` in the constructor is gone — connection results are tracked properly whether they succeed or fail.

soumojit-D48:
## fixes: #2917

MCP servers in TamboClient fail silently — connection errors were caught and only logged to console with no way to detect, retry, or react.

Now every MCP client tracks its connection status (`connecting` / `connected` / `failed`) and surfaces it through `getMcpClients()`. Failed connections can be retried by calling `connectMcpServer()` again. The new `onMcpConnectionChange` callback lets consumers react to status changes in real time.

The fire-and-forget `.catch()` in the constructor is gone — connection results are visible regardless of success or failure.

**Key changes:**
- `MCPClient.create()` no longer throws on failure — it sets `status = "failed"` and stores the error, then returns the instance
- `getMcpClients()` now returns `Record<string, MCPConnectionInfo>` with status, error, and server details instead of raw client objects
- `connectMcpServer()` clears previous failures before retrying, giving you a clean reconnect path
- `disconnectMcpServer()` properly cleans up both the client and its stored server info
- Added `onMcpConnectionChange` callback to `TamboClientOptions` — fires with `(serverKey, status, error?)` on every status change

**Files touched:** `mcp-client.ts`, `tambo-client.ts`, `mcp/index.ts`, `index.ts`

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -127,6 +127,8 @@ export {
   hasRequestedSchema,
 } from "./mcp/index";
 export type {
+  MCPClientStatus,
+  MCPConnectionInfo,
   MCPToolCallResult,
   MCPToolSpec,
   MCPElicitationHandler,
--- a/packages/client/src/mcp/index.ts
+++ b/packages/client/src/mcp/index.ts
@@ -1,5 +1,7 @@
 export {
   MCPClient,
+  type MCPClientStatus,
+  type MCPConnectionInfo,
   type MCPToolCallResult,
   type MCPToolSpec,
   type MCPElicitationHandler,
--- a/packages/client/src/mcp/mcp-client.ts
+++ b/packages/client/src/mcp/mcp-client.ts
@@ -14,7 +14,7 @@ import {
   ElicitResult,
 } from "@modelcontextprotocol/sdk/types.js";
 import { JSONSchema7 } from "json-schema";
-import { MCPTransport } from "../model/mcp-server-info";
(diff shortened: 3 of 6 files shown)


soumojit-D48:
@alecf Hi, Pls Review this PR and let me know, Thank u!

lachieh:
Closing in favor of #2918 which includes test coverage for the linked issue.

